PR #26, opened 2018-01-25
Use python and nosetests versions explicitely

vandenheuvel:
I needed these changes to get this script working on Arch. @qstokkink will this break Ubuntu or Debian?


qstokkink:
Ok to test

qstokkink:
@vandenheuvel apparently, yes

qstokkink:
If you add another case to the if for `nosetests2` and and change `python2.7` this PR should be good to go I think.

vandenheuvel:
I'm currently searching on how to check the availability of a command. What do you think is the best way to do so?

qstokkink:
You can use the line 10 format:
```bash
if nosetests --version >>/dev/null 2>&1; then
```

PR #34, opened 2018-02-02
Improved TrustChain crawler

devos50:
Added a dedicated message for a crawler response and each crawl now has an identifier. In addition, we also keep track of outstanding crawl requests in a request cache.

Also added block pair messages and added support for broadcasts of blocks.


devos50:
@qstokkink ready for review!

PR #38, opened 2018-02-09
Various bug fixes and changes to the tunnel community

devos50:


ipv8/messaging/anonymization/community.py
```diff
@@ -887,6 +887,13 @@ def on_extended(self, source_address, data, _):
         circuit = self.circuits[circuit_id]
         self._ours_on_created_extended(circuit, payload)
 
+    def handle_raw_data(self, circuit, origin, data):
```
qstokkink: This is just nitpicking, but according to Pylint callbacks should start with `on_` or `cb_` or end with `_cb`. Could you use `on_` so in the future we don't get annoying Pylint errors?

qstokkink: ⏫

devos50: Will change it. This PR is still WIP btw, I want to make some more changes 👍

ipv8/messaging/anonymization/community.py
```diff
@@ -354,15 +361,21 @@ def remove_relay(self, circuit_id, additional_info='', destroy=False, got_destro
         if destroy:
             self.destroy_relay(to_remove, got_destroy_from=got_destroy_from)
 
+        removed_relays = []
         for cid in to_remove:
             # Remove the relay
             self.logger.info("Removing relay %d %s", cid, additional_info)
-            relay = self.relay_from_to.pop(cid, None)
+            removed_relays.append(self.relay_from_to.pop(cid, None))
```
qstokkink: This may append `None` values.

ipv8/messaging/anonymization/community.py
```diff
@@ -354,15 +361,21 @@ def remove_relay(self, circuit_id, additional_info='', destroy=False, got_destro
...
 
             # Remove old session key
             if cid in self.relay_session_keys:
                 del self.relay_session_keys[cid]
 
+            if cid in self.directions:
+                del self.directions[cid]
```
qstokkink: This sort of construction is actually rather fragile. We should start using `self.somedict.pop(somekey, None)`.

PR #67, opened 2018-03-04
Added delay when removing tunnel info, fixed mid encoding

devos50:
In this PR, I've added an optional delay when removing/destroying circuits, relays and exit sockets. This allows some post-mortem data to flow over the circuit, like payout messages in Tribler.

I've also addressed mid encoding in the `Tunnel` object. The way we encode it is a common source of errors. We now always store the mid in binary. We should eventually fix it to store a `Peer` object instead but requires takes a bit more work.

### DIFF
--- a/ipv8/messaging/anonymization/community.py
+++ b/ipv8/messaging/anonymization/community.py
@@ -943,8 +943,8 @@ class TunnelCommunity(Community):
                 circuit.beat_heart()
                 self.increase_bytes_received(circuit, len(packet))
 
-                if DataChecker.could_be_dispersy(data):
-                    self.logger.debug("Giving incoming data packet to dispersy")
+                if DataChecker.could_be_ipv8(data):
+                    self.logger.debug("Giving incoming data packet to IPv8")
                     self.logger.debug("CIRCUIT ID = %d", circuit_id)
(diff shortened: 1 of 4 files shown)


qstokkink:
Question: what is the expected behavior on `unload()`?

devos50:
@qstokkink good question, I didn't consider the situation when unloading an overlay. In that case, we would like to remove the data immediately I guess? This could be realized by adding a `remove_now` keyword argument to `remove_relay`, `remove_circuit` etc and if it's true, we remove the data immediately. What do you think about this?

qstokkink:
@devos50 considering the payout (actual money) happens after circuit destruction, I think the entire community should wait until circuits have been destructed and have been paid out. Although it would take up to the `remove_tunnel_delay` for the community to unload (which is a relatively long time).

However, considering this functionality supports your Market, I'm fine with whatever you think is best.

devos50:
@qstokkink I think that when a community unloads, it shouldn't wait for additional incoming messages. I know this might cause a payout not to complete but this is not the end of the world since at most 250MB is lost (the maximum amount of data that can go over a circuit).

I've updated my PR.

PR #72, opened 2018-03-14
Wrapped unload overlay result in maybeDeferred

devos50:


qstokkink:
It's probably better to do this in the `unload_overlay()` function itself.

PR #77, opened 2018-03-17
Improved crawling in the TrustChain community

devos50:
In this PR, I improve the crawling mechanism in TrustChain. The most important change is that I send a dummy block back when receiving a crawl request and when there are no blocks to send to the request initiator. This avoids that the initiator should wait until the crawl request cache times out.

I've also added a test for crawling block pairs.


devos50:
retest this please

PR #84, opened 2018-03-25
Increased timeout of crawl request cache

devos50:
Also added a check to see whether the cache still exists when a crawl response is received (the timeout could have been triggered already which removes this cache).


devos50:
retest this please

devos50:
mac seems to be _very_ slow, leading to build timeouts. We should investigate this tomorrow.

devos50:
retest this please

PR #87, opened 2018-03-27
READY: Improved IPv8 Trustchain crawler

devos50:


devos50:
I will close this PR, in favor of #94 and a design that is more reliable and future-proof.

PR #89, opened 2018-03-28
Added shutdown method in TaskManager

devos50:


ipv8/taskmanager.py
```diff
@@ -1,5 +1,6 @@
 from threading import Lock
 
+import logging
```
qstokkink: This should be on the first line.

PR #94, opened 2018-04-02
Re-implemented live edges

devos50:
Now that the Trustchain database is faster, I re-implemented the live edges walker in IPv8. This is the basic implementation as initially designed by @qstokkink.


devos50:
@qstokkink I've implemented the requested changes.

PR #96, opened 2018-04-06
Refactored TrustChain so it's a single ledger now

devos50:
In this PR, I refactored the TrustChain overlay so it's a generic ledger that is reusable across several domains. The most important change is that each transaction now should have a type (i.e. `ask`, `tribler_bandwidth`, `attestation` etc). This transaction type is also stored in the database as a separate column for easy lookups. For instance, this allows us to get all blocks with an `attestation` transaction.

In addition, I also added a timestamp to each TrustChain record. These changes are included in this PR since it also breaks backward compatibility so I guess it's ok to combine it with the other breaking change (merging all chains together).

### DIFF
--- a/doc/trustchain.md
+++ b/doc/trustchain.md
@@ -1,18 +1,58 @@
-This document entails the implementation of TrustChain in IPv8.
-The implementation is provided in 4 files:
+This document entails a high-level overview of the implementation of TrustChain in IPv8.
+TrustChain is a scalable, tamper-proof and distributed ledger, built for secure accounting.
+For more information about TrustChain itself, we refer to our [IETF internet standard](https://tools.ietf.org/html/draft-pouwelse-trustchain-01).
+Additional information can be found [in our published scientific article](https://www.sciencedirect.com/science/article/pii/S0167739X17318988).
 
- - `block.py`: data structure of a block in the Blockchain
- - `community.py`: the ipv8 overlay providing higher order TrustChain logic
- - `database.py`: wrapper to query and update the database
- - `payload.py`: message format and datastructures for communication
- 
-## Payload
+The TrustChain implementation is provided in 4 files:
 
-The TrustChain uses 2 messages to communicate between peers: a `CrawlRequest` and a `HalfBlock` message.
+ - `block.py`: this file defines the data structure of a block in the TrustChain ledger
+ - `community.py`: the ipv8 overlay providing higher order TrustChain logic (i.e. handling incoming blocks and exploring other blocks in the network)
+ - `database.py`: wrapper to query and update the database with known TrustChain blocks
+ - `payload.py`: defines the message format and data structures for communication between peers
 
-The `CrawlRequest` request message is sent when one peer wishes to receive blocks from another peer.
-A `CrawlRequest` takes a sequence number as an argument.
-This sequence number specifies from which sequence number forward, blocks will be sent back (up to 100 blocks in response).
+The logic present in each file is now briefly explained.
+
+## block.py
+The `block.py` file contains the definition of a TrustChain half block (in the `TrustChainBlock` class).
+Note that each valid transaction contains exactly two blocks, signed by both involved parties.
+Each block in TrustChain contains the following properties:
+
+| Property | Type | Description |
+| --- | --- | --- |
+| _type_ | string | The type of the block. Can be any arbitrary string. |
+| _transaction_ | dictionary | The transaction that this block describes. |
+| _public_key_ | string | The public key in binary format of the block creator. |
+| _link_public_key_ | string | The public key of the transaction counterparty. |
+| _sequence_number_ | integer | The sequence number of the block. The sequence number of the genesis block is 1. |
+| _link_sequence_number_ | integer | The sequence number of the linked block. |
+| _previous_hash_ | string | The SHA256 hash of the previous block in the chain. |
+| _signature_ | string | The signature of the block. |
+| _timestamp_ | integer | The epoch timestamp defining when this block is created (with milliseconds precision). |
+
+
+The attributes within the `TrustChainBlock` class have a high overlap with those in a `HalfBlockPayload`.
+
+### Block Validation
+Each incoming block and new block created are verified for correctness.
+These checks are performed by the `validate` method inside the `TrustChainBlock` class.
+Discovered blocks that are invalid are ignored and not added to the database.
+The validation method also aims to detect fraudulent operations performed by users, such as a double-spend attack.
+
+## payload.py
+
+TrustChain defines six different messages types, used to request signatures in blocks and exchange knowledge of existing blocks.
+We now describe the functionality of each message:
+
+| Name | Description |
+| --- | --- |
+| _CrawlRequestPayload_ | This message describes a crawl request for another user. It contains the public key of the receiver of the crawl request, the index of the block(s) being requested and a random identifier. |
+| _CrawlResponsePayload_ | Response to a _CrawlRequest_ message. Contains a TrustChain block, the crawl identifier, the total amount of blocks being sent during the crawl, and the index of this block during the crawl. |
+| _HalfBlockPayload_ | Contains a single half block. |
+| _HalfBlockBroadcastPayload_ | Contains a single half block and a TTL value. |
+| _HalfBlockPairPayload_ | Contains a pair of half blocks. |
+| _HalfBlockPairBroadcastPayload_ | Contains a pair of half blocks and a TTL value. |
+
+The sequence number in the _CrawlRequestPayload_ specifies from which sequence number forward, blocks will be sent back (up to 100 blocks in response).
 Alternatively, the sequence number can also be negative.
 A negative sequence number implies the offset from the last block.
 For example, when peer **A** performs a crawl request from a peer **B** with 8 blocks (numbered 1 through 8):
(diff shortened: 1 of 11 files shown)


qstokkink:
Could we keep this on hold for a few days until I stabilize the attestation community again?

devos50:
Sure, no problem!

ipv8/attestation/trustchain/community.py
```diff
@@ -60,6 +59,9 @@ def __init__(self, *args, **kwargs):
         self.logger.debug("The trustchain community started with Public Key: %s",
                           self.my_peer.public_key.key_to_bin().encode("hex"))
         self.broadcast_block = True  # Whether we broadcast a full block after constructing it
+        self.listeners_map = {}  # Map of block_type -> [callbacks]
```
qstokkink: Could you refactor the listeners into a single `Listener` class? The listener object, `should_sign` and `block_class` should all be connected anyway. This is starting to look like the TunnelCommunity.

devos50: @qstokkink the `block_class` and `Listener` are not necessarily connected, they can be different objects.

Also, there can be multiple listeners for one block type but there's at most one `block_class` for each type. I don't see how we could generalize this nicely. Do you have any suggestions?

qstokkink: I don't see how a listener could possibly comprehend what it is called back with, if it does not know the block class it is getting called back with. `self.notify_listeners(block)` gives a block of type `block = self.get_block_class(block_type)`.

A Listener could be something like:

```python
class TrustChainBlock(SomeBlockSuperclass):
    TYPE = 'bandwidth!'

class MyBeautifulListener(SomeListenerSuperclass):
    BLOCK_CLASS = TrustChainBlock

    def should_sign(block):
        return True
```

ipv8/attestation/trustchain/community.py
```diff
@@ -230,10 +284,6 @@ def process_half_block(self, blk, peer):
         self.logger.debug("Block validation result %s, %s, (%s)", validation[0], validation[1], blk)
         if validation[0] == ValidationResult.invalid:
             return
-        elif not self.persistence.contains(blk):
-            self.persistence.add_block(blk)
```
qstokkink: Why was this removed?

devos50: Because the block is already added to the database in the `validate_persist_block` method if it is valid. These two lines I removed here were duplicate and unnecessary.

devos50:
@qstokkink I'm not sure why it is necessary to impose shutdown constraints on the listeners/TrustChain community. I think it is sufficient to simply not notify the listeners when the TrustChain community is shutting down or has finished doing so. I've implemented it like this for now (by keeping track of a `shutting_down` variable in the community class). Please let me know whether you agree or not 👍

ichorid:
(Disclaimer: I'm still not familiar with how TrustChain operates, so I am not sure about this one.)

@qstokkink, I wonder how this change would affect your judgment about the potential TrustChain replication problem. I guess, it was you who told me that the typical TrustChain for channel voting is only few hundred kilobytes, so it is easy for counterparties to copy it over completely, store it locally and check it.
I've checked TrustChain DB sizes for tunnel helpers on our oldest hose: some files are over 500 MB in size. If all the data is in the same domain now, this means the counterparty would have to download the whole chain including all records on traffic exchange, etc. 
Am I right?

devos50:
retest this please

devos50:
@qstokkink I've updated and improved the TrustChain documentation in a separate commit after your review.

ipv8/attestation/trustchain/community.py
```diff
@@ -39,18 +39,26 @@ class TrustChainCommunity(Community):
...
-                        "a2961b003affd2964f92c457572f2f79de0968c42698e2d1cfb371dd71b275332a0a4c19f35f16166272baae8e23"
-                        "0bba377cc5c40643b83206088075559ec2f13a090e8786d04d84802268bef12e52983978da360589a2b7e293ce4f"
-                        "16d02f37da2c3256f4703b9623d3750f7af437befebc8935c0f0726f58c1c1e9").decode("HEX"))
+    master_peer = None  # Determined at runtime
 
     def __init__(self, *args, **kwargs):
         working_directory = kwargs.pop('working_directory', '')
         db_name = kwargs.pop('db_name', self.DB_NAME)
+        self.testnet = kwargs.pop('testnet', False)
+        TrustChainCommunity.master_peer = self.TESTNET_PEER if self.testnet else self.MAINNET_PEER
```
ichorid: @devos50 Could you please instead do it in a more generalized way, like I did with `bootstrap_override` option? It would make testing much more convenient for other members of the team.

ipv8/test/attestation/trustchain/test_block.py
```diff
@@ -14,25 +14,25 @@ class TestBlock(TrustChainBlock):
...
                 self.key = key
             else:
                 self.key = crypto.generate_key(u"curve25519")
 
-            TrustChainBlock.__init__(self, (
-                encode(transaction), self.key.pub().key_to_bin(), random.randint(50, 100), other, 0,
-                sha256(str(random.randint(0, 100000))).digest(), 0, 0))
+            TrustChainBlock.__init__(self, (block_type,
+                                            encode(transaction), self.key.pub().key_to_bin(), random.randint(50, 100),
+                                            other, 0, sha256(str(random.randint(0, 100000))).digest(), 0, 0, 0))
```
ichorid: Formatting? Do all these spaces really increase readability?

ipv8/attestation/trustchain/payload.py
```diff
@@ -237,16 +264,17 @@ class HalfBlockPairBroadcastPayload(HalfBlockPairPayload):
...
+                 link_sequence_number2, previous_hash2, signature2, block_type2, transaction2, timestamp2, ttl):
         super(HalfBlockPairBroadcastPayload, self).__init__(public_key1, sequence_number1, link_public_key1,
                                                             link_sequence_number1, previous_hash1, signature1,
-                                                            transaction1, public_key2, sequence_number2,
-                                                            link_public_key2, link_sequence_number2, previous_hash2,
-                                                            signature2, transaction2)
+                                                            block_type1, transaction1, timestamp1, public_key2,
+                                                            sequence_number2, link_public_key2, link_sequence_number2,
+                                                            previous_hash2, signature2, block_type2, transaction2,
+                                                            timestamp2)
```
ichorid: Using these many unnamed arguments with basically no formatting is asking for troubles...

ipv8/attestation/trustchain/payload.py
```diff
@@ -168,27 +183,31 @@ class HalfBlockPairPayload(Payload):
...
     """
 
-    format_list = ['74s', 'I', '74s', 'I', '32s', '64s', 'varlenI'] * 2
+    format_list = ['74s', 'I', '74s', 'I', '32s', '64s', 'varlenI', 'varlenI', 'Q'] * 2
 
     def __init__(self, public_key1, sequence_number1, link_public_key1, link_sequence_number1, previous_hash1,
-                 signature1, transaction1, public_key2, sequence_number2, link_public_key2, link_sequence_number2,
-                 previous_hash2, signature2, transaction2):
+                 signature1, block_type1, transaction1, timestamp1, public_key2, sequence_number2, link_public_key2,
+                 link_sequence_number2, previous_hash2, signature2, block_type2, transaction2, timestamp2):
```
ichorid: That's a lot of unnamed arguments. Thank God we don't have to worry if we misplace one of them because Python will figure out the type conversions :see_no_evil:

ipv8/attestation/trustchain/payload.py
```diff
@@ -77,12 +83,13 @@ class HalfBlockBroadcastPayload(HalfBlockPayload):
...
-    format_list = ['74s', 'I', '74s', 'I', '32s', '64s', 'varlenI', 'I']
+    format_list = ['74s', 'I', '74s', 'I', '32s', '64s', 'varlenI', 'varlenI', 'Q', 'I']
 
     def __init__(self, public_key, sequence_number, link_public_key, link_sequence_number, previous_hash,
-                 signature, transaction, ttl):
+                 signature, block_type, transaction, timestamp, ttl):
         super(HalfBlockBroadcastPayload, self).__init__(public_key, sequence_number, link_public_key,
-                                                        link_sequence_number, previous_hash, signature, transaction)
+                                                        link_sequence_number, previous_hash, signature,
+                                                        block_type, transaction, timestamp)
```
ichorid: A lot of unnamed arguments.

ipv8/attestation/trustchain/block.py
```diff
@@ -34,12 +35,15 @@ def __init__(self, data=None, serializer=Serializer()):
...
             self.insert_time = None
         else:
-            _, self.transaction = decode(str(data[0]))
-            (self.public_key, self.sequence_number, self.link_public_key, self.link_sequence_number, self.previous_hash,
-             self.signature, self.insert_time) = (data[1], data[2], data[3], data[4], data[5], data[6], data[7])
+            _, self.transaction = decode(str(data[1]))
+            (self.type, self.public_key, self.sequence_number, self.link_public_key, self.link_sequence_number,
+             self.previous_hash, self.signature, self.timestamp, self.insert_time) = (data[0], data[2], data[3],
+                                                                                      data[4], data[5], data[6],
+                                                                                      data[7], data[8], data[9])
```
ichorid: I love that formatting! :+1:

doc/trustchain.md
```diff
@@ -1,18 +1,58 @@
...
+| --- | --- | --- |
+| _type_ | string | The type of the block. Can be any arbitrary string. |
+| _transaction_ | dictionary | The transaction that this block describes. |
+| _public_key_ | string | The public key in binary format of the block creator. |
+| _link_public_key_ | string | The public key of the transaction counterparty. |
+| _sequence_number_ | integer | The sequence number of the block. The sequence number of the genesis block is 1. |
+| _link_sequence_number_ | integer | The sequence number of the linked block. |
+| _previous_hash_ | string | The SHA256 hash of the previous block in the chain. |
+| _signature_ | string | The signature of the block. |
+| _timestamp_ | integer | The epoch timestamp defining when this block is created (with milliseconds precision). |
```
ichorid: Thumbs up for documentation! :1st_place_medal:

ipv8/attestation/trustchain/community.py
devos50: I'm not sure whether it is desirable for the TrustChain community to generalise which master peer we are using (i.e. make it possible to change the master peer from the IPv8 configuration dictionary). The master peer is rarely changed, except when we push some changes that are not backward compatible (like this PR).

In this case, I argue that it is helpful to make a distinction between a mainnet and testnet. We should use the testnet when running integration tests and experiments. Besides this mainnet/testnet, I do not see another use-case for changing the master peer and such, I don't think it is necessary to generalize this.

ipv8/test/attestation/trustchain/test_block.py
devos50: I think it increases readability and it follows [the PEP8 conventions](https://www.python.org/dev/peps/pep-0008/).

ipv8/attestation/trustchain/payload.py
devos50: I agree. Payloads which contain much data have many arguments. Maybe we can put each argument on a new line?

ipv8/attestation/trustchain/payload.py
devos50: In Python, we have to rely on proper unit tests to make sure that we do not make mistakes when passing arguments.

devos50:
@ichorid I have some comments on your comments 👍

ipv8/attestation/trustchain/payload.py
ichorid: :crying_cat_face:

ipv8/test/attestation/trustchain/test_block.py
ichorid: If it fits in 80 lines that's OK.

ipv8/attestation/trustchain/payload.py
ichorid: The problem is not in the formatting, but in the number of arguments and potential pitfall of mis-assigning them. I would prefer an intermediate object, or dict, or named arguments instead.
If you think it's OK to leave it as it is, that's up to you. Its only a minor note on style, nothing serious.

qstokkink:
retest this please

devos50:
retest this please

PR #107, opened 2018-04-12
Added support for smaller prime generation

### LINKED ISSUE #106: Add support for smaller primes in attestation library
Right now the smallest primes we can generate are 512 bit. This makes sense for RSA, but this leads to1024 bit EC keys, which is ridiculous. We can use the `gensafeprime` OpenSSL bindings for generating primes smaller than 512 bit.

```python
import gensafeprime 
gensafeprime.generate(bitspace)
```

We will need to update the Jenkins Android builder and the `requirements.txt`.

qstokkink:
Fixes #106 

This sets the default prime size to 128 bit, this leads to 256 bit EC keys (which is still 4 times larger than the curve25519 keys we use for TrustChain).

And we only use this for a 2 bit message space 🤓

### DIFF
--- a/ipv8/attestation/wallet/primitives/cryptosystem/boneh.py
+++ b/ipv8/attestation/wallet/primitives/cryptosystem/boneh.py
@@ -86,18 +86,23 @@ def get_good_wp(n, p=None):
     return p, wp
 
 
-def generate_primes(key_size=512):
+def generate_primes(key_size=128):
     """
-    Generate some primes. Key size in bits (minimum 512).
+    Generate some primes. Key size in bits.
     """
-    from cryptography.hazmat.backends import default_backend
-    from cryptography.hazmat.primitives.asymmetric import rsa
-    private_key = rsa.generate_private_key(public_exponent=65537,key_size=key_size,backend=default_backend())
-    private_numbers = private_key.private_numbers()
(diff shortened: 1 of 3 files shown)


qstokkink:
@devos50 or @xoriole if one of you has time, could you install the `gensafeprime` package on the testing machines?

xoriole:
The package is installed in bbq and mac but on windows, it did not work with the currently used OpenSSL 1.0.2 version. However, it does work with OpenSSL 1.1. Bundling both might not be a good idea. What do you think?

qstokkink:
@xoriole Thanks! Also, hmm.. I guess newer is better? How much work would this be/what are the consequences of doing the double bundling?

xoriole:
Qt is incompatible with the newer version of OpenSSL so we should have a proper discussion about it.

qstokkink:
@xoriole bbq also seems to be having some trouble with OpenSSL.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
We probably need to restart the Windows Jenkins slave for the PATH changes to take effect.

qstokkink:
retest this please

qstokkink:
@xoriole do you have time to review this?

PR #118, opened 2018-04-27
Added cooldown for intro response crawls

qstokkink:
This avoids IO bombing a node to death in (especially when a node does not know many other peers and nodes keep approaching the same counterparty).

### DIFF
--- a/ipv8/test/attestation/trustchain/test_community.py
+++ b/ipv8/test/attestation/trustchain/test_community.py
@@ -331,3 +331,42 @@ class TestTrustChainCommunity(TestBase):
         for node_nr in [0, 1]:
             self.assertIsNotNone(self.nodes[node_nr].overlay.persistence.get_latest(block1.public_key))
             self.assertIsNotNone(self.nodes[node_nr].overlay.persistence.get_latest(block2.public_key))
+
+    @twisted_wrapper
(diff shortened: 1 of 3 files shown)


ipv8/test/attestation/trustchain/test_community.py
```diff
@@ -331,3 +331,43 @@ def test_broadcast_half_block_pair(self):
...
+
+        # Perform the first crawl with all nodes
+        yield self.introduce_nodes()
+
+        for node_nr in [0, 1]:
+            self.assertIsNotNone(self.nodes[node_nr].overlay.persistence.get(my_pubkey, 1))
+            self.assertEqual(self.nodes[node_nr].overlay.persistence.get(my_pubkey, 1).link_sequence_number,
+                             UNKNOWN_SEQ)
+
+        self.nodes[0].overlay.sign_block(node3.my_peer, public_key=node3_pubkey, transaction={})
```
devos50: You can yield the `sign_block` so you don't need the `self.deliver_messages` call 😀

qstokkink: Ah, yes, good point 👍

qstokkink:
@devos50  resolved 👍

PR #122, opened 2018-04-27
Added corrupt chain recovery

### LINKED ISSUE #83: Infinite loop when receiving invalid block during crawl request when wanting to sign a block
I found an infinite loop when a node A wants to create a new block with a counterparty B. Suppose that the latest block of the chain of A is invalid (possibly due to a software bug or he could have inserted this himself). Now, when A creates a new block with node B and B does not have the latest blocks in the chain of A, B will issue a crawl request. Now, A will send the invalid block to B, which B will _not_ insert in his database. While this is expected, the problem here is that B will now call the `process_half_block` method again and the process continues.

I've written the following test to confirm this:

```
    @twisted_wrapper
    def test_invalid_block(self):
        yield self.introduce_nodes()
        invalid_block = TestBlock(key=self.nodes[0].overlay.my_peer.key)
        invalid_block.signature = 'a' * 64
        self.nodes[0].overlay.persistence.add_block(invalid_block)

        his_pubkey = self.nodes[0].network.verified_peers[0].public_key.key_to_bin()
        yield self.nodes[0].overlay.sign_block(self.nodes[0].network.verified_peers[0], public_key=his_pubkey,
                                               transaction={})

        yield self.deliver_messages()
```

@qstokkink any comments on this?

qstokkink:
Fixes #83

### DIFF
--- a/ipv8/attestation/trustchain/community.py
+++ b/ipv8/attestation/trustchain/community.py
@@ -135,7 +135,7 @@ class TrustChainCommunity(Community):
         assert linked is None or linked.link_sequence_number == UNKNOWN_SEQ, \
             "Cannot counter sign block that is not a request"
         assert transaction is None or isinstance(transaction, dict), "Transaction should be a dictionary"
-
+        self.persistence_integrity_check()
         block = self.BLOCK_CLASS.create(transaction, self.persistence, self.my_peer.public_key.key_to_bin(),
                                         link=linked, link_pk=public_key)
         block.sign(self.my_peer.key)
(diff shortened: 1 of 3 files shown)


qstokkink:
@devos50 @xoriole this is kind of an important one, please take your time reviewing this

ipv8/attestation/trustchain/community.py
```diff
@@ -334,9 +333,54 @@ def received_crawl_request(self, source_address, data):
             self.send_crawl_response(blocks[ind], payload.crawl_id, ind + 1, total_count, peer)
         self.logger.info("Sent %d blocks", total_count)
 
+    @synchronized
+    def recover_database(self):
```
xoriole: It is more like sanitization instead of recovering ;)

xoriole: I totally agree on removing the invalid blocks

qstokkink: Good point, I'll rename that.

qstokkink:
@devos50 sure, this will invalidate your reviews though :point_up:

devos50:
I guess you can merge it?

devos50:
@qstokkink please mark it as 'ON HOLD' then, like my PR :)

qstokkink:
@devos50 good point, as this is a bug fix I guess we should merge it.

PR #126, opened 2018-05-07
Reimplement UDP endpoint listener with Twisted DatagramProtocol

ichorid:
Greatly speed-ups everything!
See  Tribler/tribler#2548

### DIFF
--- a/ipv8/messaging/interfaces/udp/endpoint.py
+++ b/ipv8/messaging/interfaces/udp/endpoint.py
@@ -1,223 +1,56 @@
-import errno
-from select import select
-import socket
-import sys
-import threading
-from time import time
-
-from ..endpoint import DataTooBigException, Endpoint, EndpointClosedException, IllegalDestination
-
-if sys.platform == 'win32':
-    SOCKET_BLOCK_ERRORCODE = 10035  # WSAEWOULDBLOCK
-else:
-    SOCKET_BLOCK_ERRORCODE = errno.EWOULDBLOCK
+from twisted.internet import protocol, reactor, error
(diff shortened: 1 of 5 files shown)


qstokkink:
Ok to test

qstokkink:
Ok to test

qstokkink:
Ok to test

ipv8/messaging/interfaces/udp/endpoint.py
```diff
@@ -1,223 +1,59 @@
...
+        self.transport.write(packet, socket_address)
 
     def open(self):
-        """
-        Open this Endpoint for sending and receiving packets.
-        """
         for _ in xrange(10000):
             try:
+                self._listening_port = reactor.listenUDP(self._port, self, self._ip, UDP_MAX_SIZE)
+                #self._listening_port.startListening()
```
qstokkink: Why is this commented out code here?

ipv8/messaging/interfaces/udp/endpoint.py
```diff
@@ -1,223 +1,59 @@
...
     def close(self, timeout=10.0):
-        """
-        Close this Endpoint for sending and receiving.
-        :param timeout: the maximum amount of time to wait for the socket to close
-        :return: whether closing occurred naturally
-        """
         self._running = False
         result = True
+        #FIXME: this thing returns Deferred, need to account for that
+        self._listening_port.stopListening()
```
qstokkink: Why not just return this deferred and get rid of the timeout parameter?

ichorid:
So I just removed the timeout argument completely. Anyways, it was not used anywhere.

PR #127, opened 2018-05-09
Temporary Dispersy compat fix: remove isinstance(EndpointListener).

ichorid:
That's only temporary fix, it should go when we stop using Dispersy as an EndpointListener.


ichorid:
Ok to test

qstokkink:
Ok to test

PR #132, opened 2018-05-14
Switched from sha-512 to sha-256 attribute proofing

qstokkink:
This brings down the attestation time from 10 seconds to 0.7 seconds.


ichorid:
sha-265 is accelerated in modern CPUs by Intel CPU extensions, and sha-512 don't.
Nice optimization!

PR #134, opened 2018-05-15
Add wait for deferred Endpoint.Close on IPv8 teardown

ichorid:
When the endpoint closes, it returns a deferred. There is no unit test in IPv8 for checking proper closing of the endpoint socket, so some tests in Tribler were failing due to `dirty reactor` problem.
This little patch adds wait for socket-closing deferred.


qstokkink:
This is because the endpoint was closed inside Tribler before due to the Dispersy compatibility. You'll need to make changes in Tribler as well.

ichorid:
@qstokkink, I found the relevant lines in Tribler code and changed them accordingly for my next pull request (to Tribler base). This is the kind of trouble one gets when there is no proper isolation, but things are already broken down into submodules...

Would you please merge it?

PR #148, opened 2018-06-08
Implemented basic retry mechanism in UDP endpoint

devos50:
This PR implements a basic message send retry mechanism. This is necessary since we do not catch any `socket.error`, in contrast to Dispersy and the pre-UDP endpoint code. It fixes various socket-related errors we observed in Tribler.

I also noticed that there are no tests for the UDP endpoint so I've written a bunch of them. These tests also cover the retry mechanism I've made.

### DIFF
--- a/ipv8/messaging/interfaces/udp/endpoint.py
+++ b/ipv8/messaging/interfaces/udp/endpoint.py
@@ -1,4 +1,8 @@
+import socket
+
 from twisted.internet import protocol, reactor, error
+from twisted.internet.error import MessageLengthError
+
 from ..endpoint import Endpoint, EndpointClosedException
 
 UDP_MAX_SIZE = 2 ** 16 - 60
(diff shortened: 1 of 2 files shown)


ichorid:
The code is written in a clean and straightforward way. However, our previous endpoint implementation was not using deferreds for this. Instead, it used a queue. This deferred-based approach introduces **packet reordering**. I'm not sure if our current communities would handle it better than just a packet drop, or would handle it at all. 
For my taste, @xoriole proposed a better solution to this problem: just convert the error to warning and log it.

devos50:
I’m also not sure how communities would respond to both packet dropping or reordering. The communities I’ve written should be able to deal with both reordering and drops. But this might not hold for all communities.

Personally, I think simply dropping the packet and logging it is a viable and acceptable solution (for now?). Since the ‘old’ code was retrying to send packets on failures, I made the retry mechanism again. Since we do not run the endpoint on a separate thread anymore, I used Deferreds instead of a queue.

Shall I simplify the code to just drop and log the packet?

ichorid:
Please, simplify the code and add packet drop with logging. Packet drop is a much more frequent event in networking than packet reordering. One packet more, one packet less...

devos50:
@qstokkink is more in favor of a retry mechanism. Let's reach consensus on this when we're back. For now, I will put this PR on hold.

qstokkink:
Also, dropping is a very serious concern in our code. Reordering pretty much does not matter.

PR #174, opened 2018-06-27
Keep track of outgoing/incoming bytes in endpoint

devos50:
We tested the impact of this PR on the performance of the endpoint; impact is negligible.


ipv8/messaging/interfaces/udp/endpoint.py
```diff
@@ -3,6 +3,7 @@
 import socket
 import sys
 
+from math import log
```
qstokkink: !

PR #189, opened 2018-07-09
Various improvements for the attestation community

### LINKED ISSUE #165: Add metadata checkpoint in attestation verification request
Right now attestation verification requests [are immediately answered](https://github.com/Tribler/py-ipv8/blob/master/ipv8/attestation/wallet/community.py#L188).

We should do the same as in the `AttestationRequest` and [allow for metadata to be sent with the request](https://github.com/Tribler/py-ipv8/blob/master/ipv8/attestation/wallet/community.py#L90).

This will require some edits to the wire-format and REST API.

### LINKED ISSUE #225: Expose attesting party in AttestationEndpoint
Right now the attestation endpoint does not show who attested per attribute: https://github.com/Tribler/py-ipv8/blob/master/ipv8/REST/attestation_endpoint.py#L108 .

This can easily be added by also forwarding `base64.b64encode(hashlib.sha1(b.link_public_key).digest())`.

Furthermore, only the last/newest attribute per `base64.b64encode(hashlib.sha1(b.link_public_key).digest())` should be shared.

qstokkink:
Fixes #165 ,  fixes #225.

This can be merged once we need it.

### DIFF
--- a/ipv8/REST/attestation_endpoint.py
+++ b/ipv8/REST/attestation_endpoint.py
@@ -1,3 +1,4 @@
+from hashlib import sha1
 from base64 import b64decode, b64encode
 import json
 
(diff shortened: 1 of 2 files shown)


ipv8/REST/attestation_endpoint.py
```diff
@@ -55,6 +58,18 @@ def on_attestation_complete(self, for_peer, attribute_name, attribute_hash, from
             self.identity_overlay.add_known_hash(attribute_hash, attribute_name, for_peer.public_key.key_to_bin(),
                                                  metadata)
 
+    @inlineCallbacks
```
devos50: Making this method an `inlineCallback` seems a bit unnecessary to me. I would rather just return the deferred.

qstokkink: This is consistent with the rest of the code. We can make enhancements like that later in a different PR.

devos50: Ok then!

PR #190, opened 2018-07-10
ON HOLD: Added tracker plugin

qstokkink:
This PR implements a tracker plugin which **does not** handle the old DiscoveryCommunity introduction request overrides.


qstokkink:
retest this please

PR #197, opened 2018-07-11
ON HOLD: Added XPR serializer + test

devos50:
In this PR, I've added an XPR serialiser to each payload. I also added two test, that sends 10000 messages between (local) communities. One test serializes packets using Structs and the other test serializes it with XDR. Each payload sent contains 5 strings (with random length), 5 integer values and 5 floating point values.

Preliminary results (by running the tests included in this PR) show that **XDR is ~20% slower than Struct serialization**. However, this might be to my first (likely inefficient) implementation. I will leave the PR open until we have decided whether we want this in IPv8 or not. I can then finish the PR.

### DIFF
--- a/ipv8/messaging/serialization.py
+++ b/ipv8/messaging/serialization.py
@@ -1,4 +1,5 @@
 import abc
+import xdrlib
 from struct import pack, unpack, unpack_from, Struct
 import sys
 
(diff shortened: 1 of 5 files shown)


ichorid:
There's no surprise that Python structs are faster than XDRlib. If you'll look into it's implementation in Python, there is only 250 lines of pure Python code. I.e., there is not pre-compiled C lib, it's just Python interpreter and structs under the hood.
The main advantages of XDR are simplicity and portability. In time, if we are really concerned with XDR speed, we can trivially produce/port our own C-based implementation of XDR.

ipv8/test/util.py
```diff
@@ -166,4 +166,4 @@ def twisted_wrapper(arg):
     """
     if isinstance(arg, (int, long)):
         return lambda x: deferred(arg)(inlineCallbacks(x))
-    return deferred(timeout=1)(inlineCallbacks(arg))
+    return deferred(timeout=10000)(inlineCallbacks(arg))
```
qstokkink: 👎 🤢

devos50: I accidentally checked this in, removed it from the commit.

devos50:
@ichorid I'm neutral in this. I leave the decision what to do and which serializer to use up to you and @qstokkink.

qstokkink:
@ichorid correct me if I'm wrong, but the data will already come out of the database as an XDR object anyway, right? In that case we do not need to serialize it again.

ichorid:
@qstokkink, in the current design, metadata is stored twice: once in the form of rows/columns in SQLite, and once as a torrent of blobs of serialized entries. I've considered storing blobs in SQLite to save some cycles by omitting serialization but abandoned the idea. Typically, we don't do bulk transfers through communities (i.e., gossip), and for bulk transfers with libtorrent we already have the stuff pre-serialized on disk.

qstokkink:
@ichorid great, if we have the serialized ball anyway we don't need to do this again on the community level. We can just send the pre-serialized ball in the overlay message.

qstokkink:
It doesn't seem we will be needing this.

PR #202, opened 2018-07-20
WIP: Raise exception when cursor cannot be acquired

qstokkink:


qstokkink:
This is not consistent with what the sql library does.

PR #203, opened 2018-07-20
Toned down default crawl response blocks count

qstokkink:
Related to https://github.com/Tribler/tribler/issues/3751


qstokkink:
retest this please

PR #209, opened 2018-08-07
Added community statistics

xoriole:
Added community statistics

### DIFF
--- a/ipv8/deprecated/community.py
+++ b/ipv8/deprecated/community.py
@@ -7,16 +7,15 @@ Community instance.
 @organization: Technical University Delft
 @contact: dispersy@frayja.com
 """
+import sys
 from random import choice, random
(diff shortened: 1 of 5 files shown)


ipv8/deprecated/community.py
```diff
@@ -105,6 +105,20 @@ def __init__(self, my_peer, endpoint, network):
             chr(235): "missing-last-message"
         }
 
+    def get_prefix(self):
+        return self._prefix
+
+    def enable_statistics(self, enable):
```
qstokkink: This should be on the IPv8 level, not on the community. Please remove the changes from `community.py` and instead intialize your endpoint with the IPv8 instance. For example:

```python
if enable_statistics:
    self.endpoint = StatisticsEndpoint(self, self.endpoint)
```

PR #210, opened 2018-08-08
Use new key in whitewashing attack

qstokkink:
The existing whitewashing attack was not good enough: Trustchain detected it as double spending. It turns out we need an entirely new key to subvert the detection mechanism. Good for Trustchain, bad for the attacker.


devos50:
I think I'm missing some context here. What 'existing whitewashing' attack do you mean?

qstokkink:
@devos50 the REST API exposes a call to perform a whitewashing attack on the system (called `drop_identity`). This is used for testing. However, the current implementation still gets flagged by Trustchain as a double spend, even though the server has removed the blocks as well, through the random block propagation in the network. To do the attack properly the user needs an entirely new key.

PR #223, opened 2018-08-14
Deprecated _ez_unpack_* function calls swap with  lazy_wrapper decorators

DanGraur:
Removed the calls to the `_ez_unpack_*` methods, and added calls to the `lazy_wrapper` decorators in lieu. Also refactored the deprecated/community.py module in order to move the `lazy_wrapper` decorators and the `EZPackOverlay` class to a new module:  deprecated/lazy_community.py.

### DIFF
--- a/ipv8/attestation/trustchain/community.py
+++ b/ipv8/attestation/trustchain/community.py
@@ -14,6 +14,7 @@ from .block import TrustChainBlock, ValidationResult, EMPTY_PK, GENESIS_SEQ, UNK
 from .caches import CrawlRequestCache, HalfBlockSignCache, IntroCrawlTimeout
 from .database import TrustChainDB
 from ...deprecated.community import Community
+from ...deprecated.lazy_community import lazy_wrapper, lazy_wrapper_unsigned, lazy_wrapper_unsigned_wd
 from ...deprecated.payload import IntroductionResponsePayload
 from ...deprecated.payload_headers import BinMemberAuthenticationPayload, GlobalTimeDistributionPayload
 from .payload import *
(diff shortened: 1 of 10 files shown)


devos50:
ok to test

DanGraur:
Seems like some unit tests fail. I'll get back to fixing them tomorrow.

DanGraur:
There's still some Pylint errors left, but I think most, if not all, are from the older code. I can try to remove some of them if you'd like.

qstokkink:
Other than the merge conflict, seems good. Once you fixed the merge conflict could you try this in Tribler and see what breaks?

DanGraur:
Sure!

DanGraur:
I looked into using this new version into Tribler. I ran all the unit tests in the Tribler package, and found that only those in the [`TestMarketEndpoint`](https://github.com/Tribler/tribler/blob/0537ae0d4ff81c54b12ba2ffece13fe1efd5bd54/Tribler/Test/Core/Modules/RestApi/test_market_endpoint.py#L24) class fail. However, I believe this to be a separate issue of that which refers to removing the `_ez_unpack_*` calls, and substituting them with `@lazy_wrapper` decorators, since all of them throw an error due to not being able to find the key `dht` in the `kwargs` dictionary when creating a `MockIPv8` object: in the Tribler version, IPv8 uses [this](https://github.com/Tribler/py-ipv8/blob/9ad0b85c28454a0f4fbae2eb01f34bc7410538ed/ipv8/test/mocking/ipv8.py#L30) type of object initialization while in this branch's version, IPv8 uses [this](https://github.com/DanGraur/py-ipv8/blob/e09110f13fb471be78b6bc795a1dfcbfb79cdc9e/ipv8/test/mocking/ipv8.py#L24) type of object initialization.

qstokkink:
Alright, that is no problem. So, is this PR `READY` now?

DanGraur:
Yes

PR #227, opened 2018-08-15
Optionally load a DHT community in tests

devos50:
Similar to how a TrustChain community can be side-loaded during the tests.


devos50:
@qstokkink could you review this PR again (and merge it if it's ok?) 👍

PR #229, opened 2018-08-17
DO NOT MERGE: Updated IPv8 to Python 3

devos50:
Proof-of-concept. I'm not sure whether we want such an invasive change but we can use it as a reference when upgrading to Python 3 eventually.

I've skipped some unit tests that were failing due to unknown reasons. In particular, there seems to be an error in the attestation crypto which must be fixed before even thinking about merging this.

### DIFF
--- a/ipv8/REST/attestation_endpoint.py
+++ b/ipv8/REST/attestation_endpoint.py
@@ -102,17 +102,17 @@ class AttestationEndpoint(resource.Resource):
             return ""
         if request.args['type'][0] == 'outstanding':
             formatted = []
-            for k, v in self.attestation_requests.iteritems():
+            for k, v in self.attestation_requests.items():
                 formatted.append(k + (v[1], ))
             return json.dumps(formatted)
         if request.args['type'][0] == 'outstanding_verify':
             formatted = []
-            for k, v in self.verify_requests.iteritems():
+            for k, v in self.verify_requests.items():
                 formatted.append(k + (v[1], ))
             return json.dumps(formatted)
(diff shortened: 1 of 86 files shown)


qstokkink:
We should investigate the use of the `__future__` library (https://docs.python.org/2/library/__future__.html#module-__future__) to maintain both `2.7` and `3.x` compatibility.

qstokkink:
This is probably what is causing the attestation crypto to die: https://docs.python.org/3/howto/pyporting.html#division .

Long story short: integer division is now automatically turned into float division in Python 3. You need `//` instead of `/` to do integer division (and override `__floordiv__` instead of `__truediv__`).

devos50:
I changed that one already but it turned out to not fix it :/

qstokkink:
Hmm.. maybe we should go file-by-file. Then we can do the low-hanging fruit early and give the hard files some extra love.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
I added `pylint --py3k` to the PR checks.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
I'll port IPv8 file-by-file by hand. Let's keep this one open for inspiration though 👍

devos50:
IPv8 is now Python3-compatible 👍

PR #230, opened 2018-08-17
ON HOLD: Not spawning our custom reactor thread in tests

devos50:
Twisted's testing framework (trial) should already do this for us.


qstokkink:
Already merged in another PR.

PR #233, opened 2018-08-21
Added a test suite for the REST API, which contains unit tests for th…

DanGraur:
…e various REST requests, a set of dummy peers which can be used for testing, and an interface for different types of inter-peer communication which should facilitate decoupling between the implementation of the communication, and its usage. Currently, the interface has been implemented in the test suite for the HTTP type requests.

### DIFF
--- a/ipv8/REST/attestation_endpoint.py
+++ b/ipv8/REST/attestation_endpoint.py
@@ -43,7 +43,8 @@ class AttestationEndpoint(resource.Resource):
         Return the measurement of an attribute for a certain peer.
         """
         deferred = Deferred()
-        self.attestation_requests[(b64encode(peer.mid), attribute_name)] = (deferred, b64encode(json.dumps(metadata)))
+        self.attestation_requests[(b64encode(peer.mid), attribute_name)] = \
+            (deferred, b64encode(json.dumps(metadata).encode('utf-8')))
         self.attestation_metadata[(peer, attribute_name)] = metadata
         out = yield deferred
         returnValue(out)
@@ -69,7 +70,7 @@ class AttestationEndpoint(resource.Resource):
         block = self.identity_overlay.get_attestation_by_hash(attribute_hash)
         if not block:
             returnValue(None)
-        attribute_name = block.transaction["name"]
+        attribute_name = block.transaction[b"name"]
         deferred = Deferred()
         self.verify_requests[(b64encode(peer.mid), attribute_name)] = deferred
         out = yield deferred
@@ -103,29 +104,29 @@ class AttestationEndpoint(resource.Resource):
         type=peers -> [mid_b64]
         type=attributes&mid=mid_b64 -> [(attribute_name, attribute_hash)]
         """
-        if not request.args or 'type' not in request.args:
+        if not request.args or b'type' not in request.args:
             return ""
-        if request.args['type'][0] == 'outstanding':
+        if request.args[b'type'][0] == b'outstanding':
             formatted = []
             for k, v in self.attestation_requests.items():
                 formatted.append(k + (v[1], ))
-            return json.dumps(formatted)
-        if request.args['type'][0] == 'outstanding_verify':
+            return json.dumps([(x.decode('utf-8'), y, z.decode('utf-8')) for x, y, z in formatted]).encode('utf-8')
+        if request.args[b'type'][0] == b'outstanding_verify':
             formatted = []
             for k, v in self.verify_requests.items():
                 formatted.append(k)
-            return json.dumps(formatted)
-        if request.args['type'][0] == 'verification_output':
+            return json.dumps([(x.decode('utf-8'), y) for x, y in formatted]).encode('utf-8')
+        if request.args[b'type'][0] == b'verification_output':
             formatted = {}
             for k, v in self.verification_output.items():
-                formatted[b64encode(k)] = [(b64encode(a), m) for a, m in v]
-            return json.dumps(formatted)
-        if request.args['type'][0] == 'peers':
+                formatted[b64encode(k).decode('utf-8')] = [(b64encode(a).decode('utf-8'), m) for a, m in v]
+            return json.dumps(formatted).encode('utf-8')
+        if request.args[b'type'][0] == b'peers':
             peers = self.session.network.get_peers_for_service(self.identity_overlay.master_peer.mid)
-            return json.dumps([b64encode(p.mid) for p in peers])
-        if request.args['type'][0] == 'attributes':
-            if 'mid' in request.args:
-                mid_b64 = request.args['mid'][0]
+            return json.dumps([b64encode(p.mid).decode('utf-8') for p in peers]).encode('utf-8')
+        if request.args[b'type'][0] == b'attributes':
+            if b'mid' in request.args:
+                mid_b64 = request.args[b'mid'][0]
                 peer = self.get_peer_from_mid(mid_b64)
             else:
                 peer = self.identity_overlay.my_peer
@@ -136,13 +137,14 @@ class AttestationEndpoint(resource.Resource):
                     # Only include identity blocks we initiated (otherwise it would be an attestation)
                     if b.link_sequence_number == UNKNOWN_SEQ:
                         attester = b64encode(sha1(b.link_public_key).digest())
-                        previous = trimmed.get((attester, b.transaction["name"]), None)
+                        previous = trimmed.get((attester, b.transaction[b"name"]), None)
                         if not previous or previous.sequence_number < b.sequence_number:
-                            trimmed[(attester, b.transaction["name"])] = b
-                return json.dumps([(b.transaction["name"], b64encode(b.transaction["hash"]), b.transaction["metadata"],
-                                    b64encode(sha1(b.link_public_key).digest()))
-                                   for b in trimmed.values()])
-        if request.args['type'][0] == 'drop_identity':
+                            trimmed[(attester, b.transaction[b"name"])] = b
+                return json.dumps([(b.transaction[b"name"], b64encode(b.transaction[b"hash"]).decode('utf-8'),
+                                    b.transaction[b"metadata"],
+                                    b64encode(sha1(b.link_public_key).digest()).decode('utf-8'))
+                                   for b in trimmed.values()]).encode('utf-8')
+        if request.args[b'type'][0] == b'drop_identity':
             self.identity_overlay.persistence.execute('DELETE FROM blocks')
             self.identity_overlay.persistence.commit()
             self.attestation_overlay.database.execute('DELETE FROM %s' % self.attestation_overlay.database.db_name)
@@ -151,7 +153,7 @@ class AttestationEndpoint(resource.Resource):
             my_new_peer = Peer(default_eccrypto.generate_key(u"curve25519"))
             self.identity_overlay.my_peer = my_new_peer
             self.attestation_overlay.my_peer = my_new_peer
-        return ""
+        return b""
 
     def render_POST(self, request):
         """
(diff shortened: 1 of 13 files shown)


ipv8/REST/attestation_endpoint.py
```diff
@@ -5,8 +5,8 @@
 from twisted.internet.defer import Deferred, inlineCallbacks, returnValue
 from twisted.web import resource
 
-from ..attestation.identity.community import IdentityCommunity
-from ..attestation.wallet.community import AttestationCommunity
+from ipv8.attestation.identity.community import IdentityCommunity
```
qstokkink: This will break when imported from Tribler. We need relative imports.

DanGraur: Yeah, I think I mentioned something about this on Slack. I used absolute imports instead of relative ones here deliberately. The reason for that is because for some reason, this time round, https://github.com/DanGraur/py-ipv8/blob/da8d4941247ad82905e3006c3e11b3a7606a9063/ipv8/REST/attestation_endpoint.py#L24 and https://github.com/DanGraur/py-ipv8/blob/da8d4941247ad82905e3006c3e11b3a7606a9063/ipv8/REST/attestation_endpoint.py#L25 would both return empty lists, which would break the `AttestationEndpoint` later on. The built-in function `isinstance` checks for the path of the packages and if I remember correctly, due to the relative import in the `attestation_endpoint.py` module for both `IdentityCommunity` and `AttestationCommunity`, the packages would differ when being tested by Jenkins on the Linux platform (I'm not exactly sure, but the packages would have something like `linuxunittest.attestation.identity.community` or something relatively similar). The paths for the classes of the overlays in the `session.overlays` list would remain constant across all of the platforms on which it was being tested (and if I remember correctly, they were the expected ones).

qstokkink: We'll need a solution without absolute imports though. Using absolute imports will simply not run: this is a show stopper.

DanGraur: Ok, I'll look into a solution for this which doesn't require absolute imports.

qstokkink:
https://stackoverflow.com/a/46709896

ipv8/test/REST/rtest/test_rest_api_peer.py
```diff
@@ -0,0 +1,407 @@
...
+from string import hexdigits
+from time import time
+
+from twisted.internet import reactor
+from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.internet.task import deferLater
+from twisted.web import server
+
+from ipv8.attestation.identity.community import IdentityCommunity
+from ipv8.attestation.wallet.community import AttestationCommunity
```
qstokkink: This could be the reason the `isinstance` stuff was screwing up.

DanGraur:
I think this might finally be ready for another review.

ipv8/REST/attestation_endpoint.py
```diff
@@ -21,8 +21,12 @@ class AttestationEndpoint(resource.Resource):
     def __init__(self, session):
         resource.Resource.__init__(self)
         self.session = session
-        attestation_overlays = [overlay for overlay in session.overlays if isinstance(overlay, AttestationCommunity)]
-        identity_overlays = [overlay for overlay in session.overlays if isinstance(overlay, IdentityCommunity)]
+
+        attestation_overlays = [overlay for overlay in session.overlays if overlay.__class__.__name__ ==
```
qstokkink: Could you revert these changes (go back to `isinstance`)? I'll try to find out what is causing it not to work.

DanGraur: Sure!

qstokkink:
Hmm.. it seems like all machines have something on the `sys.path` which shouldn't be there.

Mac:

```python
['/home/jenkins/ci/workspace/ipv8', '/usr/bin', '/home/jenkins/workspace/pers/Cloudomate_test', '/home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux', '/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/usr/lib/python2.7/lib-old', '/usr/lib/python2.7/lib-dynload', '/home/jenkins/.local/lib/python2.7/site-packages', '/usr/local/lib/python2.7/dist-packages', '/usr/local/lib/python2.7/dist-packages/cryptography-2.3.dev1-py2.7-linux-x86_64.egg', '/usr/lib/python2.7/dist-packages', '/usr/lib/python2.7/dist-packages/gtk-2.0', '/usr/lib/pymodules/python2.7', '/usr/lib/python2.7/dist-packages/wx-3.0-gtk3']
```

Linux:

```python
['/Users/tribler/Documents/workspace2/workspace/ipv8', '/usr/local/bin', '/Users/tribler/Documents/workspace2/workspace/ipv8/test_ipv8_PR_unittests_mac', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python27.zip', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/plat-darwin', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/plat-mac', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/plat-mac/lib-scriptpackages', '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-tk', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-old', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload', '/Users/tribler/Library/Python/2.7/lib/python/site-packages', '/Users/tribler/Library/Python/2.7/lib/python/site-packages/cryptography-2.3.dev1-py2.7-macosx-10.10-intel.egg', '/usr/local/lib/python2.7/site-packages', '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/PyObjC', '/Library/Python/2.7/site-packages', '/Library/Python/2.7/site-packages/nose-1.3.7-py2.7.egg', '/Library/Python/2.7/site-packages/qrcode-6.0-py2.7.egg', '/Library/Python/2.7/site-packages/PyInstaller-3.4.dev0+gd45fab4-py2.7.egg', '/Users/tribler/Library/Python/2.7/lib/python/site-packages']
```

Windows gives another more interesting error (the workspace is actually `C:\workspace\jenkins2\workspace\ipv8\test_ipv8_PR_unittests_windows\ipv8`):

```
twisted: INFO: Starting factory <twisted.web.client._HTTP11Clien['C:\\workspace\\jenkins2\\workspace\\ipv8'
```

qstokkink:
But wait, there's apparently more :)

qstokkink:
On the tester:

```
Known overlays:
<class 'ipv8.attestation.wallet.community.AttestationCommunity'> /home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux/ipv8/attestation/wallet/community.pyc AttestationCommunity
<class 'ipv8.attestation.identity.community.IdentityCommunity'> /home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux/ipv8/attestation/identity/community.pyc IdentityCommunity
Checking for:
<class 'test_ipv8_PR_unittests_linux.ipv8.attestation.wallet.community.AttestationCommunity'> /home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux/ipv8/attestation/wallet/community.pyc AttestationCommunity
<class 'test_ipv8_PR_unittests_linux.ipv8.attestation.identity.community.IdentityCommunity'> /home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux/ipv8/attestation/identity/community.pyc IdentityCommunity
```

On my machine (passing):

```
Known overlays:
<class 'ipv8.attestation.wallet.community.AttestationCommunity'> /home/quinten/tribler/py-ipv8/ipv8/attestation/wallet/community.pyc AttestationCommunity
<class 'ipv8.attestation.identity.community.IdentityCommunity'> /home/quinten/tribler/py-ipv8/ipv8/attestation/identity/community.pyc IdentityCommunity
Checking for:
<class 'ipv8.attestation.wallet.community.AttestationCommunity'> /home/quinten/tribler/py-ipv8/ipv8/attestation/wallet/community.pyc AttestationCommunity
<class 'ipv8.attestation.identity.community.IdentityCommunity'> /home/quinten/tribler/py-ipv8/ipv8/attestation/identity/community.pyc IdentityCommunity
```

DanGraur:
That's what I saw as well when testing this. The tester would indeed look into a different module starting with `test_ipv8_PR_unittests_linux` for the instances imported (in a relative fashion in the `attestation_endpoint.py` module). Similarly, when testing on my machine all tests would pass, but would fail when run by Jenkins due to the aforementioned module path difference (in combination with `isinstance`) .

qstokkink:
@DanGraur absolute imports are the devil. I'll try creating an IPv8 object for testing, one moment.

qstokkink:
@DanGraur I created a `TestIPv8` which mimicks the real `IPv8` object. It is, however, completely local and working with memory databases. Could you take my quick fix and polish it up to proper standards?

DanGraur:
Certainly!

DanGraur:
I think this branch is finally ready for a final review.

ipv8/test/REST/test_attestation_endpoint.py
```diff
@@ -0,0 +1,477 @@
...
+            excluded_peer_mids = set()
+        elif isinstance(excluded_peer_mids, list):
+            excluded_peer_mids = set(excluded_peer_mids)
+
+        peer_list = yield self._get_style_requests.make_peers(dict_param)
+        peer_list = set(peer_list)
+
+        # Keep iterating until peer_list is non-empty
+        while not peer_list - excluded_peer_mids:
+            yield deferLater(reactor, 0.1, lambda: None)
```
qstokkink: This is a bit ugly, please refactor this into a `yield self.sleep(0.1)` call (defined in `RESTTestBase`).

ipv8/test/mocking/rest/ipv8.py
```diff
@@ -0,0 +1,63 @@
...
+
+from .comunities import TestIdentityCommunity, TestAttestationCommunity
+from ....keyvault.crypto import ECCrypto
+from ....messaging.interfaces.udp.endpoint import UDPEndpoint
+from ....peer import Peer
+from ....peerdiscovery.discovery import RandomWalk
+from ....peerdiscovery.network import Network
+
+
+class TestIPv8(object):
```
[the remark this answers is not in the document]
DanGraur: Ok. I'll try to change it to `MockIPv8`. If that won't work, then I'll change the name of the class to `TestRestIPv8`.

ipv8/test/mocking/rest/peer_communication.py
```diff
@@ -0,0 +1,113 @@
+
+from abc import abstractmethod
+
+
+class PostStyleRequests(object):
```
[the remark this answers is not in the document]
DanGraur: Certainly. I initially added this class, with the purpose of it functioning like an interface. When I initially wrote it, there was only an `attestation_endpoint` (now there are multiple endpoints), and this class defined an interface for all possible `POST` type requests then. I thought it would be a good idea to add it since it would separate the interface from the implementation, that is to say, there may be another protocol, other than HTTP, which may implement these methods, and still be RESTful, and this protocol may be implemented in IPv8 later on in the future. This interface would be used in asserts, to ensure that the passed object does have these methods (by being an instance of `PostStyleRequests`). The same applies for `GetStyleRequests`, but for `GET` requests. If need be, they can both be removed.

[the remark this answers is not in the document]
DanGraur: Should I remove them? Moreover, if I should remove them, should I then merge `HTTPGetRequester` and `HTTPPostRequester` (i.e. the classes which implement these 2 interfaces) together into a big class which implements  request methods, or should I leave them separated?

ipv8/test/mocking/rest/rest_api_peer.py
```diff
@@ -0,0 +1,324 @@
...
+        Add a new verified peer
+
+        :param peer: the new peer
+        :return: None
+        """
+        self._ipv8.network.add_verified_peer(peer)
+
+    def add_peer_to_all_services(self, peer):
+        """
+        Add a pier to the identity service
```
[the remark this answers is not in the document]
DanGraur: Yes, it is added to all services.

ipv8/test/mocking/rest/rest_api_peer.py
```diff
@@ -0,0 +1,324 @@
...
+
+    @property
+    def port(self):
+        return self._port
+
+    @property
+    def interface(self):
+        return self._interface
+
+    class RestAPITestWrapper(TaskManager):
```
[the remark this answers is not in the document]
DanGraur: Wanted to reduce its scope, since it seems to be very particular to the tests concerning the REST API, hence it only gets used by the `TestPeer` class.

ipv8/test/mocking/rest/ipv8.py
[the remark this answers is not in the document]
DanGraur: I'm not particularly sure if that's possible. Indeed, the classes do share some degree of similarity, but the `TestIPv8` class, written for the REST API, has two significant differences:

* It is not limited to only one overlay (besides the DHT and TrustChain overlays) which is important, since certain endpoints require that the IPv8 object has a number of varying endpoints, in order for them to work. This, I think, increases the versatility of this new IPv8 wrapper and makes it more suitable for testing the REST API. From what I've seen of the other IPv8 wrapper, it is generally used in tests which focus on a particular part of IPv8 (a particular overlay for instance). Endpoints, however, make use of the functionality provided by multiple overlays. That being said, a better way of configuring endpoints is required in this new wrapper. Currently, overlays are hard-coded. I'll work on improving that.  

* The `endpoint` used in the new wrapper is an actual `UDPEndpoint`, which enables HTTP requests to be captured. This time around, peers actually have a local address on which HTTP requests can be received, rather than a bogus one which only complies with the IPv4 format. This should allow the endpoints to be tested against HTTP requests.

DanGraur:
I pushed a new commit which I think addresses most of the issues flagged in the previous code review. That being said, I should make a couple of mentions:
* I did not remove the `TestIPv8` (now renamed to `TestRestIPv8`). I still think this class, while similar to the `MockIPv8` class, has a number of differences which make the two incompatible, or rather, it would be quite messy and one would probably need to use a set of bad programming practices to use  the `MockIPv8` class in the tests involving the REST API. The reasons why I think these classes are dissimilar are listed [here](https://github.com/Tribler/py-ipv8/pull/233#discussion_r226826625).
* I did not remove the `PostStyleRequests` class, and its counterpart `GetStyleRequests` (the two are now renamed `IPostStyleRequests` and `IGetStyleRequests`). The reason for this is that they are used in a number of `assert` statements. These `assert` statements guarantee that the communication objects (those that should implement the construction and transmission of HTTP requests, in this particular case), do indeed implement a set of methods which are central to the test cases (the methods that implement specific requests). Maybe in the future the implementation will be changed to something more efficient, or at least different, but the use of these abstract classes guarantees that the implementing classes fulfill a contract which continues to allow the unit tests to function normally (or even non-test code, since the current implementations actually send proper HTTP requests). What should also be mentioned about these two 'interfaces' is that they currently feature only abstract methods for communicating with the `AttestationEndpoint` (since only this endpoint existed when I initially wrote these classes). As the number of endpoints has since increased, I'll have to make certain changes in this regard, to ensure that certain aspects of the test framework are more flexible and accommodating to such (new) interfaces, which feature contracts for communication with other endpoints. That being said, do let me know if you think I should remove them.

DanGraur:
In this latest commit I also addressed the issues flagged with respect to the `PostStyleRequests` and the `GetStyleRequests` (now named `IPostStyleRequestsAE` and `IGetStyleRequestsAE`). The scope of this 'interfaces' has been reduced to the `AttestationEndpoint` test cases. This should increase the reusability of the framework. I think it's a change for the better.

qstokkink:
@devos50 and @xoriole any further comments?

DanGraur:
I squashed the commits, and addressed the merge conflicts. Also I made a some small changes to the unit tests in the `TestAttestationEndpoint` class. Nothing significant, just to address the changes made in the meantime to the `AttestationEndpoint` class.

PR #245, opened 2018-08-30
Trial testing system test branch

ichorid:
The pull request for testing the Trial testing system.


ichorid:
retest this please

ichorid:
retest this please

ichorid:
retest this please

testlist.py
```diff
@@ -0,0 +1,88 @@
+from unittest import makeSuite, TestSuite
+
+
+
+class ipv8_testsuite(TestSuite):
```
qstokkink: You should be able to use this: https://docs.python.org/2/library/unittest.html#unittest.TestLoader.loadTestsFromModule so you don't have to specify all of these manually.

qstokkink: Then you could also just get rid of this entire file.

run_all_tests.py
```diff
@@ -4,19 +4,40 @@
 from twisted.python.runtime import platformType
 from twisted.scripts.trial import _makeRunner, Options, _initialDebugSetup
 
-from testlist import ipv8_testsuite
+from unittest import TestLoader
+import unittest
+
+def cleanup_testsuite(arg):
```
qstokkink: Usually test suites only include methods which start with `test_`.

ichorid: disable the cleaner and witness the effect :wink:

qstokkink: Why are you not doing the same?

ichorid:
```python
Running 367 tests.
ipv8.test.base
  TestBase
    runTest ...                                                            [OK]
ipv8.test.dht.test_routing
  TestBucket
    test_add_update ...                                                    [OK]
    test_get ...                                                           [OK]
Unhandled Error
  File "/home/vader/.local/lib/python2.7/site-packages/twisted/protocols/amp.py", line 1043, in _commandReceived
    deferred = self.dispatchCommand(box)
  File "/home/vader/.local/lib/python2.7/site-packages/twisted/protocols/amp.py", line 1101, in dispatchCommand
    return maybeDeferred(responder, box)
  File "/home/vader/.local/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/vader/.local/lib/python2.7/site-packages/twisted/protocols/amp.py", line 1188, in doit
    return maybeDeferred(aCallable, **kw).addCallback(
--- <exception caught here> ---
  File "/home/vader/.local/lib/python2.7/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/vader/.local/lib/python2.7/site-packages/twisted/trial/_dist/worker.py", line 77, in addSuccess
    self._result.addSuccess(self._testCase)
  File "/home/vader/.local/lib/python2.7/site-packages/twisted/trial/_dist/distreporter.py", line 83, in addSuccess
    self.running[test.id()].append((self.original.addSuccess, test))
exceptions.KeyError: 'ipv8.test.base.TestBase.runTest'

                                                        [FAIL]
```

That's why.

ichorid: Or what was your question? Why don't I check for "test_" there?

qstokkink: Yes

ichorid: Because detecting something based on names is not reliable. It is more reliable to mark the class we want to exclude by adding some property/method.

ichorid: Name-based detection is the reason why we even got this problem in the first place.

ichorid:
Let's just merge it directly into "Master" branch when the Tribler PR will be ready, shall we?

run_all_tests.py
qstokkink: I'm not sure if we're talking about the same thing. What I'm trying to say is that the default behavior of all available test suites is to only `addTest()` methods from test classes which start with `test_`. This is what programmers expect. What is unexpected is methods like `runTest()` are being executed.

ichorid: Yes, it is unexpected. It is a `unittests` bug. However, I don't like the name-based detection method in the first place. I think it is bad design. I would expect this method to be used in `bash`-style scripting languages, and I see no reasons why any language with proper module import system, like Python, would use this.
That's why I didn't use the text-based detection in my fix. Because I think this kind of design is wrong.
However, it's your code. If you insist that the cleaner's detection method should be implemented with text-scanning, I can change it.

qstokkink:
```
===============================================================================
[FAIL]
*--- Failure #6 ---
/home/quinten/.local/lib/python2.7/site-packages/twisted/internet/defer.py:1418: _inlineCallbacks(...)
/home/quinten/tribler/py-ipv8/ipv8/test/messaging/interfaces/udp/test_endpoint.py:67: test_send_many_messages(...)
/home/quinten/.local/lib/python2.7/site-packages/twisted/trial/_synctest.py:432: assertEqual(...)
/usr/lib/python2.7/unittest/case.py:513: assertEqual(...)
/usr/lib/python2.7/unittest/case.py:506: _baseAssertEqual(...)
 [Capture of Locals and Globals disabled (use captureVars=True)]
twisted.trial.unittest.FailTest: 0 != 50
*--- End of Failure #6 ---

ipv8.test.messaging.interfaces.udp.test_endpoint.TestUDPEndpoint.test_send_many_messages
-------------------------------------------------------------------------------
```

ichorid:
@qstokkink , you got this error on the same branch, without rebasing it on top of `master`?

qstokkink:
@ichorid I checked out your branch without any modifications.

Also, how can you get logging statements on failure? I tried to patch that in myself, but it seems like trial just magically eats logging output whatever I do.

ichorid:
That's... strange. I debugged trial all the way by just doing "trial > ./out" and it put everything there. Or do you mean some test-code-produced logging statements?

ichorid:
Tests pass without error on both my home and work machines and on Jenkins test machines.
What version of Twisted do you use?

qstokkink:
@ichorid I'm using `Version('Twisted', 18, 7, 0)`.

Also yes, normal logging output. I've tested by adding this to a random test:

```python
import logging
logging.debug("Some very important debug message.")
raise RuntimeError('hi!')
```

Which only gives me the stacktrace for the `RuntimeError` and not the very important debug message.

ichorid:
We've hit a bug in the Twisted poll/epoll reactors. It's hard to trigger, but Egbert's DHT manages to do that...
When a delayed call is cancelled, its `cancelled` attribute is set to `1`. At the next iteration of the reactor loop, this marked call should be removed. The thing is, the implemen
_pendingTimedCalls
`mainLoop`:
```python

                    # Advance simulation time in delayed event
                    # processors.
                    self.runUntilCurrent()
                    t2 = self.timeout()
                    t = self.running and t2
                    self.doIteration(t)
```

`runUntilCurrent()` :
```python
 while self._pendingTimedCalls and (self._pendingTimedCalls[0].time <= now):
            call = heappop(self._pendingTimedCalls)
            if call.cancelled:
                self._cancellations-=1
                continue

            if call.delayed_time > 0:
                call.activate_delay()
                heappush(self._pendingTimedCalls, call)
                continue

            try:
                call.called = 1
                call.func(*call.args, **call.kw)
```

`timeout`:
```python
        delay = self._pendingTimedCalls[0].time - self.seconds()
```

`_pendingTimeCalls` is a (list)heap of delayed calls, sorted on time remaining to a call.

Here is what happens:
when something like `TaskManager` cancels **all** the calls in the `_pendingTimedCalls` list, the reactor loop *first* gets the timeout of the topmost call in the heap, and *then* proceeds into `doIteration` method that, for `*poll` based reactors just calls `select` with this timeout! If the topmost item in the heap was scheduled to be run in X seconds - the whole thing is now stuck in the `select` for X seconds.

This only happens with `select`-based reactors on Unix. There are two workarounds for this:
1. Change the reactor to GTK-something
2. Add and cancel a dummy 0.1-second delayed call with TaskManager each time we cancel something, so it will end up on top of the calls heap.
 
Demonstration of method 2 on a somewhat mutilated DHT test:
```python
   def test_find_nodes(self):
        d = self.introduce_nodes()
        nodes = d.addCallback(lambda _: self.nodes[0].overlay.find_nodes(self.key))
        from twisted.internet import reactor

        def nothing(_):
            pass
        reactor.callLater(0.1, nothing , None)
        return d
```

qstokkink:
retest this please

ichorid:
@qstokkink, I've merged your diff with the reporter and test detection changes. However, I don't see the report style being changed... The error-debug info is just dumped to the stderr, so I guess we're fine for now.

ichorid:
The tests runtime for different configurations of reactors and parallel execution options:
(note that with -jX stdout is swallowed by test runners)

|  | epoll | poll | select | glib2 | gtk2 | gtk3 | gi |
|-----------------|-------|------|--------|-------|------|------|------|
| Sequential | 46s | 46s | 46s | 9s | 8.7s | fail | fail |
| Parallel (-j12) | 5s | 5s | 5s | fail | 0.8s | 0.8s | fail |

(Ubuntu 18.04 6 core 3.6 GHz Xeon W-2133.)

So, I guess the best thing to do would be to run Linux tests with `--reactor=gtk2` in sequential mode.
And it is still possible to do `-jX` for fast checks on developers' PCs.

qstokkink:
Why do the other reactors fail?

ichorid:
I dunno :man_shrugging: 
You can easily check yourself if you're interested. I'm more surprised that some of them actually *work*.

ichorid:
retest this please

ichorid:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

ichorid:
retest this please

qstokkink:
The windows tester seems to be broken, investigating.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

ichorid:
@qstokkink , we have a problem: jenkins-run test on Linux errors out with "reactor already installed". When I run the same command on the same machine from ssh session - it works. It's complete and utter craziness. That's why I'am adding some garbage "stupid test" commits - to test Jenkins.

qstokkink:
@ichorid ah I see, that's why it doesn't fail the check if the tests failed

ichorid:
retest this please

ichorid:
The problem is Twisted loads IPv8 plugin when running through Jenkins. And the plugin loads reactor on its own.
The plugin is not loaded when I run the command by hand. WTF?

qstokkink:
Are you by chance running the main folder instead of the `ipv8/test` folder?

ichorid:
yes, both on jenkins and from console

ichorid:
retest this please

ichorid:
We've [hit this thing](https://www.chiark.greenend.org.uk/~cjwatson/blog/mysterious-bug-with-twisted-plugins.html).
" ... `twisted.application.reactors.installReactor` calls `twisted.plugin.getPlugins`, so the very start of this bootstrapping sequence is going to involve loading all plugins found on the module path (I assume it’s possible to write a plugin that adds an alternative reactor implementation). ... "

ichorid:
retest this please

ichorid:
retest this please

qstokkink:
I'm still not seeing the logging output on failure:

```
===============================================================================
[ERROR]
*--- Failure #8 ---
/home/quinten/.local/lib/python2.7/site-packages/twisted/internet/defer.py:1418: _inlineCallbacks(...)
/home/quinten/Documents/tribler/Tribler/pyipv8/ipv8/test/attestation/trustchain/test_community.py:64: test_sign_half_block(...)
 [Capture of Locals and Globals disabled (use captureVars=True)]
exceptions.RuntimeError: hello there!
*--- End of Failure #8 ---

ipv8.test.attestation.trustchain.test_community.TestTrustChainCommunity.test_sign_half_block
-------------------------------------------------------------------------------
Ran 147 tests in 13.980s

FAILED (errors=1, successes=146)
quinten@quinten-VirtualBox:~/Documents/tribler/Tribler/pyipv8$ echo $?
0
```

Also, the exit status of the script is 0 on failure.

ichorid:
retest this please

ipv8/test/base.py
```diff
@@ -52,12 +52,12 @@ def setUp(self):
 
     def tearDown(self):
         try:
-            super(TestBase, self).tearDown()
             for node in self.nodes:
                 node.unload()
             internet.clear()
         finally:
             shutil.rmtree("temp", ignore_errors=True)
+        super(TestBase, self).tearDown()
```
qstokkink: This should be in the `try`.

ichorid: Why?

qstokkink: My original argument three days ago was that `rmtree` can take a while to complete, so the test class should not wait for it.

After some reflection time, we might as well leave this as is. Then we are at least sure one temp dir is gone before the next test case starts.

ipv8/test/base.py
```diff
@@ -144,6 +144,4 @@ def introduce_nodes(self):
         yield self.deliver_messages()
 
     def temporary_directory(self):
-        d = os.path.join("temp", self.__class__.__name__ + str(int(time.time())))
-        os.makedirs(d)
-        return d
+        return mkdtemp()
```
qstokkink: This is the root cause of a lot of problems, you do not have permissions on the temp folder of a lot of operating systems. `mkdtemp` is horribly broken and should never be used anywhere.

qstokkink: Previous version is the only one that actually works on Windows.

ichorid: It fails on Linux, though, when run in parallel. Because it deletes the whole "temp" subtree, that could be used by other tests at that moment. Also, the combination of class and timestamp are not random enough and could result in subtle errors when tests run in parallel.

ichorid:
retest this please

ichorid:
It's rebased and ready for merge.

qstokkink:
There are errors in your linux tests:
https://jenkins-ci.tribler.org/job/ipv8/job/test_ipv8_PR_unittests_linux/335/console

ichorid:
oops

ichorid:
it was oops about the merge, and not about the failing tests

ichorid:
Basically, the tests are not failing. This errors are by-products of stuff like crawl requests. They were always there, they just were not reported. And they would always be there unless we disable all background activity of communities for the duration of tests.

qstokkink:
We can only test the stability of tests using a stable testsuite.

Conversely, we can only test the stability of a new test suite using stable tests.

If the old tests are producing uncaught errors somewhere, please point us to them so we can fix them. Then, using the stable tests, we can show that your new test suite is stable.

PR #186, opened 2018-07-06
ON HOLD: Gossip overlay

qstokkink:
Primitive for gossip.

PR #246, opened 2018-08-30
Gossip Overlay 

DanGraur:
Further development of PR #186.

### DIFF
--- a/ipv8/gossip/community.py
+++ b/ipv8/gossip/community.py
@@ -1,5 +1,6 @@
 from abc import abstractmethod
-from random import choice, randint
+from random import sample, choice
+from string import hexdigits
 
 from twisted.internet import reactor
 from twisted.internet.base import DelayedCall
@@ -37,7 +38,7 @@ class GossipRule(object):
 
 
 class GossipMessage(Serializable):
-    format_list = ['64s', '74s', 'c', '74s', 'raw']
+    format_list = ['64s', '74s', 'B', '74s', 'raw']
 
     def __init__(self, my_private_key, rule, target_public_key, payload):
         self.public_key = my_private_key.pub().key_to_bin()
@@ -52,7 +53,7 @@ class GossipMessage(Serializable):
         return [
             ('64s', self.signature),
             ('74s', self.public_key),
-            ('c', str(self.rule)),
+            ('B', self.rule),
             ('74s', self.target_public_key),
             ('raw', self.payload)
         ]
@@ -61,7 +62,7 @@ class GossipMessage(Serializable):
     def from_unpack_list(cls, signature, public_key, rule, target_public_key, payload):
         key = LibNaCLPK(binarykey=public_key[10:])
 
-        if key.verify(signature, public_key + rule + target_public_key + payload):
+        if key.verify(signature, public_key + str(rule) + target_public_key + payload):
             out = object.__new__(GossipMessage)
             out.public_key = public_key
             out.signature = signature
@@ -165,6 +166,9 @@ class GossipOverlay(Overlay):
         self.listeners = []
         self.prefix = "\x01\x00GossipCommunity\x00\x00\x00\x00\x00"
 
+        # Avoid conflicts when it comes to naming voting tasks
+        self._vote_ballot_names = set()
+
         # Loops on this, and periodically sends a message based on the rule set it's got
         self.update_list = []
         self.register_task("update_key", LoopingCall(self.take_step)).start(loop_interval)
@@ -213,7 +217,7 @@ class GossipOverlay(Overlay):
                                    bloomfilter.bytes)
         elif rule == GossipRule.SPREAD:
             # Send a random message from the DB associated to the public_key peer
-            self.send_to_neighbors(rule, public_key, choice(self.message_db[public_key]))
+            self.send_to_neighbors(rule, public_key, sample(self.message_db[public_key], 1)[0])
 
     def add_listener(self, listener):
         """
@@ -356,14 +360,9 @@ class GossipOverlay(Overlay):
         if rule == GossipRule.SUPPRESS:
             # If this peer is suppressed, ignore all messages
             return
-        if message.rule == GossipRule.SUPPRESS:
-            # message.rule = the rule which is associated to the message.target and the message itself
-            if is_neighbor:
-                # Our neighbor asked us to ignore someone else
-                self._vote_suppress(message)
-            else:
-                # If I don't trust you, you are not allowed to influence me.
-                pass
+        if message.rule == GossipRule.SUPPRESS and is_neighbor:
+            # Our neighbor asked us to ignore someone else
+            self._vote_suppress(message)
         elif message.rule == GossipRule.DEFAULT or message.rule == GossipRule.SPREAD:
             (signature, data), _ = self.serializer.unpack_multiple(['64s', 'raw'], message.payload)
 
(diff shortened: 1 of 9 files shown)


qstokkink:
Could you give an explanation of what functionality this PR is adding?

DanGraur:
This PR builds on top of the work in PR #186. It adds the `GossipRuleChangeBallot` which implements a means towards majority voting when it comes to changing the Gossip Rule for a peer, and hopefully can prevent abuse against `SUPPRESS` Gossip Rule changes. For a peer to change rule for another peer to `SUPPRESS`, it will need to collect a majority of votes from other peers within a given amount of time before it commits to the rule change. This means that a set of peers independently decide to vote against a peer at relatively the same point in time. Perhaps this is where further development can be done in the future: one peer initiates the vote, and asks its neighbors to vote against a particular peer (they will independently make their decision based on some logic). If they do decide to vote against they inform all their neighbors of this, and if they don't then they simply do nothing. If a peer receives a `SUPPRESS` vote for the first time, then he employs the same behavior as described above (decide to vote and inform neighbors if favorable vote). If you think this is a good idea, then I can try to implement this idea as well.

This PR also addressed a possible issue of force deleting a message stored locally in a peer when receiving a `GossipRule.DEFAULT` message from a non neighbor, as this may have been [previously](https://github.com/qstokkink/py-ipv8/blob/5ea903ad6a35d76aea405c000c31c5359ac8f417/ipv8/gossip/community.py#L163) possible. 

Other changes include: addressed a few issues with the `GossipMessage` packet, as well as added a set of convenience methods for packing and unpacking such messages / packets. Fixed the issue with the non-validating message, by extending the signature size in the Gossip Message packets to 64 as well as generating a `LibNaCLPK` from a public key whose prefix is dropped. Added a Gossip Listener abstract class, which needs to be implemented by the Gossip listeners. Added the `add_listener` convenience method in `GossipOverlay`. Added the `TestGossipOverlay` class for testing the `GossipOverlay`, along with 3 unit tests: for majority voting in the `SUPPRESS` case, message persistence against abuse from non-neighbor, and message removal (when legal) due to non-neighbor message. Also added the `MockGossipOverlayListener` class which implements the `IGossipOverlayListener` interface for testing purposes.

ipv8/gossip/community.py
```diff
@@ -0,0 +1,504 @@
+from abc import abstractmethod
+from random import choice, randint
+
+from twisted.internet import reactor
+from twisted.internet.base import DelayedCall
+from twisted.internet.task import LoopingCall
+
+from ipv8.peer import Peer
```
qstokkink: Please use relative imports.

ipv8/test/gossip/test_community.py
```diff
@@ -0,0 +1,144 @@
+from twisted.internet import reactor
+from twisted.internet.task import deferLater
+
+from ipv8.gossip.community import GossipOverlay, GossipRule, IGossipOverlayListener
```
qstokkink: Relative imports here as well please.

DanGraur:
I removed the absolute imports.

ipv8/gossip/community.py
```diff
@@ -0,0 +1,504 @@
...
+        self.target_public_key = target_public_key
+
+        self.signature = my_private_key.signature(self.public_key + str(self.rule) + self.target_public_key +
+                                                  self.payload)
+
+    def to_pack_list(self):
+        return [
+            ('64s', self.signature),
+            ('74s', self.public_key),
+            ('c', str(self.rule)),
```
qstokkink: Why not make this a `B`? Then you can just put in the `int`, instead of casting the rule to `str` (which is quite fragile and will break horribly if you go over `9`).

ipv8/gossip/community.py
```diff
@@ -0,0 +1,504 @@
...
+        :param action: a function which should be called when the vote passes. It should have the following signature:
+                       (target_public_key : str) => None
+        :return: None
+        """
+        assert message.target_public_key not in self.rule_change_db or \
+               isinstance(self.rule_change_db[message.target_public_key], GossipRuleChangeBallot), \
+            "The contents of the rule_change_db for this target_public_key are not correct."
+
+        if self.rule_change_db.get(message.target_public_key, None) is None:
+            callback_id = reactor.callLater(VOTE_TIMEOUT, self._reset_vote, message.target_public_key)
```
qstokkink: Please use a deferLater which is properly `register_task`'d.

ipv8/test/gossip/test_community.py
```diff
@@ -0,0 +1,144 @@
...
+        Test the voting scheme for changing the GossipRule for a particular peer
+
+        :return: None
+        """
+        target_public_key = self.nodes[0].overlay.my_peer.public_key.key_to_bin()
+
+        self.nodes[1].overlay.set_rule(target_public_key, GossipRule.SUPPRESS)
+        self.nodes[2].overlay.set_rule(target_public_key, GossipRule.SUPPRESS)
+
+        yield deferLater(reactor, 1, lambda: None)
```
qstokkink: We have `self.sleep(1)` for this, but requiring a sleep is a sign of a bad threading model in the first place.

DanGraur: The reason for the wait here is due to the `LoopingCall` in the `GossipOverlay`, which fires every `0.5` seconds. I've added a parameter (called `loop_interval`) for this interval which defaults to the aforementioned value. It is possible to avoid using a `deferLater` by simply reducing it to say `0.005` seconds. I can modify the creation of peers such that the `loop_interval` has a small value.

ipv8/gossip/community.py
DanGraur: Most of the examples of the `register_task` calls, which I've seen in the codebase were being used together with a `reactor.callLater`, that's why I opted for this. I'll change it to `deferLater`, but it will require some extra modifications, I think.

ipv8/test/gossip/test_community.py
qstokkink: It is notoriously very difficult to debug looping calls. To solve this, we usually cancel all pending tasks in the overlay of each node/peer (https://github.com/Tribler/py-ipv8/blob/master/ipv8/test/messaging/anonymization/test_community.py#L35) and then invoke the periodic callback manually.

This ensures that you do not depend on sleeps or get very nasty race conditions.

ipv8/gossip/community.py
qstokkink: You can also register task this indeed. It's just that `Deferred`s are easier to handle than `DelayedCall`s.

ipv8/test/gossip/test_community.py
DanGraur: Ah, I see. This makes more sense. I'll opt for this solution then.

qstokkink:
retest this please

qstokkink:
retest this please

DanGraur:
Addressed most of the issues flagged in the code review, bar the observation regarding the `reactor.callLater`.

ipv8/test/gossip/test_community.py
```diff
@@ -0,0 +1,635 @@
...
+
+        # Check that it was removed
+        self.assertFalse(self.nodes[0].overlay.has_message(first_node_pk, message_contents), "The message could not be "
+                                                                                             "deleted from the first "
+                                                                                             "node's DB.")
+
+    @twisted_wrapper
+    def test_peer_rule_DEFAULT_non_neighbor(self):
+        """
+        Test the case in which
```
qstokkink: 🎏

DanGraur:
I think I addressed the issues with the documentation.

qstokkink:
The implementation seems good now. I'll put this PR on hold until we actually have some scientific basis for the algorithms.

qstokkink:
@DanGraur could you fix the `test_classes_list.txt` so we can park this PR in a branch?

DanGraur:
Certainly, I'll do this as soon as possible!

DanGraur:
I think this PR is now ready to be parked.

PR #248, opened 2018-09-04
Added dht/tunnel endpoints

egbertbouman:


qstokkink:
Retest this please

ipv8/dht/community.py
```diff
@@ -350,7 +350,7 @@ def _find(self, target, force_nodes=False):
                          for sender, node in to_puncture.iteritems()]
 
             # Wait for punctures (if any)...
-            yield DeferredList(deferreds)
+            yield DeferredList(deferreds, consumeErrors=True)
```
qstokkink: I believe @ichorid discovered that if you cancel any of the deferreds this is waiting for (for example on shutdown) this will wait forever. Could you test this?

egbertbouman: I've tested this, but I can't seem to reproduce that problem.

Since I'm not providing my own cancel function, the errback should be called with a CancelledError, which would prevent the DeferredList callback from never getting called.

qstokkink: Alright, then that should be fine.

Issue #253, opened 2018-09-10
Taskmanager fails to register task with duplicate name


ichorid:
https://github.com/Tribler/py-ipv8/blob/d89d7bb42e4647cf34f14966b9c3ae0bf1632f8a/ipv8/taskmanager.py#L35-L45

When one tries to add a new task with a name that's already registered, Taskmanager will:
1. Cancel this new task (L43)
2. **Unregister the original task** (L44)

This results in unmanaged delayed calls dangling in the reactor. For example, calling `remove_exit_socket` twice for the same exit socket triggers this behavior.

ichorid:
Gonna fix it in https://github.com/Tribler/tribler/pull/3842, but we have to discuss what kind of behavior is desirable when trying to add a task with duplicate name.

qstokkink:
Please put it in a separate PR. The expected behavior is to error out.

PR #254, opened 2018-09-10
Fix adding task with duplicate name to taskmanager

ichorid:
Fix for https://github.com/Tribler/py-ipv8/issues/253
Raises an error in case of duplicate name and replaces the task with the new one.
The shutdown check is moved to be the 1st one to prevent replacing tasks before the shutdown.

### DIFF
--- a/ipv8/taskmanager.py
+++ b/ipv8/taskmanager.py
@@ -44,11 +44,11 @@ class TaskManager(object):
                     stopfn()
                 return task
 
+            assert isinstance(task, (Deferred, DelayedCall, LoopingCall)), (task, isinstance(task, Deferred))
+
             if self.is_pending_task_active(name):
                 self.replace_task(name, task)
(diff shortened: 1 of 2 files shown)


ipv8/taskmanager.py
```diff
@@ -37,21 +37,20 @@ def register_task(self, name, task, delay=None, value=None, interval=None):
...
             if self._shutdown:
                 self._logger.warning("Not adding task %s due to shutdown!", str(task))
                 is_active, stopfn = self._get_isactive_stopper(task)
                 if is_active and stopfn:
                     stopfn()
                 return task
 
+            if self.is_pending_task_active(name):
+                    raise RuntimeError("Task already exists: '%s'" % name)
+                    self.replace_task(name, task)
```
qstokkink: This is dead code. Was this even tested?

PR #257, opened 2018-09-11
Network now depends on naked mid

### LINKED ISSUE #252: get_walkable_addresses uses a considerable amount of CPU cycles
To address issue #78, I enabled Yappi for the TrustChain crawler and monitored CPU usage over a period of a few hours. See the following report:

yappi_ipv8_report

The `get_walkable_addresses` method takes a significant amount of processing power (38.5%!). A breakdown shows that almost all this time is spent in the `<dictcomp>` method, which in turn has the following performance:

yappi_breakdown

This makes walking through the network expensive in terms of CPU usage. @qstokkink is there any easy way to make this more efficient?

qstokkink:
Fixes #252

### DIFF
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -1,5 +1,3 @@
-from base64 import b64encode
-from hashlib import sha1
 from threading import RLock
 from socket import inet_aton, inet_ntoa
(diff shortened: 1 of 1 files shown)


devos50:
Significant improvement:

utimes

PR #261, opened 2018-09-13
Allow for anonymous tasks in TaskManager

### LINKED ISSUE #260: TaskManager requires unique ids
Right now to register a task with the `TaskManager` you need to supply a unique id.

This makes sense for externally managed tasks. But we see that unmanaged "pop-up" tasks implement some form of random number generation to create a unique id.

We can offer a new method to register task which takes care of creating a unique id. We can simply postfix the user supplied string with the `id()` of the supplied task.

qstokkink:
Fixes #260

### DIFF
--- a/ipv8/taskmanager.py
+++ b/ipv8/taskmanager.py
@@ -68,6 +68,12 @@ class TaskManager(object):
             self._pending_tasks[name] = task
             return task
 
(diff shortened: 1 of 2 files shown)


qstokkink:
retest this please

PR #263, opened 2018-09-13
Refactor TestBase.temporary_directory for parallel execution safety

ichorid:
It is safer to generate dir names in this way. It is important if we want to run tests simultaneously.

### DIFF
--- a/ipv8/test/base.py
+++ b/ipv8/test/base.py
@@ -1,5 +1,7 @@
 import os
+import random
 import shutil
+import string
 import sys
 import threading
 import time
@@ -13,7 +15,6 @@ from twisted.trial import unittest
 from .mocking.endpoint import internet
 from .mocking.ipv8 import MockIPv8
 from ..peer import Peer
-
 
 twisted.internet.base.DelayedCall.debug = True
 
@@ -31,6 +32,7 @@ class TestBase(unittest.TestCase):
         self.nodes = []
         self.overlay_class = object
         internet.clear()
+        self._tempdirs = []
 
     def initialize(self, overlay_class, node_count, *args, **kwargs):
         self.overlay_class = overlay_class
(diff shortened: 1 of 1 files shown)


qstokkink:
@xoriole you are working on the same functionality in https://github.com/Tribler/tribler/pull/3869, could you review?

ipv8/test/base.py
```diff
@@ -144,6 +147,8 @@ def introduce_nodes(self):
         yield self.deliver_messages()
 
     def temporary_directory(self):
-        d = os.path.join("temp", self.__class__.__name__ + str(int(time.time())))
+        rndstr = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
```
qstokkink: Could you use `uuid` for this like in  https://github.com/Tribler/tribler/pull/3869/files#diff-c2974363c4135bb26324c12fb3724032R119 ? Also, please prefix the folder name with `test_` or `temp_` or something (or put all of the dirs inside the `temp` dir like previously).

ichorid: if we put everything inside a single `temp` dir, we return to the same problem - the problem of one common resource shared by each test.

ichorid: and we don't want to use UUIDs because they are too long. On Windows there is a limitation of 255 characters for the whole filesystem *path*.

qstokkink: @xoriole ⏫ 

Alright, could you prefix the folder with something that shows it's a temporary directory though? This makes it easier to delete it, should something go wrong with its deletion.

ichorid: fine, just forgot to add the prefix

qstokkink:
Great, also please rebase this PR on the latest HEAD instead of the ugly merge commit.

ichorid:
here you are, sir...

PR #266, opened 2018-09-13
ON HOLD: Trial testing system: take 2

ichorid:
It is all cleaned up now (continued from https://github.com/Tribler/py-ipv8/pull/245). Only commits relevant to Trial migration here.

### Test output
Basically nose still run the tests, nothing changed on the outside. However, inside the `unittest` object is provided by `twisted.trial` module. 

### Alternative reactor Nose runner
To ensure that the runtime is not increased due to https://github.com/Tribler/py-ipv8/pull/245#issuecomment-418434274 we have to use our own runner for `nosetest` utility to be able to select an alternative reactor. We only do this on Unix systems, though. By default we try to use `glib2reactor` that is stable, does not produce any spurious messages and does not hit https://github.com/Tribler/py-ipv8/pull/245#issuecomment-418434274 problem. If it not available, we fall back to `selectreactor` that is available for all platforms.

### Trial-based runner
In addition, we now have the option to use Trial-based runner, that automatically scans the dirs for tests. It does not collect the logs, though. However, in parallel mode, it runs the whole suite in under 1 second, so it is useful for doing quick checks (only on Unix platforms).
To use it, one must first regenerate the Twisted plugins cache with provided utility (or just run it more than one time).


ipv8/test/messaging/anonymization/test_hiddenservices.py
```diff
@@ -28,6 +27,10 @@ def announce(self, info_hash):
             global_dht_services[info_hash] = [self.address]
 
 
+class Twisted(object):
```
qstokkink: Why do we need this?

regen_plugins_cache.py
```diff
@@ -0,0 +1,2 @@
+from twisted.plugin import IPlugin, getPlugins
```
qstokkink: I take it this is to fix the weird plugin loading thing. Please add comments here so nobody removes it in the future. Also, why not add this to the `__init__.py`?

run_nose.py
```diff
@@ -0,0 +1,14 @@
...
+
+if __name__ == '__main__':
+    try:
+        from twisted.internet import glib2reactor as reactor
+    except ImportError:
+        from twisted.internet import selectreactor as reactor
+
+    reactor.install()
+
+    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
```
qstokkink: Why is this here? Please add some comments.

qstokkink:
Apparently Linux and Windows are running the old test scripts and Mac is running the new one, we need to look at this.

Also, when using `trial` Mac is showing an error. The unit tests for Mac still pass though. For some reason the uncaught exceptions are not being inspected properly.

qstokkink:
We may want to first make sure this PR is merged and passing using the old test scripts, before we start using trial though.

ipv8/test/messaging/anonymization/test_hiddenservices.py
ichorid: the reason for this is too much alt+enter smashing in PyCharm...
Funny, though, that pylint don't catch it.

qstokkink: We run a quick version of Pylint. The older version we run doesn't catch all of the errors, it does finish in 5 seconds instead of 2 minutes though.

ichorid:
Also, please **DON'T** merge it until https://github.com/Tribler/tribler/pull/3842 is ready and approved for merging.

ichorid:
retest this please

ichorid:
retest this please

ipv8/test/mocking/ipv8.py
```diff
@@ -48,9 +48,9 @@ def __init__(self, crypto_curve, overlay_class, create_trustchain=False, create_
             self.endpoint.enable_community_statistics(self.overlay.get_prefix(), True)
 
     def unload(self):
-        self.endpoint.close()
+        self.overlay.unload()
```
devos50: What was the reason for swapping these statements? Since now it could be that there are incoming messages after unloading an overlay right?

run_all_tests_trial.py
```diff
@@ -0,0 +1,36 @@
...
+    config.parseOptions()
+    config['tbformat'] = 'verbose'
+    config['exitfirst'] = True
+
+    _initialDebugSetup(config)
+    trialRunner = _makeRunner(config)
+
+    test_loader = TestLoader()
+    test_loader.suiteClass = CustomSuite
+    test_suite = test_loader.discover('./ipv8/test', top_level_dir='.')
```
devos50: You could consider using `os.path.join` here (which is more compatible across OSses).

run_all_tests_unix.sh
```diff
@@ -9,12 +9,7 @@ tty -s && tput bold
 echo -n "Starting IPv8 testsuite: "
 if nosetests --version >>/dev/null 2>&1; then
     echo "using test runner 'nosetests'!"
-    # Use nosetests2 if available and if not, try version possibly incompatible with python2
-    if [ -x "$(command -v nosetests2)" ]; then
-        test_command="nosetests2"
-    else
-        test_command="nosetests"
-    fi
+    test_command="python run_nose.py"
```
devos50: It's not safe to assume that the `python` command one ones system always points to a Python 2 interpreter. Maybe we should check specifically for `python2` availability first (like we did with `nosetests2`)?

regen_plugins_cache.py
ichorid: Included running this script in `run_all_tests_unix.py`

run_all_tests_unix.sh
ichorid: added a separate check for python interpreter version

run_nose.py
ichorid: umm, what's unclear?

ichorid: Oh, that's the old comment.

run_all_tests_trial.py
ichorid: did that

ipv8/test/mocking/ipv8.py
ichorid: The reason was `None objects` kind of errors appearing in tests. At init, we first load the overlay, and then open the endpoint. Therefore, we *must* do it the other way around at shutdown, or the overlay tries to do stuff with the endpoint, which is not there...
The problem is, the overlay does not check if the endpoint object is still there.
For now its fine, no errors after this fix. But we'll have to clean up the init-shutdown sequences of overlays in the future.

ichorid:
retest this please

ipv8/test/test_taskmanager.py
```diff
@@ -34,22 +33,22 @@ def tearDown(self):
         self.tm.shutdown_task_manager()
         super(TestTaskManager, self).tearDown()
 
-    @twisted_wrapper
+    @inlineCallbacks
     @untwisted_wrapper
```
qstokkink: Do we still need the `inlineCallbacks` and `untwisted_wrapper` with trial?

run_all_tests_trial.py
qstokkink: `abspath` is not the same as `os.path.join`. `join` takes care of the OS dependent separator (`/`, which on Windows is `\`). The fix would be:

```python
os.path.join('.', 'ipv8', 'test')
```

ipv8/test/test_taskmanager.py
ichorid: I don't know.
Like, I really tried to minimize the invasiveness of the changes. The plan was to remove the cruft after this PR is merged and we already have working Trial-based tests.

qstokkink: Fair enough.

ichorid:
rebased on top of latest `master`.
@qstokkink, time to merge it!

ichorid:
hallelujah! :angel:

PR #267, opened 2018-09-14
Correct shutdown order for mock IPv8

ichorid:
On creating a mock IPv8 instance we *first* create the node and *second* open the endpoint. The teardown process should be done in reverse order.
This patch fixes a lot of "unclean reactor" errors produced by Tribler tests, when using Trial-based `unittest` module.


qstokkink:
Would this be a problem in the wild? Right now we are first closing the endpoint and then unloading the IPv8 communities in LaunchManyCore: https://github.com/Tribler/tribler/blob/next/Tribler/Core/APIImplementation/LaunchManyCore.py#L934-L950

PR #282, opened 2018-09-18
Various improvements to TrustChain

devos50:
In this PR, I made quite a few changes to the TrustChain code and protocol. It includes:
- A fix for the block broadcast mechanism, where each block was unintentionally pushed to 1110 users. When this PR is merged, it is pushed to _only_ 25 users.
- An extension of the introduction request/response, where information about the length of ones chain is included (sequence number of their latest block).
- A separate file to store various important TrustChain settings. This includes settings for the block validation ranges (how many blocks do I need prior to signing one) and maximum number of blocks in the database.
- Made the database upgrade script idempotent.
- A database pruning loop that removes old blocks when the database grows too big.
- Made the crawler less aggressive when we already have over 1000 blocks of a specific user.
- A refactor of the crawl mechanism. In particular, each crawl request now contains a 'begin' and 'end' sequence number, or a range. I've also added an additional message when there are no blocks to return. This is more clean than the current mechanism where we return an invalid block in the `send_crawl_response` with type `noblocks`.
- Storing double spend attempts in a separate database table.
- Removed most of the `introduce_nodes` invocations from the TrustChain tests. Peer introduction is already done when creating the node itself.
- Refactored `test_intro_response_crawl` test to make it more simple.
- Added a separate 'crawler' mode to TrustChain, where a node attempts to crawl the whole chain of a target peer.

PR #284, opened 2018-09-19
Fixed imports in ipv8_service

devos50:
This fixes compatibility with Tribler.


devos50:
Yes, #282 will take longer than I thought

PR #283, opened 2018-09-18
Ported several more classes to Python 3

qstokkink:
Related to #265.

PR #281, opened 2018-09-18
Ported several classes to Python 3

qstokkink:
Related to #265.

PR #280, opened 2018-09-17
Ported easy files to Py3 compatibility

qstokkink:
Related to #265.

This PR also exposes the moved `StringIO` class.

PR #277, opened 2018-09-17
Easy Python 3 ports

qstokkink:
Related to #265. This PR ports all easy-to-port files to Python 3.

This PR introduces the `xrange` function to `util.py`.

PR #287, opened 2018-09-20
WIP: Finished last Python3 quirks

qstokkink:


qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
Well that job was 💩. Anyway, ready for review now.

devos50:
Nice, if I have time tomorrow, I will go through it 👍

ipv8/attestation/identity/community.py
```diff
@@ -31,7 +33,7 @@ def add_known_hash(self, attribute_hash, name, public_key, metadata=None):
     def get_attestation_by_hash(self, attribute_hash):
         blocks = self.persistence.get_all_blocks()
         for block in blocks:
-            if block.transaction and block.transaction.get("hash", None) == attribute_hash:
+            if block.transaction and block.transaction.get(b"hash", None) == attribute_hash:
```
devos50: Could you elaborate on why keys in a TrustChain transaction became `byte`? I would argue these strings are (always?) human-readable so there is no need to make them byte.

ipv8/attestation/trustchain/block.py
```diff
@@ -44,19 +44,18 @@ def __init__(self, data=None, serializer=Serializer()):
...
                                                                                       data[7], data[8], data[9])
-            if isinstance(self.public_key, database_blob):
-                self.public_key = str(self.public_key)
-            if isinstance(self.link_public_key, database_blob):
-                self.link_public_key = str(self.link_public_key)
-            if isinstance(self.previous_hash, database_blob):
-                self.previous_hash = str(self.previous_hash)
-            if isinstance(self.signature, database_blob):
-                self.signature = str(self.signature)
+            self.type = self.type if isinstance(self.type, bytes) else str(self.type)
```
devos50: RIP to the elegance of TrustChain code :(

ichorid:
We should definitely NOT merge it into 7.1. We should do a separate IPv8 branch `python3` branch, and link `devel` against it after the 7.1 release though. It would provide a very nice kick-start for migration to Python3 by forcing the whole team to implicitly work on it.

ipv8/attestation/trustchain/block.py
qstokkink: By all means, look at the code one line up.

qstokkink:
Everything seems to be fine:

afbeelding

But there was a lock-up in the Market tests on Tribler (see https://github.com/Tribler/tribler/pull/3899) not sure if it is due to my changes.

ipv8/messaging/deprecated/encoding.py
```diff
@@ -93,9 +104,9 @@ def _a_encode_dictionary(values, mapping):
     {'foo':'bar', 'moo':'milk'} --> ['2', 'd', '3', 's', 'foo', '3', 's', 'bar', '3', 's', 'moo', '4', 's', 'milk']
     """
     assert isinstance(values, dict), "VALUE has invalid type: %s" % type(values)
-    encoded = [str(len(values)).encode("UTF-8"), "d"]
+    encoded = [str(len(values)).encode("UTF-8"), b"d"]
     extend = encoded.extend
-    for key, value in sorted(values.items()):
```
qstokkink: Oh crap we need this, but: https://portingguide.readthedocs.io/en/latest/comparisons.html#unorderable-types

qstokkink: @devos50 this is a big problem, TrustChain will never work on Python 3 like this:

The problem:
 - When we load a block from the database we repack it, which will likely alter the ordering of the transaction.
 - We depend on the `sorted()` producing the same sorting of differently typed inputs.
 - Given the previous two points, the signature does not match the newly packed block.

If we do not fix this now, we are guaranteed a hard fork of TrustChain later.

devos50: Right... this is a problem indeed.

In #282, I'm changing quite some stuff regarding TrustChain so something that would fix it (and requires a protocol break) can go in that PR. Note that I'm not doing a full block wipe (yet) but I have this planned for the actual release of Tribler 7.1.

Another problem, though unrelated to this PR, is that encoding/decoding of transactions takes quite some time and this is a good opportunity to explore a faster approach.

devos50: I would prefer to get rid of `encoding.py` since it's slow and yet another serialization protocol we have to support. As far as I know, TrustChain is the only component using functionality in `encoding.py`.

Using the payload system is a good idea: it's well tested and fast. I think however that we have to extend it to support nested dictionaries (like I'm using in the market).

qstokkink:
@devos50 since `b''` is exactly the same as the normal Python 2 string I'm confident that 99% of this PR is perfectly innocent. I would consider the `encoding.py` thing a game-breaker for merge though. It's definitely a good idea to have a talk on monday to figure out what to do.

ipv8/messaging/deprecated/encoding.py
qstokkink: Anyway, it should be pretty easy to make a `Payload` directive for the serializer. I'll try to get that up and running today.

qstokkink: As expected adding `sorted()` here again makes the Tribler PR pass: https://github.com/Tribler/tribler/pull/3899

qstokkink:
retest this please. I updated the Python 3.6 tests to no longer exit-on-failure (removed `-x`).

qstokkink:
Alright, I'll leave the last remaining Python 3 incompatiblity (`sorted()`) for a later PR as the fix would require changes in upstream projects. This one is fully Python 2 *and* Python 3 compatible and ready to merge though.

devos50:
So do we want to merge this and include it in the 7.1 release?

qstokkink:
By leaving in the `sorted()` this PR is no different than #268, #271, #272, #277, #280, #281, #283, #285, #286, it is fully compatible with both Python 2 and Python 3. I see no reason not to merge this.

However, before we release 7.1 #294 and #295 should definitely be merged.

devos50:
That's true; most of the Python 3 changes have already been merged.

Let's discuss #295 tomorrow.

PR #293, opened 2018-09-23
Added support for nested Payloads

qstokkink:
This PR allows you to nest Serializable objects inside of each other. For instance:

```python
class B(Serializable):
    format_list = ["H"]
    def __init__(self, number):
        self.number = number
    def to_pack_list(self):
        return [("H", self.number)]
    @classmethod
    def from_unpack_list(cls, *args):
        return B(*args)

class A(Serializable):
    format_list = [B]
    def __init__(self, b):
        self.b = b
    def to_pack_list(self):
        return [("payload", self.b)]
    @classmethod
    def from_unpack_list(cls, *args):
        return A(*args)

a = A(B(123))
```

You can also chain these even deeper than the single level shown here. Do note that this is the first type of serializable content which differs in format list and pack list (`<class instance>` vs. `"payload"`).

### DIFF
--- a/ipv8/messaging/serialization.py
+++ b/ipv8/messaging/serialization.py
@@ -11,6 +11,77 @@ class PackError(RuntimeError):
     pass
 
 
+class NestedPayload(object):
+    """
+    This is a special type of format. Allowing for nested packing.
+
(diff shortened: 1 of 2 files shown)


devos50:
retest this please

PR #298, opened 2018-09-25
Added custom sorting method for encoding

### LINKED ISSUE #295: Encoding sorted() unportable to Python 3
After #287 we learned that our current strategy for making the `encoding.py` dictionary encoding deterministic, will not work on Python 3. In Python 3 the `sorted()` function can no longer handle different datatypes in the same sortable container. We will need some other mechanism or as @devos50 suggested, move away from `encoding.py` and use something better.

```
  File "/home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux_py36/ipv8/messaging/deprecated/encoding.py", line 109, in _a_encode_dictionary
    for key, value in sorted(values.items()):
TypeError: '<' not supported between instances of 'str' and 'int'
```

qstokkink:
This is a platform agnostic sorting method for mixed-type datastructures, this allows us to port to other languages like Java (or even Python 3).

Fixes #295. Requires a hard reset of the TrustChain database.

afbeelding

### DIFF
--- a/ipv8/messaging/deprecated/encoding.py
+++ b/ipv8/messaging/deprecated/encoding.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 import logging
 from json import dumps
 
+from .sorting import sortable_sort
 from ...util import cast_to_bin, cast_to_long, is_long_or_int, is_unicode, urllib_future
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


qstokkink:
Quick update, also added a unit test for the current vulnerability:

https://github.com/Tribler/py-ipv8/blob/9807457a9a2028fc3dd863163037191b5e007526/ipv8/test/messaging/deprecated/test_sorting.py#L290-L304

PR #304, opened 2018-09-29
Fixed sorted sets not returning as set type

### LINKED ISSUE #303: TrustChain database update script
Due to #298 we need an update script for the TrustChain.

We could investigate if this is even required, as both the old `sorted` and the new sorting provide similiar sortings.

EDIT: **tl;dr** We don't actually *need* to reset. But we might as well, due to other fixes.

qstokkink:
Fixes #303 

As it turns out, none of the existing block types suffered from `sorted` mania. So we don't need a TrustChain reset. 🎉 🎈 

However, after testing Tribler, I noticed that some communities do type checking for "*`set`-in `set`-out*". As there is no ordered set type available (arguably for a good reason), I had to make my own to conform to Tribler's expectations. This `OrderedSet` is a `set` but iterates like a sorted `list`.


qstokkink:
Note: this should be merged before we update the pointer in Tribler.

PR #320, opened 2018-10-06
Various community/tracker fixes

qstokkink:
Fixes:

 - Superfluous comma.
 - `on_introduction_response` passing unused `data`.
 - Tracker plugin not NAT puncturing correctly.


qstokkink:
Small amendment: by using `intro_peer = None`, `create_introduction_response` would use `get_peer_for_introduction` of the parent class, which is the `DiscoveryCommunity`. To fix this, `get_peer_for_introduction` should always be `None`.

PR #332, opened 2018-10-15
ON HOLD: Refactor reactor.callLater usage into TaskManager Clock

qstokkink:
Fixes #330

 - [X] Expose `TaskManager._reactor`.
 - [x] Pass instance of `Clock` into all `TaskManager` subclasses.
 - [x] Time management in `IPv8` instances.

PR #334, opened 2018-10-16
WIP: Fixed leaking callLater calls in TestBase

### LINKED ISSUE #330: Refactor reactor.callLater usage
Right now we allow programmers to use Twisted's `reactor.callLater` directly. As pointed out before (by @ichorid) this blocks the reactor (ex. `select()` or `poll()`) for a certain time period, before the reactor figures out that a call was cancelled. This especially hurts the `DHTCommunity` tests.

What we should do is move all of the `reactor.callLater` calls to a `Clock` object which we can advance manually. We already do this for the `TestTaskManager` tests. We can then manually advance time using the `Clock` object using (what should be) the only measure of time in `IPv8` instances: `IPv8.on_tick()`. The usage can be sketched as follows:

afbeelding

One downside to this approach is that we get coarser time granularity for callbacks and thus batching of calls. If programmers ever start massively scheduling `callLater` calls we will see CPU spikes. Right now this is not the case, we mostly use proper `Deferred` callback chains.
Another downside is that we will now need to pass `Clock` instances to all `TaskManager`s. We _cannot_ make this a singleton as this would violate the ability for `IPv8` instances to coexist in the same interpreter.

qstokkink:
Fixes #330, (supersedes) closes #332.


qstokkink:
Leaving this for now. No solution has been found.

Issue #273, opened 2018-09-16
DHT Looping call for peer churn


qstokkink:
The DHT community has a `ping_all` looping call for peer management. This mechanism should use the appropriate `DiscoveryStrategy` class instead. Perhaps the other `LoopingCall`s can also be moved to this mechanism, but I have not investigated this.

In part this also resolves the testability issue pointed out by @ichorid, that we should be able to advance time manually as much as possible for staging (instead of relying on background tasks managed by Twisted). The `DiscoveryStrategy` class exposes `take_step` which allows us to control when peers are pinged or removed.

This also removes a lot of clutter from the community.

ichorid:
That would be very nice, indeed!

DanGraur:
I've implemented this, and forwarded a PR: #352.

PR #352, opened 2018-11-04
DHT looping call churn

DanGraur:
This is an implementation of the enhancement suggested in #273.

### DIFF
--- a/ipv8/peerdiscovery/churn.py
+++ b/ipv8/peerdiscovery/churn.py
@@ -66,3 +66,23 @@ class RandomChurn(DiscoveryStrategy):
                             self._pinged[peer.address] = time()
                             packet = self.overlay.create_ping()
                             self.overlay.endpoint.send(peer.address, packet)
+
+
+class PingChurn(DiscoveryStrategy):
+
+    def __init__(self, overlay, ping_interval=25):
+        super(PingChurn, self).__init__(overlay)
+        self.ping_interval = ping_interval
+
+    def take_step(self, service_id=None):
+        with self.walk_lock:
+            self.overlay.routing_table.remove_bad_nodes()
+
+            pinged = []
+            now = time()
+            for bucket in self.overlay.routing_table.trie.values():
+                for node in bucket.nodes.values():
+                    if node.last_response + self.ping_interval <= now:
+                        self.overlay.ping(node).addErrback(lambda _: None)
+                        pinged.append(node)
+            return pinged
--- a/ipv8/dht/community.py
+++ b/ipv8/dht/community.py
@@ -14,6 +14,7 @@ from twisted.python.failure import Failure
 
 from ..peer import Peer
 from ..requestcache import RandomNumberCache, RequestCache
+from ..peerdiscovery.churn import PingChurn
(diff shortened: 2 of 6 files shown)


DanGraur:
Ok, so I implemented what was suggested in the aforementioned issue. I initially moved all the behavior from the `ping_all` method to the new `PingChurn` class. However, during testing, I noticed that the `DHTDiscoveryCommunity` had its own `ping_all` method, which would call the `ping_all` in its superclass, i.e. in `DHTCommunity`. Since that wasn't there the behavior of `ping_all` in `DHTDiscoveryCommunity` was faulty, and hence could not pass a unit tests. Since calling the `PingChurn` strategy object stored in the `strategies` list of the `IPv8` object would imply a lot of bad practices, I decided to move `ping_all` back from `PingChurn` to `DHTCommunity` (`PingChurn` will now only call the method, rather than hold its entire logic). I suppose to a certain extent, this may improve reusability, as long as the base `overlay` has an implementation for `ping_all`. Is this ok, or should I find a way in which I actually move the entire behavior of `ping_all` to `take_step` in `PingChurn`? It should be mentioned, that there is no `LoopingCall` for `ping_all`, so the focus point of the ticket has been addressed.

qstokkink:
I like this as a first step, eventually we want to also move the innards of the current `ping_all` to the strategy.

This PR also shows me that we want to do the `configuration.py` differently. Overlays/communities should be able to expose strategy names (with corresponding classes) themselves, instead of having them globally known in `ipv8_service.py`. Your `PingChurn` is very much only for the DHT community. We also have the existing `RandomChurn` which can only be used by the DiscoveryCommunity.

qstokkink:
I moved my own suggestion to #354.

DanGraur:
If there's currently no changes left to be made to this branch, can this be accepted and merged (after I'll rebase)?

ipv8/test/mocking/community.py
```diff
@@ -23,3 +24,21 @@ def __init__(self):
     def bootstrap(self):
         super(MockCommunity, self).bootstrap()
         self.last_bootstrap = 0
+
+
+class MockDHTCommunity(DHTCommunity):
```
qstokkink: It seems like this can be removed now.

ipv8/peerdiscovery/churn.py
```diff
@@ -66,3 +66,22 @@ def take_step(self, service_id=None):
...
+
+
+class PingChurn(DiscoveryStrategy):
+
+    def __init__(self, overlay, ping_interval=25):
+        super(PingChurn, self).__init__(overlay)
+        self.ping_interval = ping_interval
+
+    def take_step(self, service_id=None):
+        self.overlay.routing_table.remove_bad_nodes()
```
qstokkink: All DiscoveryStrategies should claim the `with self.walk_lock:` on `take_step`.

DanGraur:
I removed the `MockDHTCommunity` class, and added the `walk_lock` acquisition code in the `take_step` method of the `PingChurn` class. Also, I rebased this branch. Should I squash the commit history of this branch as well?

Issue #353, opened 2018-11-05
Undeprecate frequently used components


qstokkink:
Many components in IPv8 are so frequently used, it makes no sense that they are in a deprecated folder. Specifically we should perform the following refactorings;

 - [ ] `ipv8/deprecated/bloomfilter.py` ➡️ `ipv8/messaging/bloomfilter.py`
 - [ ] `ipv8/deprecated/community.py` ➡️ `ipv8/community.py`
 - [ ] `ipv8/deprecated/lazy_community.py` ➡️ `ipv8/lazy_community.py`
 - [ ] `ipv8/deprecated/network_stats.py` ➡️ `ipv8/messaging/interfaces/network_stats.py`
 - [ ] `ipv8/deprecated/payload.py` ➡️ `ipv8/messaging/payload.py`
 - [ ] `ipv8/deprecated/payload_headers.py` ➡️ `ipv8/messaging/payload_headers.py`
 - [ ] `ipv8/peerdiscovery/deprecated/discovery.py` ➡️ `ipv8/peerdiscovery/community.py`
 - [ ] `ipv8/peerdiscovery/deprecated/discovery_payload.py` ➡️ `ipv8/peerdiscovery/payload.py`

Note 1: I do believe the files in `ipv8/messaging/deprecated` should no longer be used.
Note 2: We will have to change the Tribler downstream as well.

qstokkink:
`DiscoveryCommunity` addressed in #357.

PR #357, opened 2018-11-09
Added custom strategy declaration

### LINKED ISSUE #354: Allow communities to supply custom DiscoveryStrategies
We observed the need for communities to supply a custom `DiscoveryStrategy` subclass.

This is possible in the current system, but requires the strategy to be registered in the global name dictionary. Instead each community should have something like a `get_available_strategies()` which returns a map of names (`str`) to `DiscoveryStrategy` `class`es.

The `IPv8` instance can then map the configuration on the available global strategies and local community exposed strategies.

qstokkink:
Fixes #354, related to #353.

### DIFF
--- a/ipv8/overlay.py
+++ b/ipv8/overlay.py
@@ -78,6 +78,16 @@ class Overlay(six.with_metaclass(abc.ABCMeta, EndpointListener, TaskManager)):
         if global_time > self.global_time:
             self.my_peer.update_clock(global_time)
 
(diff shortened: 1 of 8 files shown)


ichorid:
Everything is fine. Though you still have to add me to me to reviewers list so I would be able to hit the "Approve" button.

qstokkink:
@ichorid oh that's weird, let me see if there's some repository setting for that.

PR #388, opened 2018-12-14
ON HOLD: Added ezpayload definition

qstokkink:
This PR uses some of the most rancid Python magic available to allow for shorter `Payload` definitions.

```python
class A(Payload):
    format_list = ["I", "H"]

    def __init__(self, a, b):
        self.a = a
        self.b = b

    def to_pack_list(self):
        return [("I", self.a),
                ("H", self.b)]

    @classmethod
    def from_unpack_list(cls, *args):
        return A(*args)
```

Becomes:

```python
@ezpayload
def A(a='I', b='H'):
    return A
```

See [the test file](https://github.com/Tribler/py-ipv8/pull/388/files#diff-c3b5edee13f972a762a9728a27cf9de0) for more examples.

Do note that aside from being shorter in definition, this is worse than the old format in every way imaginable. That's why I put this on `on hold`.

### DIFF
--- a/ipv8/messaging/lazy_payload.py
+++ b/ipv8/messaging/lazy_payload.py
@@ -1,0 +1,100 @@
+from __future__ import absolute_import
+
+import inspect
+
+from six.moves import xrange
+
+from .payload import Payload
+
+
+class RuntimePayload(Payload):
+    """
+    A "Runtime" (actually on module load) generated Payload definition.
+
+    You should never be initializing this directly.
+    Instead, creation should be done through the ezpayload wrapper.
+    """
+
+    ez_names = []
+
+    def __init__(self, *args, **kwargs):
+        """
+        Interpret the keyword arguments as class field assignments.
+
+        Ergo:
+            ..
+            >> mypayload = MyPayload(a=2)
+
+            >> mypayload.a == 2
+
+            True
+        """
+        # Extract the arguments used by the super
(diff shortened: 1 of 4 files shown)


qstokkink:
retest this please

qstokkink:
@ichorid do we want this badly enough to take it out of `on hold`?

ichorid:
We want it. We are going to try out a lot of new communities, and this will really help to keep the code in check.

ichorid:
The question is, can we inherit from the payloads that are defined the usual way?
Please, look at [`serialization.py`](https://github.com/Tribler/tribler/blob/c20089a0b50b034653d503a885ac17a72fff1e90/Tribler/Core/Modules/MetadataStore/serialization.py) and tell if it will work fine with methods like [`._serialized`](https://github.com/Tribler/tribler/blob/c20089a0b50b034653d503a885ac17a72fff1e90/Tribler/Core/Modules/MetadataStore/serialization.py#L144) ?

qstokkink:
That should all still work. My code is thusly nasty that at module-load time, the function definition gets rewritten as a fully fledged `Payload` object. This means that all of the existing stuff you could do with `Payload` instances, you can still do with this new format.

qstokkink:
We could also do this a lot less nasty by using the `kwargs`. I'll prototype that system soon. It should be a lot better.

PR #389, opened 2018-12-19
Added ez (un)pack methods to Serializer

qstokkink:
By popular demand and because I am a crow pleaser: slightly smaller pack and unpack methods for Serializable instances.

Adds:
 - `Serializer.ez_pack_serializables` which takes a list of `Serializable` objects instead of having to concatenate any lists or outputs yourself.
 - `Serializer.ez_unpack_serializables` which throws an exception if there is remaining data instead of forwarding the remaining data. This avoids some duplicate code we saw in the wrappers.

### DIFF
--- a/ipv8/lazy_community.py
+++ b/ipv8/lazy_community.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import
 
-from binascii import hexlify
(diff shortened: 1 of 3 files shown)


ichorid:
images

Issue #398, opened 2019-01-15
Include ipv8_service in pip


qstokkink:
The `setup.py` should also include `ipv8_service.py` in the list of `data_files`.

PR #402, opened 2019-01-17
Fixed a bug in the setup.py, where the main service class was not included

mitchellolsthoorn:
#398


devos50:
ok to test

devos50:
add to whitelist

PR #405, opened 2019-01-25
Don't delete blocks that are vital to your own chain

Captain-Coder:
The remove_old_blocks call in the database would delete blocks that are linked to a peers own chain. If this happens, the peer can't present full blocks of his past transactions.


qstokkink:
ok to test

PR #408, opened 2019-01-27
Various fixes

### LINKED ISSUE #406: Trackers broken
I just did some low level inspection into our peer discovery and we seem to be contaminated by the following "fake" addresses:

```
[('81.171.8.20', 35200), ('81.171.8.19', 35100), ('81.171.8.18', 35160), ('81.171.8.19', 35040), ('81.171.8.17', 35005), ('81.171.8.20', 35155), ('81.171.8.18', 35160)]
```

From a quick `traceroute` these appear to be our own trackers forwarded to us. This is very bad for the following reasons:
 1. Because these are not the same addresses as specified in `community.py` IPv8 does not recognize them as trackers.
 2. These peers will never answer community requests, but they will advertise themselves as part of the community (because that's what trackers do).
 3. IPv8 thinks it has enough options to walk to and stops bootstrapping.
 4. We get stuck in a loop of fake peers and never find any real peers :(

We should make sure that:
 1. The trackers are known by their correct ip and port in `community.py`.
 2. `RandomWalk` sometimes `get_new_introduction`'s instead of picking from `available`.

The first would be the proper fix, the second one should also be done to make sure this cannot be used as an attack vector by actual malicious nodes.

### LINKED ISSUE #407: IPv8 fails to create introduction requests when the chain length is higher than 65535
We are unable to reliably walk through the network if the chain size is higher than 65535. This is because we are using the `H` encoding for adding the chain length to an introduction request when walking to other peers in TrustChain, [see here](https://github.com/Tribler/py-ipv8/blob/master/ipv8/attestation/trustchain/community.py#L653).

@qstokkink would this require a breaking change to fix?

qstokkink:
Fixes #406, fixes #407, related to https://github.com/Tribler/tribler/issues/4171

### DIFF
--- a/ipv8/attestation/trustchain/community.py
+++ b/ipv8/attestation/trustchain/community.py
@@ -47,11 +47,8 @@ class TrustChainCommunity(Community):
     """
     Community for reputation based on TrustChain tamper proof interaction history.
     """
-    master_peer = Peer(unhexlify("3081a7301006072a8648ce3d020106052b8104002703819200040672297aa47c7bb2648ba0385275bc"
-                                 "8ade5aedc3677a615f5f9ca83b9b28c75e543342875f7f353bbf74baff7e3dae895ee9c9a9f80df023"
-                                 "dbfb72362426b50ce35549e6f0e0a319015a2fd425e2e34c92a3fb33b26929bcabb73e14f63684129b"
-                                 "66f0373ca425015cc9fad75b267de0cfb46ed798796058b23e12fc4c42ce9868f1eb7d59cc2023c039"
-                                 "14175ebb9703"))
+    master_peer = Peer(unhexlify("4c69624e61434c504b3a5730f52156615ecbcedb36c442992ea8d3c26b418edd8bd00e01dce26028cd"
+                                 "1ebe5f7dce59f4ed59f8fcee268fd7f1c6dc2fa2af8c22e3170e00cdecca487745"))
 
     DB_CLASS = TrustChainDB
     DB_NAME = 'trustchain'


qstokkink:
I ran this for +- 45 minutes inside of Tribler. Seems good.

devos50:
Nice, let me review 👍

devos50:
Also, should we tolerate a higher number of peers for the tunnel community? Otherwise, there is the risk that the peer list is only filled with 'normal' non-exit peers.

qstokkink:
@devos50 We could make a strategy to throw out peers we don't care about (maybe try to get some golden ratio of exitnodes/normal peers). Just gathering more peers doesn't really solve the problem long-term.

devos50:
Yup, I agree with a custom strategy for the tunnels. We can walk through the network as long as we do not have at least `max_circuits` exit nodes in our list (this number is defined in `TunnelSettings` and defines the maximum number of exit nodes we are using)?

qstokkink:
@devos50 want me to include it in this PR?

ipv8_service.py
```diff
@@ -128,16 +130,23 @@ def __init__(self, configuration, endpoint_override=None, enable_statistics=Fals
...
                             # We wrap the take_step into a general except as it is prone to programmer error.
                             try:
                                 strategy.take_step(service)
                             except:
                                 logging.error("Exception occurred while trying to walk!\n" +
                                               ''.join(format_exception(*sys.exc_info())))
+                        ticker -= 1 if ticker else 0
+                        sleep_time = smooth - (time.time() - start_time)
+                        if ticker and sleep_time > 0.01:
+                            time.sleep(sleep_time)
```
devos50: I assume `sleep_time` is very low? Otherwise, it might be a better idea to twistify this logic.

PR #412, opened 2019-01-28
Move tunnel crypto

egbertbouman:


qstokkink:
Just as a sanity check: did you run this in Tribler with an anonymous download?

egbertbouman:
@qstokkink I only tested it with an earlier version, but apparently I broke later on since it wasn't working. Of course, the unit tests in both IPv8 and Tribler all passed. Anyway, it's fixed now.

Note that Tribler still needs some small changes before the IPv8 pointer can be updated. I'll make a PR for this as soon as this one gets merged.

PR #428, opened 2019-02-06
Added default config Community intro stresstest

### LINKED ISSUE #427: Expand peer_discovery.py to check for Community specific failures
Right now `stresstest/peer_discovery.py` uses a `RandomWalk` on a naked `Community` instance.

We should either extend this test or make a new test which isolates and tests each of the default communities. This should allow us to spot regressions in the Community specific logic (instead of the default IPv8 logic).

qstokkink:
Fixes #427 

This allows us to spot regressions in the peer discovery logic of full-blown Communities.

Note that this is quite a bit heavier than `peer_discovery.py` and loads the entire stack.


qstokkink:
Woops, wrong blacklist.

PR #450, opened 2019-02-16
WIP: Tunnel fixes

egbertbouman:
Currently, when we fail to extend a circuit with new node, the circuit will timeout, and we start the entire process of circuit creation again. With this PR we will retry to extend the same circuit with different nodes. This should enable faster circuit creating for larger hop counts. Also fixes https://github.com/Tribler/tribler/issues/4089.


ipv8/requestcache.py
```diff
@@ -146,13 +146,14 @@ def _on_timeout(self, cache):
...
 
         self._logger.debug("timeout on %s", cache)
-        cache.on_timeout()
 
         # the on_timeout call could have already removed the identifier from the cache using pop
         identifier = self._create_identifier(cache.number, cache.prefix)
         if identifier in self._identifiers:
             del self._identifiers[identifier]
 
+        cache.on_timeout()
```
qstokkink: Why did you move this?

egbertbouman: I'm essentially using the requestcache to loop through possible candidates for circuit extension. The next requestcache is added in the `on_timeout` of the previous one. However, since the requestcache is currently only removed until after the `on_timeout` has finished executing, there will be an error (since we're adding a requestcache with a duplicate identifier). I've now fixed this by removing the requestcache before the `on_timeout` gets executed. An alternative way to work around this is to add a `callLater` from within the `on_timeout`, so we wait for the requestcache to be removed.

qstokkink: Well, we also have this beautiful construction in the TrustChainCommunity:

https://github.com/Tribler/py-ipv8/blob/8122e1b12a8fc012025eeefc5fb7df22ba2e0a7c/ipv8/attestation/trustchain/caches.py#L114

If we change this we need to make triple sure that nobody, either in IPv8 or Tribler, is depending on the order of ((identifier <-> cache)) removal.

egbertbouman: I agree that changing this does sound a bit risky. Then again, if we need to rely on this "beautiful construction" in multiple places, maybe the order is just wrong.

qstokkink: I never said it was pretty, but that is the way it works everywhere right now.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
Make sure you rebase this, we have Flake8 now in the PR tests.

egbertbouman:
@qstokkink The only thing that needs changing on the Tribler side is renaming a RequestCache.

PR #458, opened 2019-02-20
Added new endpoint base class with OPTIONS method and CORS preflight functionality

mitchellolsthoorn:
Moved CORS header to be included in every request
Added new base class for all API endpoints
All endpoints now have the HTTP OPTIONS method
All endpoints now have CORS preflight functionality
Renamed method in network endpoint because it was interfering with Twisted naming convention, causing it to show up in the allowed HTTP methods

### DIFF
--- a/ipv8/REST/base_endpoint.py
+++ b/ipv8/REST/base_endpoint.py
@@ -1,0 +1,35 @@
+from __future__ import absolute_import
+
+from twisted.web import resource
+from twisted.web.resource import _computeAllowedMethods
+
+
+class BaseEndpoint(resource.Resource):
+    """
+    The base endpoint from which all other endpoints should extend to make them compatible with Cross-Origin Resource
+    Sharing requests.
+    """
+
+    def render_OPTIONS(self, request):
+        """
+        This methods renders the HTTP OPTIONS method used for returning available HTTP methods and Cross-Origin Resource
+        Sharing preflight request checks.
+        """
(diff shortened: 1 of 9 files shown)


mitchellolsthoorn:
Changed 
```return ""```
to 
```return b""```
to comply with the other code

File: ipv8/REST/base_endpoint.py [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+from twisted.web import resource
+from twisted.web.resource import _computeAllowedMethods
+
+
+class BaseEndpoint(resource.Resource):
+    """
+    The base endpoint from which all other endpoints should extend to make them compatible with Cross-Origin Resource
+    Sharing requests.
+    """
+
```
qstokkink: Because `resource.Resource` is an old-style Python class, all of its base classes aren't really `object`s. Now that you introduced a common base class this should have a proper `__init__` so as to not screw with new-style Python objects. I did this before, so you can copy paste that from here:

https://github.com/qstokkink/py-ipv8/blob/7f8d76e00b66fd69dd72d2a01c9062b8bbc44ee4/ipv8/REST/formal_endpoint.py#L27-L31

Then all of the classes should also properly call `super(..., self).__init__(...)`.

mitchellolsthoorn: It took some tries to get it working, but they all pass now. The order of the superclasses is very specific

PR #465, opened 2019-03-25
Update hidden services

egbertbouman:


ipv8/messaging/anonymization/community.py
```diff
@@ -91,67 +89,44 @@ def __init__(self):
...
+        self.max_time_ip = 24 * 60 * 60
         # Maximum number of seconds before a circuit is considered inactive (and is removed)
         self.max_time_inactive = 20
         self.max_traffic = 250 * 1024 * 1024
 
         self.max_packets_without_reply = 50
-        self.dht_lookup_interval = 30
         # Maximum number of seconds circuit creation is allowed to take. Within this time period, the unverified hop
         # of the circuit can still change in case it is unresponsive.
         # TODO: set this to ~60s next time we change the community ID, or after enough nodes are using the newer code
```
qstokkink: Should we set the timeout to `60` now then?

ipv8/messaging/anonymization/community.py
```diff
@@ -805,6 +791,12 @@ def on_extend(self, source_address, payload, _):
             if not extend_candidate:
                 extend_candidate = Peer(payload.node_public_key, payload.node_addr)
 
+            # Ensure that we are able to contact this peer
+            if extend_candidate.last_response + 57.5 < time.time():
```
qstokkink: Why not use `self.settings.unstable_timeout` instead of `57.5`?

ipv8/messaging/anonymization/hidden_services.py
```diff
@@ -388,44 +410,67 @@ def on_linked_e2e(self, source_address, payload, circuit_id):
...
+        self.logger.info('Established introduction point for %s', binascii.hexlify(payload.public_key))
+
         circuit = self.exit_sockets[circuit_id]
-        self.intro_point_for[payload.info_hash] = circuit
+        self.intro_point_for[payload.public_key] = circuit, payload.info_hash
+
+        if not self.ipv8:
+            self.logger.error('No IPv8 service object available, cannot start PEXCommunity')
+        elif payload.info_hash not in self.pex:
+            community = PexCommunity(self.my_peer, self.ipv8.endpoint, self.ipv8.network, info_hash=payload.info_hash)
```
qstokkink: You may want to use a new `Network()` here, to completely isolate the PEX peers from the other peers.

ipv8/messaging/anonymization/community.py
egbertbouman: I prefer to fix this in a separate PR, since I'm still expecting some issues with changing this value.

ipv8/messaging/anonymization/community.py
egbertbouman: Because they are 2 different things. The `57.5` refers to how long after receiving a message from a peer do you consider this peer connectable, whereas the `unstable_timeout` refers to the period during which a circuit is allowed to change.

ipv8/messaging/anonymization/hidden_services.py
egbertbouman: Sure, I'll change this.

egbertbouman:
@qstokkink That depends. If we're going to do another beta, we can merge it, otherwise it's better to postpone until after 7.3.

devos50:
@egbertbouman we will do another beta but I guess this PR also needs changes in Tribler?

egbertbouman:
@devos50 Yep, as soon as this PR is merged I'll create another PR for the changes in Tribler.

ipv8/messaging/interfaces/endpoint.py
```diff
@@ -17,29 +17,29 @@ class Endpoint(six.with_metaclass(abc.ABCMeta, object)):
 
     def __init__(self):
         self._logger = logging.getLogger(self.__class__.__name__)
-        self._listeners = []
+        self._listeners = {}
 
-    def add_listener(self, listener):
+    def add_listener(self, prefix, listener):
```
qstokkink: We have a lot of infrastructure that depends on this. If you need mapping, make a new Community subclass.

synctext:
For raw performance I hope we can accept _some_ not ideal code.. 50% boost in anon download is really nice.

qstokkink:
> 
> 
> For raw performance I hope we can accept _some_ not ideal code.. 50% boost in anon download is really nice.

Yes. Though there is a delicate balance between performance, flexibility and maintainability. I don't want to start from scratch _again_ because we didn't make the core system flexible enough.

Also, this PR is big enough with just TunnelCommunity changes, it shouldn't also include random fixes.

egbertbouman:
retest this please

Issue #451, opened 2019-02-18
DHTEndpoint improvements


qstokkink:
Migrated from https://github.com/Tribler/tribler/issues/4227#issuecomment-464483808 by @egbertbouman, regarding the REST `DHTEndpoint` :

>Looking at this code I do wonder about the chances of everything going OK. I see no retry mechanism, and the blocklistener seems to get notified only once about a new block.. Also note that values in the DHT are currently not automatically republished, meaning that the value will dissapear after relatively short period.

> Additionally, if the number of chunks grow beyond MAX_VALUES_IN_FIND, you'll never be able to get the full block from the DHT. Then again, blocks are probably small.

DanGraur:
I think I can take a look at this.

qstokkink:
I believe this issue has been sufficiently addressed now (and the OP is too confusing to keep this one open). @DanGraur if there is anything left to do for the DHTEndpoint please make a new issue.

DanGraur:
Sure thing!

PR #468, opened 2019-04-23
Retry mechanism for new TC block publishing to DHT

DanGraur:
This PR attempts to address one of the improvements to the DHT flagged in issue #451.  When a new TC block is created it will automatically be published to the DHT (by the peer which owns the block) using a mechanism which will attempt to publish each chunk of the block individually, only moving to the next chunk of the block when the previous chunk was successfully published. If a chunk for whatever reason cannot be published, the current implementation will seek to retry to publish the chunk. There is a limited number of retries per chunk (currently 3), thus after that many tries, no more attempts will be made, and the block will not be published completely. More specifically, the current code will not try to publish the next chunks in the block at all.  As a side effect of this, a block may be partially published. 

Additionally, this improved implementation will reattempt to fetch the blocks under a particular key (since this is an operation required prior to publishing the new block), if the operation initially fails (again, there's a limit of 3 attempts).

### DIFF
--- a/ipv8/REST/dht_endpoint.py
+++ b/ipv8/REST/dht_endpoint.py
@@ -88,9 +88,12 @@ class DHTBlockEndpoint(BaseEndpoint, BlockListener):
         for entry in block_chunks:
             package = self.serializer.unpack_to_serializables([DHTBlockPayload, ], entry)[0]
 
-            if public_key.verify(package.signature, str(package.version).encode('utf-8')
-                                 + str(package.block_position).encode('utf-8')
-                                 + str(package.block_count).encode('utf-8') + package.payload):
+            pre_signed_content = self.serializer.pack_multiple([('H', package.version),
+                                                                ('H', package.block_position),
+                                                                ('H', package.block_count),
+                                                                ('raw', package.payload)])[0]
+
+            if public_key.verify(package.signature, pre_signed_content):
                 max_version = max_version if max_version > package.version else package.version
 
                 if package.version not in new_blocks:
@@ -134,9 +137,11 @@ class DHTBlockEndpoint(BaseEndpoint, BlockListener):
                 # Prepare and pack the chunk for publishing to the DHT
                 my_private_key = self.trustchain.my_peer.key
                 chunk = latest_block[slice_pointer: slice_pointer + self.CHUNK_SIZE]
-                signature = my_private_key.signature(str(self.block_version).encode('utf-8')
-                                                     + str(chunk_idx).encode('utf-8')
-                                                     + str(total_chunks).encode('utf-8') + chunk)
+
+                pre_signed_content = self.serializer.pack_multiple([('H', self.block_version), ('H', chunk_idx),
+                                                                    ('H', total_chunks), ('raw', chunk)])[0]
+
+                signature = my_private_key.signature(pre_signed_content)
                 blob_chunk = self.serializer.pack_multiple(
                     DHTBlockPayload(signature, self.block_version, chunk_idx, total_chunks, chunk).to_pack_list())
 
--- a/ipv8/REST/dht_endpoint.py
+++ b/ipv8/REST/dht_endpoint.py
@@ -51,10 +51,9 @@ class DHTBlockEndpoint(BaseEndpoint, BlockListener):
 
         :param block: the latest block added to the Database. This is not actually used by the inner method
         """
-        # We'll retrieve the latest block for this key
-        #TODO: inquire if it's safe to always used the block passed as parameter, rather than latest_block, which might introduce a race condition
-        latest_block = self.trustchain.persistence.get_latest(self.trustchain.my_peer.public_key.key_to_bin())
-        deferLater(reactor, 0, self.publish_latest_block, latest_block, 1)
+        # Check if the block is not null, and if it belongs to this peer
+        if block and block.public_key == self.trustchain.my_peer.public_key.key_to_bin():
+            deferLater(reactor, 0, self.publish_latest_block, block, 1)
 
     def should_sign(self, block):
         pass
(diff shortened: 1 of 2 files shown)


DanGraur:
I have a few remarks and questions to make regarding the current implementation:

* I should also mention that the current implementation has a limitation when it comes to the *retry mechanism*. In order to implement this mechanism, I used `Deferred`s. More specifically I used the `Deferred` object [returned](https://github.com/DanGraur/py-ipv8/blob/54fd18e91958621ab4626c01e63acbae75347416/ipv8/REST/dht_endpoint.py#L143) by `DHTCommunity.store_value`, to which I hooked a recursive call to the same method which contains the call (`publish_chunk`) but for the next chunk. I also hooked an `errback` in case the current `publish_chunk` call fails. This `errback` is also a recursive call (to `publish_chunk`), but this time for the same chunk, but with an increased attempt count. The issue here is the following: if the method fails before the call to `DHTCommunity.store_value` (where the proper `errback` is hooked up, i.e. for this block), then due to the chaining rules in `Deferred`s it will jump back to the previously hooked `errback` (which should always be the the last hooked `errback` of the previous block, as the error can only stem from one call, and it is always deterministic). I looked through the code, and came to the conclusion that the only reasonable source of an error will be [here](https://github.com/DanGraur/py-ipv8/blob/54fd18e91958621ab4626c01e63acbae75347416/ipv8/REST/dht_endpoint.py#L136), i.e. the call to `signature`. However, from what I can see, the result of this call will be deterministic, and regardless of how many times we call it with the same input, there will always raise an error. As for the 'jumping back to the previous block' part, this shouldn't affect the DHT, since a chunk [will not be republished if its version is the same](https://github.com/DanGraur/py-ipv8/blob/54fd18e91958621ab4626c01e63acbae75347416/ipv8/dht/storage.py#L51) (as it was already exists int the DHT), thus the DHT's size will not increase. In the end, publishing the block will fail, but not immediately, as it will move to the previous block first, succeed in publishing, retry to publish the current block, fail, and so on, until the previous block's attempts run out. The only exception to this execution path is when the first chunk fails, and no `errback` has been hooked up. In this case, the failure will be immediate. Thus, I'm not sure if this is a concern or not, but I thought I should mention it. 

* I did not add any explicit test cases for this PR, since I'm not sure how to simulate the failure of certain methods without changing their source code. I did however, test the implementation informally (by changing the source code such that errors are simulated, and `errback`s are triggered). I can look further into this issue, and try to find a way to test these new additions (perhaps by subclassing and changing the code there?). 

* In the current implementation, when `DHTBlockEndpoint` is instantiated, the `self.block_version` will be set to 0. Should this always be the case, or might there be cases where it should start from another value (inherited from a previous run)?

* Is it ever possible that calling `pack` on a `TrustChainBlock` will ever return an empty string, and as a consequence, is it worth using defensive programming [here](https://github.com/DanGraur/py-ipv8/blob/54fd18e91958621ab4626c01e63acbae75347416/ipv8/REST/dht_endpoint.py#L163)?

Let me know if you'd like me to look into ways of testing this, and also if there are concerns with the limitations and concerns I've flagged in this comment.

ipv8/REST/dht_endpoint.py
```diff
@@ -99,53 +103,65 @@ def reconstruct_all_blocks(self, block_chunks, public_key):
...
+                # If we've attempted to publish this block more times than allowed, give up on the chunk and the block
+                if chunk_attempt > self.ATTEMPT_LIMIT:
+                    logging.error("Publishing latest block failed after %d attempts on chunk %d", self.ATTEMPT_LIMIT,
+                                  chunk_idx)
+                    return
+
+                # Prepare and pack the chunk for publishing to the DHT
+                my_private_key = self.trustchain.my_peer.key
+                chunk = latest_block[slice_pointer: slice_pointer + self.CHUNK_SIZE]
+                signature = my_private_key.signature(str(self.block_version).encode('utf-8')
```
egbertbouman: I think it's better to use the IPv8 serializer to construct the data to sign.  This would also avoid the problem of giving an attacker the ability to construct valid DHT entries by replaying signatures.

DanGraur: Should be done now, I think.

ipv8/REST/dht_endpoint.py
```diff
@@ -100,53 +107,67 @@ def reconstruct_all_blocks(self, block_chunks, public_key):
...
+                    return
 
-                self.block_version += 1
+                # We now know the block is novel, so we'll try to publish it
+                deferLater(reactor, 0, publish_chunk, None, 0, 0)
 
-            self.dht.find_values(self._hashed_dht_key).addCallbacks(on_success, on_failure)
+            # On the off chance that the block is actually empty, avoid publishing it
+            if total_chunks > 0:
+                self.dht.find_values(self._hashed_dht_key).addCallbacks(on_success_find_values, on_failure_find_values)
```
egbertbouman: Quick question: why are we first doing a `find_values` before storing stuff to the DHT? Isn't the `BlockListener` called only once for every block? If so, I'm not sure whether this is needed. Other than that, this PR looks ready to merge.

DanGraur: It's just a bit of defensive programming. Indeed, I don't think there's a significant risk of a block being duplicated (since a callback is called only once for a new block), but in an effort to avoid trying to republish the same block twice (if somehow that might happen), I added that part in as well. I'll remove it, since it's likely not very useful.

qstokkink:
Other than what @egbertbouman mentioned, I am fine with this PR.

synctext:
Libtorrent would be great to use in future for publishing TC blocks. Better bulk performance vs. DHT.

qstokkink:
@synctext this is a mechanism for last-block withholding, not for syncing the entire chain.

ipv8/REST/dht_endpoint.py
DanGraur: Should be fixed now.

DanGraur:
Should I clean up the commit history?

qstokkink:
Commit history seems fine. @egbertbouman could you review this again?

PR #469, opened 2019-04-29
Return connected peers

xoriole:
Added an option to fetch the connected peers to a public key (user). It is added to build a beautiful trust graph in Tribler.

Instead of this (expensive)
all_peers

Only show connected peers (cheaper)
circular_graph


qstokkink:
Also please add the new test file to the [test classes list](https://github.com/Tribler/py-ipv8/blob/master/test_classes_list.txt), then this one is good to go i.m.o.

xoriole:
Done

PR #473, opened 2019-05-16
Support for large block retrieval in DHTs 

DanGraur:
This PR attempts to address one of the improvements to the DHT flagged in issue #451. Added support for block entries in the DHT, whose chunks do not fit into a single response returned after a DHT block request. The implementation forwarded in this PR is able to return a block which is complete and has the maximal version, as long as it is within the scope of the query. The following high level description of the algorithm should hopefully make what this means more clear (*chunk* and *block* are interchangeably used):

1. Given the public key of a peer, forward a REST request via HTTP for the peer's latest block. 
1. The peer which receives this request will query its own DHT object for the entries under a special key for which the target peer's TC blocks are stored. Set `query_start = 0`.  
1. The query will return `MAX_VALUES_IN_FIND = 8` DHT entries  under the aforementioned key, starting from the index `query_start` in the absolute order of the DHT entries (i.e. from `query_start` to `query_start + MAX_VALUES_IN_FIND`, sorted by the distance to querying peer). 
1. Given the new set of entries, the algorithm will try to build new blocks. If we have already started to build blocks, then the algorithm will try to further build on top of those.
1. Each chunk has an associated version (its location in the chain). The algorithm keeps track of the maximal version which it has seen so far. 
1. If the maximal version block (seen so far) is complete, stop the process, and return the block. If there are no more blocks under this key, return the block which is complete and also has the greatest version. Otherwise, increment `query_start = query_start + MAX_VALUES_IN_FIND`, and forward another query for entries in the DHT. Go to step 3.

I'm not sure if this behavior generally suffices, but the alternative (for always returning the block with the guaranteed greatest version) to this would be to either retrieve all the chunks stored in the DHT under some peer's key, or hold the maximal version of a block in a trusted location which is accessible to any peer. Currently, this is a *best effort* approach.

### DIFF
--- a/ipv8/REST/dht_endpoint.py
+++ b/ipv8/REST/dht_endpoint.py
@@ -234,7 +234,7 @@ class DHTBlockEndpoint(BaseEndpoint, BlockListener):
                         u"error": {
                             u"handled": True,
                             u"message": "Could not reconstruct any block successfully." if start_idx != 0 else
-                            u"Could not find any blocks for the specified key."
+                                        u"Could not find any blocks for the specified key."
                         }
                     }))
 
--- a/ipv8/REST/dht_endpoint.py
+++ b/ipv8/REST/dht_endpoint.py
@@ -14,7 +14,7 @@ from .base_endpoint import BaseEndpoint
 from ..attestation.trustchain.community import TrustChainCommunity
 from ..attestation.trustchain.listener import BlockListener
 from ..attestation.trustchain.payload import DHTBlockPayload
-from ..dht.community import DHTCommunity, MAX_ENTRY_SIZE
+from ..dht.community import DHTCommunity, MAX_ENTRY_SIZE, MAX_VALUES_IN_FIND
 from ..dht.discovery import DHTDiscoveryCommunity
 from ..keyvault.public.libnaclkey import LibNaCLPK
 from ..messaging.serialization import PackError, Serializer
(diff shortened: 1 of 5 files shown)


DanGraur:
I should mention that in order to get this to work, and in order to avoid rewriting a new DHT query mechanism which would be marginally different from what is already there, I slightly modified the mechanism that was already there. That is, I added the possibility of specifying which entries should be returned via the `start_idx` parameter. The parameter is used in the `DHTCommunity` class starting [here](https://github.com/DanGraur/py-ipv8/blob/6be329a9a4783b681569ccc581824c647ff329ae/ipv8/dht/community.py#L416) for the part which sends a find request, and starting [here](https://github.com/DanGraur/py-ipv8/blob/6be329a9a4783b681569ccc581824c647ff329ae/ipv8/dht/community.py#L431) for the part which responds to a find request. I also changed the the `get` method in the [`Storage`](https://github.com/DanGraur/py-ipv8/blob/6be329a9a4783b681569ccc581824c647ff329ae/ipv8/dht/storage.py#L58) class. Finally, I also had to change the [`FindRequestPayload`](https://github.com/DanGraur/py-ipv8/blob/6be329a9a4783b681569ccc581824c647ff329ae/ipv8/dht/payload.py#L106) so that it stores and sends the `start_idx` as well.

qstokkink:
Hmm.. we're not really at a good point to break the wire format. We should wait with merging this until after Tribler 7.3 releases.

DanGraur:
I've changed the hash which is used to initialize the `master_peer` in `DHTCommunity`. 

I do agree that this change might not yet be necessary, as it does break the clean design of the DHT. However, if later, the underlying mechanism forwarded in this PR, i.e. that of lazily forwarding DHT requests for new entries under a DHT key is necessary (especially for TrustChain blocks), then this might be worthwhile to be merged in the `master` branch.

qstokkink:
Tribler 7.3 has been released, this is no longer on hold. Could you patch this PR up to work on the latest `master`?

qstokkink:
Thanks 👍

PR #481, opened 2019-05-21
Properly encoding to/from JSON in Python 3

devos50:
In this commit, I copied the json_util.py file over from Tribler. These methods are used in the IPv8 REST API to encode to/from JSON, in a Python 3-compatible way. I also fixed a few inconsistencies and code style errors.

### DIFF
--- a/ipv8/REST/attestation_endpoint.py
+++ b/ipv8/REST/attestation_endpoint.py
@@ -1,11 +1,12 @@
 from __future__ import absolute_import
 
-import json
 from base64 import b64decode, b64encode
 from hashlib import sha1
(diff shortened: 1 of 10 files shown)


devos50:
My IDE is not complaining about the Pylint error for some reason, I think we can ignore it?

ichorid:
Indeed, there *is* a pylint error, according to our pylint...

qstokkink:
Also, please fix the Pylint error.

qstokkink:
Third (and final?) thought, please move `json_util` to the `ipv8/REST` folder.

PR #484, opened 2019-05-21
Changed get_latest_blocks behaviour

devos50:
You can now pass a list of block types instead of a single block type. This will help us if we want to query the latest blocks in a specific chain for some application that only uses a limited number of different block types.


qstokkink:
I trust you will make the appropriate changes to Tribler?

Issue #485, opened 2019-05-21
Use `python2 -m pip` instead of `pip` in documentation


TimSpeelman:
When running `pip install ..`, the packages are installed for the default Python version, which may very well be Python 3. This lead me into some problems of the `cannot find module X` kind, which I resolved by using `python2 -m pip ..` after figuring out the problem.

Although it is more of a general pip thingy, given that we use python 2, we may want to be explicit with the pip installs as well. This saves people some trouble. If agreed, I'll make a PR.

qstokkink:
We don't just use Python 2. You can also use Python 3. You should use `pip2` or `pip3` accordingly.

qstokkink:
By using `pip` the system selects the most appropriate Python distribution for you.

TimSpeelman:
Allright, would you say this is something we could clarify in the docs or let the user figure that out? We do explicitly mention `python2` in the README, attestation and overlay tutorial. 

Also, `pip2` does not work out of the box for me on Windows, but that could be my problem.

qstokkink:
The average user will probably only have one Python version installed. Nevertheless clarity in the documentation is always good.

The mentions of Python 2 in the documentation are out of date, that should be updated. We now support both Python 2 and Python 3.

devos50:
> The average user will probably only have one Python version installed.

When installing a clean Ubuntu version, you usually have Python 2 and 3 installed (since there are some dependencies that use either Python 2 or Python 3). However, usually `python2` is symlinked to `/usr/bin/python`.

qstokkink:
Closed by #486

PR #486, opened 2019-05-21
Remove python2 specifics from docs and unix tests

TimSpeelman:
As discussed in #485, I replaced the outdated `python2` with `python` througout the docs. Also, I removed the python2 specific test from the Unix tests, otherwise this may throw a confusing error when the user `pip installs` to python3 but the tests decide to run v2.


qstokkink:
ok to test

run_all_tests_unix.sh
```diff
@@ -8,13 +8,7 @@ test_files="$(grep ^[^#] test_classes_list.txt)"
...
 tty -s && tput bold
 
-# 2.1 Check for python interpreter version
-if [ -x "$(command -v python2)" ]; then
-    interpreter="python2"
-else
-    echo "No python2 command found! Will use the default python command."
-    interpreter="python"
-fi
+interpreter="python"
```
qstokkink: We can't change this right now, our PR testing suite on Jenkins depends on this being Python 2.

TimSpeelman:
Ik don't understand why the checks fail, when trying to open the Flake8 details I get a 403.

qstokkink:
Our flake8 test assumes all PRs modify at least some code. You can ignore the failure.

PR #489, opened 2019-05-21
READY: Automatic DHT entry republishing for stored TrustChain block chunks 

DanGraur:
This PR addresses one of the issues flagged in #451. TrustChain (TC) block chunks stored in the DHT are now automatically republished when they are close to expiry. To enable this behavior, a `LoopingCall` has been added to `DHTCommunity` which periodically calls the new method `block_chunk_refresh`. The method `block_chunk_refresh` is actually tasked with refreshing the blocks by inspecting the expiry time of each chunk under the node's DHT TC block entries. If `entry.age + TC_BLOCK_DELTA > entry.max_age`, where `TC_BLOCK_DELTA` represents some delta before expiry after which a refresh is eligible, then a refresh is executed on that chunk by calling `_store` on it (the refresh will extend to other nodes as well, which is intended behavior). The current value of `TC_BLOCK_DELTA` is set to `310`. This is just a random value that I chose, and it can be easily changed. The `LoopingCall`'s period is `TC_BLOCK_DELTA - 10 = 300`.

### DIFF
--- a/ipv8/dht/community.py
+++ b/ipv8/dht/community.py
@@ -467,9 +467,8 @@ class DHTCommunity(Community):
         """
         Republish any TrustChain block chunks which are about to expire in the DHT
         """
-        from ..REST.dht_endpoint import DHTBlockEndpoint
-
-        hashed_dht_key = hashlib.sha1(self.my_peer.public_key.key_to_bin() + DHTBlockEndpoint.KEY_SUFFIX).digest()
+        from ..REST.dht_endpoint import TC_KEY_SUFFIX
+        hashed_dht_key = hashlib.sha1(self.my_peer.public_key.key_to_bin() + TC_KEY_SUFFIX).digest()
 
         # for value in self.storage.get(hashed_dht_key):
         for value in self.storage.items[hashed_dht_key]:
--- a/ipv8/dht/community.py
+++ b/ipv8/dht/community.py
@@ -467,8 +467,9 @@ class DHTCommunity(Community):
         """
         Republish any TrustChain block chunks which are about to expire in the DHT
         """
-        key_suffix = b'_BLOCK'
-        hashed_dht_key = hashlib.sha1(self.my_peer.public_key.key_to_bin() + key_suffix).digest()
+        from ..REST.dht_endpoint import DHTBlockEndpoint
+
+        hashed_dht_key = hashlib.sha1(self.my_peer.public_key.key_to_bin() + DHTBlockEndpoint.KEY_SUFFIX).digest()
 
         # for value in self.storage.get(hashed_dht_key):
         for value in self.storage.items[hashed_dht_key]:
--- a/ipv8/dht/community.py
+++ b/ipv8/dht/community.py
@@ -53,6 +53,9 @@ MSG_STORE_RESPONSE = 10
 MSG_FIND_REQUEST = 11
 MSG_FIND_RESPONSE = 12
 
+# Indicates the delta before the DHT entry expiration time, after which the entry is eligible for refresh
+TC_BLOCK_DELTA = 310  # seconds
+
 
 def gatherResponses(deferreds, **kwargs):
     def on_finished(results):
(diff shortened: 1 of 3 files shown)


DanGraur:
I should mention that this PR has 3 commits. The first one actually implements the core of this PR. The latter two are just alternative implementations for trying to avoid circular dependencies in a more elegant way than it is implemented in the first commit. 

The code I wrote for the TC blocks via DHT makes use of a literal value which is used to generate the DHT key under which the TC chunks are stored. The literal is suffixed to the end of a node's public key before the entire string is hashed. The hash itself is the key. An example can be seen [here](https://github.com/Tribler/py-ipv8/blob/eafaf9e7edea466179701ec6ff69ae60894a436b/ipv8/REST/dht_endpoint.py#L72). Previously, I only had to use this suffix in the `DHTBlockEndpoint` class, and in the unit test class (where I could import it). However, now I need to make use of it in `DHTCommunity`, since I need to get the local DHT entries.

In order to reduce the use of literals, and the need of having to make changes in multiple places later on, if this value needs to be changed, I want to only have a single place where it is defined, and import that reference whenever I need the value. However, since `DHTBlockEndpoint` is dependent on `DHTCommunity` the opposite cannot happen at the same time. 

The changes I'm referring to are all in the `block_chunk_refresh` method. In the initial commit, I'm using two literals (the original + the new one). In the second and third, I'm using a local import (I know it's not a very good idea, but in these circumstances I hope it might be acceptable). In the second commit, I have to import the entire `DHTBlockEndpoint` class, since the literal is one of its fields. In the third, I changed this structure, and moved the literal out of any class, but left it in the `dht_endpoint.py` module. This way, in the third commit I can only import the value, and not the entire class. Which one do you think is better / do you have any better alternative solutions?

ipv8/dht/community.py
```diff
@@ -106,6 +109,8 @@ def __init__(self, *args, **kwargs):
         self.token_secrets = deque(maxlen=2)
         self.register_task('value_maintenance', LoopingCall(self.value_maintenance)).start(3600, now=False)
         self.register_task('token_maintenance', LoopingCall(self.token_maintenance)).start(300, now=True)
+        self.register_task('tc_block_maintenance', LoopingCall(self.block_chunk_refresh)).start(TC_BLOCK_DELTA - 10,
```
egbertbouman: I would prefer to keep TrustChain-specific logic outside of the DHTCommunity.

ipv8/dht/community.py
```diff
@@ -458,6 +463,18 @@ def on_find_response(self, peer, dist, payload):
...
 
+    def block_chunk_refresh(self):
+        """
+        Republish any TrustChain block chunks which are about to expire in the DHT
+        """
+        from ..REST.dht_endpoint import TC_KEY_SUFFIX
+        hashed_dht_key = hashlib.sha1(self.my_peer.public_key.key_to_bin() + TC_KEY_SUFFIX).digest()
+
+        # for value in self.storage.get(hashed_dht_key):
+        for value in self.storage.items[hashed_dht_key]:
```
egbertbouman: We don't know if the owner of the block stores its values in `Storage`, as it depends on the distance between node ID's. It's better to just query your TrustChain database to get the required values. Please only republish your latest block, to avoid ending up storing your entire TrustChain in the DHT.

ipv8/dht/community.py
```diff
@@ -458,6 +463,18 @@ def on_find_response(self, peer, dist, payload):
...
+            if value.age + TC_BLOCK_DELTA > value.max_age:
```
egbertbouman: We can't wait until we get (close to) max_age. Nodes on the network continuously join/leave the network, and chances are that all nodes that originally stored our values will have left the network. It's better to simply republish your latest block hourly (assuming you haven't created any new block for the past hour).

DanGraur:
Given this review, I think the implementation forwarded in this PR is completely unsuitable. If it's ok, I'll close the PR, and forward a new one, with a different implementation based on @egbertbouman's suggestions.

qstokkink:
👍

PR #500, opened 2019-05-24
Various fixes

egbertbouman:
Fixes https://github.com/Tribler/tribler/issues/4516.


qstokkink:
@egbertbouman is this ready for review yet?

egbertbouman:
@qstokkink Not yet, I would also like to change the behavior of `TunnelEndpoint` so that it creates a circuit using peers that have the `PEER_FLAG_EXIT_IPV8` flag set.

PR #512, opened 2019-05-29
Automatic DHT entry republishing for the latest stored TrustChain block chunks

DanGraur:
This PR addresses part of issue #451. The latest stored TrustChain block in the local TrustChain database will be automatically republished to the DHT by the node which owns it. A block that is stored in the local TrustChain database should already have been published as soon as the block was added to the database, so the newly added mechanism should just republish it periodically. Currently, the refresh is done on an hourly basis. The chunks will not be duplicated if they are already stored in the DHT. 

To enable the periodical refresh, a `LoopingCall` in the `DHTBlockEndpoint`, which is attached to the peer's `TrustChainCommunity`'s `TaskManager` object, schedules the `publish_latest_block` every hour. This method implements the refresh. When a new block is added, the `LoopingCall` is `reset`. This means that after a new block is added, the `publish_latest_block` method is called after one hour, regardless of how much time was left in the `LoopingCall`'s timer. 

When a block is **republished**, the `block_version` in `DHTBlockEndpoint` is not incremented, so as to avoid chunk duplication (since the stored entities in the DHT would be different). Since the `block_version` always points to the next block, it will actually be temporarily decremented, so as to publish an identical block as before. The `block_version` is only incremented permanently when a brand new block is published. 

Two unit tests have been added in order to check that the block refresh behavior is correct, and does not imply any duplication in the DHT storage.

### DIFF
--- a/ipv8/REST/dht_endpoint.py
+++ b/ipv8/REST/dht_endpoint.py
@@ -6,7 +6,7 @@ from binascii import hexlify, unhexlify
 from hashlib import sha1
 
 from twisted.internet import reactor
-from twisted.internet.task import deferLater
+from twisted.internet.task import LoopingCall, deferLater
 from twisted.web import http
 from twisted.web.server import NOT_DONE_YET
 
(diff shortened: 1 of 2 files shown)


DanGraur:
This is a follow-up to the closed PR #489, and implements @egbertbouman's suggestions.

qstokkink:
This seems fine, @egbertbouman any concerns?

PR #515, opened 2019-06-04
Made attestation proof flexible

qstokkink:
This is the first step to turn the attestation framework into an SDK.


qstokkink:
I'll rework the REST endpoint, such that the `id_format` defaults to `"id_metadata"`.

qstokkink:
retest this please

PR #516, opened 2019-06-04
WIP: Fix Python 3 support for endpoints

egbertbouman:


ipv8/attestation/trustchain/block.py
```diff
@@ -445,8 +445,12 @@ def __iter__(self):
...
             if key == 'key' or key == 'serializer' or key == 'crypto' or key == '_transaction':
                 continue
-            if isinstance(value, string_types) and key != "insert_time" and key != "type":
-                yield key, hexlify(value)
+            if isinstance(value, binary_type) and key != "insert_time" and key != "type":
+                yield key, hexlify(value).decode()
+            elif key == 'type':
+                yield key, value.decode()
+            elif key == 'transaction':
+                yield key, {k.decode(): (v.decode() if isinstance(v, binary_type) else v) for k, v in value.items()}
```
qstokkink: This is the wrong place to do this. Transactions can be of any depth, you would have to iterate over all JSON compatible children of the transaction and convert everything. Doing this properly should be left to the transaction/overlay designer.

PR #518, opened 2019-06-07
Added Deferred return value to crawl_chain method

devos50:
This Deferred is called when the crawler finishes with the crawl. It can be used as notification for other components to perform a specific action when a crawl has finished.


devos50:
@qstokkink should we get rid of the wildcard import there?

qstokkink:
Wildcard imports are more efficient than `from` imports.

However, I have come to appreciate the readability of `from` imports, so I've been steadily phasing these wildcard imports out of IPv8.

You can leave it for now.

PR #519, opened 2019-06-07
Made should_sign method asynchronous

devos50:
The `should_sign` currently returns a boolean value, indicating whether we should sign an incoming block or not. However, it is possible that some asynchronous operation needs to happen before we can determine whether an incoming block should be signed (e.g., fetching some network data). To allow for this use-case, I slightly refactored the `should_sign` method in the `TrustChainCommunity`, such that it can deal with listeners that are returning a `Deferred` in the should_sign method.

The method is fully backwards compatible with the existing implementations of `should_sign`.

### DIFF
--- a/ipv8/attestation/trustchain/community.py
+++ b/ipv8/attestation/trustchain/community.py
@@ -13,7 +13,7 @@ from functools import wraps
 from threading import RLock
 
 from twisted.internet import reactor
-from twisted.internet.defer import Deferred, fail, succeed
+from twisted.internet.defer import Deferred, fail, inlineCallbacks, maybeDeferred, returnValue, succeed
 from twisted.internet.task import LoopingCall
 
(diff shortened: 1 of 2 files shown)


ipv8/attestation/trustchain/community.py
```diff
@@ -386,29 +388,32 @@ def process_half_block(self, blk, peer):
...
         self.logger.info("Received request block addressed to us (%s)", blk)
 
-        # determine if we want to sign this block
-        if not self.should_sign(blk):
-            self.logger.info("Not signing block %s", blk)
-            return succeed(None)
+        def on_should_sign_outcome(should_sign):
+            if not should_sign:
+                self.logger.info("Not signing block %s", blk)
+                return None
```
qstokkink: `addCallback` and `self.sign_block` return a `Deferred`. How are the unit tests even passing?

PR #520, opened 2019-06-08
Crawling now works without knowing chain length

devos50:
Our current TrustChain crawling depends on introduction requests that contain an integer with the chain length. However, we cannot always assume that the length of the chain is known when invoking `crawl_chain`. Therefore, if the `latest_block_num` argument is `None` and there is no information about the crawled peer in the database so far, we should first send out a request to fetch the latest block of that peer. By doing so, we find out the chain length and proceed with the crawl.

### DIFF
--- a/ipv8/test/attestation/trustchain/test_community.py
+++ b/ipv8/test/attestation/trustchain/test_community.py
@@ -1,17 +1,18 @@
 from __future__ import absolute_import
 
 from six.moves import xrange
+
 from twisted.internet.defer import inlineCallbacks
 
+from ...attestation.trustchain.test_block import TestBlock
+from ...base import TestBase
+from ...mocking.ipv8 import MockIPv8
 from ....attestation.trustchain.block import TrustChainBlock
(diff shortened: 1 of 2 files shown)


ipv8/test/attestation/trustchain/test_community.py
```diff
@@ -572,6 +573,32 @@ def test_chain_crawl(self):
...
         self.assertEqual(self.nodes[1].overlay.persistence.get_number_of_known_blocks(), 4)
 
+    @inlineCallbacks
+    def test_chain_crawl_unknown_length(self):
+        """
+        Test crawling a chain with unknown length
+        """
+        def create_blocks(num):
+            self.nodes[0].endpoint.close()
+            key = default_eccrypto.generate_key(u'very-low').pub().key_to_bin()
```
qstokkink: I'd like to phase out the use of the old M2Crypto format, could you use `u'curve25519'` here instead?

devos50: Done

PR #524, opened 2019-06-12
Added range proofs + attestation SDK refactoring

qstokkink:
This PR:

 - Refactors the Boneh exact value ZKP mechanism to a more logical folder structure
 - Refactors the internal attestation SDK API
 - Adds Boudot commitment checks (exact value and proof-of-square)
 - Adds the Peng and Bao range proof scheme
 - Fixes test timeout in the `test_*_attestation_big` tests


qstokkink:
If anyone wants to review this, the variables in the code correspond to the symbols in the following papers:

 - [An Efficient Range Proof Scheme](https://ieeexplore.ieee.org/abstract/document/5591457) by K. Peng and F. Bao
 - [Efficient Proofs that a Committed Number Lies in an Interval](https://link.springer.com/chapter/10.1007/3-540-45539-6_31) by F. Boudot

ipv8/attestation/wallet/pengbaorange/structs.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def __init__(self, m1, m2, m3, r1, r2, r3):
+        self.m1 = m1
+        self.m2 = m2
+        self.m3 = m3
+        self.r1 = r1
+        self.r2 = r2
+        self.r3 = r3
+
+    def generate_response(self, s, t):
```
xoriole: `s` and `t` cannot be zero, isn't it? A check or assertion somewhere would be good.

ipv8/test/attestation/wallet/test_attestation_community.py
```diff
@@ -200,6 +233,39 @@ def callback(rhash, values):
...
+    @inlineCallbacks
+    def test_verify_attestation_range(self):
+        """
+        Check if an attestation can be verified for id_metadata_range_18plus.
+        """
+        filename = os.path.join(os.path.dirname(__file__), 'attestation_range.txt')
+        with open(filename, 'r') as f:
+            serialized = unhexlify(f.read().strip())
+        attestation = PengBaoAttestation.unserialize_private(self.private_key, serialized, "id_metadata_range_18plus")
+        attestation_hash = unhexlify('b40c8734ba6c91a49670c1f0152c7f4dac2a8272')
```
xoriole: would be easier to follow if the attestation hash is derived instead of hardcoded one.
`hashlib.sha1(attestation.serialize()).digest()` or `attestation.get_hash()`

ipv8/attestation/wallet/pengbaorange/structs.py
qstokkink: Not only should `s` and `t` be non-zero, they are also expected to be "large integers". The paper doesn't really specify how large, other than strictly non-zero. I opted for `LARGE_INTEGER`.

I think this also answers your other two comments above. This function is only strictly called after the previous assertion that `s` and `t` are larger than `LARGE_INTEGER`. Also, from the other side, when generating `s` and `t` they are made larger than `LARGE_INTEGER`.

qstokkink: I can also add some additional comments to the code about this of course 👍

ipv8/test/attestation/wallet/test_attestation_community.py
qstokkink: I agree, but considering the amount of logic, I follow this way of working in the attestation unit tests. The unit tests will simply take too long if we go full circle in every test. I compromised by using premade attestations.

xoriole: alright then

qstokkink:
Rebased on master and added comments 👍

PR #532, opened 2019-06-24
Python 3 fixes

egbertbouman:


ipv8/attestation/trustchain/block.py
```diff
@@ -445,8 +445,8 @@ def __iter__(self):
         for key, value in self.__dict__.items():
             if key == 'key' or key == 'serializer' or key == 'crypto' or key == '_transaction':
                 continue
-            if isinstance(value, string_types) and key != "insert_time" and key != "type":
-                yield key, hexlify(value)
+            if isinstance(value, binary_type) and key != "insert_time" and key != "type":
```
qstokkink: Wait, this doesn't seem correct. If it's binary type then the matching with the unicode strings will never pass.

egbertbouman: Which matching?

qstokkink: `key != "insert_time" and key != "type"`

qstokkink: Under Python 3 you cannot compare bytes and str:

```python
> print(b"insert_time" == "insert_time")
False
```

egbertbouman: It's matching the key but converting the value. If the value is of type bytes, then it encodes to hex.

qstokkink: Oops, my bad.

PR #535, opened 2019-06-30
Improve attestation_endpoint responses

TimSpeelman:
Before: `attestation/?type=blah` returns `{ success: True }`.

After: it returns `{ error: "type argument incorrect"}`


devos50:
ok to test

TimSpeelman:
@qstokkink Fixed it. Also added all improvements suggested in #536

qstokkink:
Thanks 👍

PR #538, opened 2019-06-30
Separate concerns of Attestation Endpoint into API and Service

TimSpeelman:
@qstokkink, you recommended to use the Attestation Endpoint instead of IPv8 as a library, as it contained quite some logic. I feel like API logic (parsing, validating requests and formatting responses) should not be mixed with keeping track of state and executing the attestation/verification logic. 

Hence, I propose to split this. It was a lot easier than you made me believe, and the result is also more readable than the original. The formatting and parsing no longer clutter the attestation logic.

Furthermore, it provides the added benefit of being able to use AttestationService directly instead of over HTTP. When we add some hooks/callbacks to this service, we will even be able to make our attestation logic push instead of poll based, thereby improving latency.


qstokkink:
ok to test

qstokkink:
We'll do this later down the line, when we have Swagger endpoints and other cool stuff.

PR #547, opened 2019-08-09
Added ping measurements for peers

### LINKED ISSUE #546: Add ping times to Peer objects
We have come to appreciate the existence of latency data per peer. Therefore, ping data should be added into the deepest layer of IPv8.

The plan is to:
 
 - Add a `RequestCache` to `DiscoveryCommunity.create_ping()`.
 - Pop the corresponding `RequestCache` in `DiscoveryCommunity.on_pong()`.
 - Add a shift register of a certain length (5?) to the `Peer` object.
 - Push the difference between the `ping` and `pong` message from the `DiscoveryCommunity` into the `Peer` ping register.
 - Expose a method in the `Peer` class to get the median and/or average of the last known ping times.
 - Caveat 1: Make very sure the aforementioned method is thread safe.
 - Caveat 2: the `(IP+PORT) -> PUBLIC_KEY` mapping is unstable and expensive.
 - Caveat 3: the `DiscoveryCommunity` will become Twistified and will need to maintain a `request_cache` object.
 - Caveat 4: the `ping` and `pong` messages are unsigned.

qstokkink:
Fixes #546

### DIFF
--- a/ipv8/peerdiscovery/community.py
+++ b/ipv8/peerdiscovery/community.py
@@ -14,6 +14,7 @@ from ..messaging.payload import IntroductionRequestPayload
 from ..messaging.payload_headers import BinMemberAuthenticationPayload, GlobalTimeDistributionPayload
 from ..messaging.serialization import PackError
 from ..peer import Peer
+from ..requestcache import NumberCache, RequestCache
 from ..util import cast_to_bin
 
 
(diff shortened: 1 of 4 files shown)


ipv8/peerdiscovery/community.py
```diff
@@ -148,12 +177,14 @@ def create_similarity_response(self, identifier, peer):
 
         return self.custom_pack(peer, 2, [auth, dist, payload])
 
-    def create_ping(self):
+    def send_ping(self, peer):
```
devos50: I think it would be helpful if this `send_ping` would return a `Deferred` that fires if a `pong` message is received. Might be something for a later PR?

qstokkink: Nobody should be using this directly, rather the `Peer` object should be used to access the successful ping history.

devos50: What's the reason why application developers should not use it directly? It is a lightweight method to determine whether a peer is online and connected, which is helpful information in various applications.

qstokkink: This is not the correct interface for that. That should be on the `Peer` object level. Otherwise the IPv8 codebase turns into spaghetti.

devos50: I agree, the `Peer` object is indeed a better location for this.

PR #558, opened 2019-09-04
Added convenience struct for TrustChainBlock creation

### LINKED ISSUE #553: Change "data" to a dict or class instance.
https://github.com/Tribler/py-ipv8/blob/1b08fb2c8f05330bcf578af72f75c62a3c64af4c/ipv8/attestation/trustchain/block.py#L30

Changing data to a dictionary or a Data class instance would reduce the complexity of the code below (in a TrustChain constructor). It would also make it a little bit more clear and easier to read.

qstokkink:
Fixes #553

### DIFF
--- a/ipv8/attestation/trustchain/block.py
+++ b/ipv8/attestation/trustchain/block.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import time
 from binascii import hexlify
+from collections import namedtuple
 from hashlib import sha256
 
 from six import binary_type
@@ -14,7 +15,7 @@ from ...messaging.serialization import default_serializer
 from ...util import old_round
 
 
-GENESIS_HASH = b'0' * 32    # ID of the first block of the chain.
+GENESIS_HASH = b'0' * 32  # ID of the first block of the chain.
 GENESIS_SEQ = 1
 UNKNOWN_SEQ = 0
 EMPTY_SIG = b'0' * 64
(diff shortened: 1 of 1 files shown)


ipv8/attestation/trustchain/block.py
```diff
@@ -26,8 +27,73 @@ class TrustChainBlock(object):
...
+
+        **[3] sequence_number:** The height of this block in the chain of the intiator of this block.
+
+        *> type:* int
+
+        **[4] link_public_key:** The serialized public key of the counterparty of this block (binary data).
+
+        *> type:* bytes (Py3) or str (Py2)
+
+        **[5] link_sequence_number:** The height of this block in the chain of the intiator of this block,
```
ichorid: I don't get what's the difference between this `[5]` and `[3]`?

ipv8/attestation/trustchain/block.py
```diff
@@ -26,8 +27,73 @@ class TrustChainBlock(object):
...
+        **[7] signature:** The signature of the initiator of this block for this block (binary data).
+
+        *> type:* bytes (Py3) or str (Py2)
+
+        **[8] timestamp:** The time in milliseconds since the UNIX epoch that this block was created
+        (according to the initiator).
+
+        *> type:* int
+
+        **[9] insert_time:** The time in milliseconds since the UNIX epoch that this block was inserted into the local
```
ichorid: "The time in milliseconds since the UNIX epoch ~that~ when this block was inserted". Also, the same for `[8]`

ipv8/attestation/trustchain/block.py
```diff
@@ -26,8 +27,73 @@ class TrustChainBlock(object):
...
+
+        **[1] transaction:** Metadata dictionary, consisting of binary type strings.
+
+        *> type:* dict
+
+        **[2] public_key:** The serialized public key of the initiator of this block (binary data).
+
+        *> type:* bytes (Py3) or str (Py2)
+
+        **[3] sequence_number:** The height of this block in the chain of the intiator of this block.
```
ichorid: I wasn't able to immediately understand what "height of the block" means. Maybe change to something like "the serial number of this block"?

ipv8/attestation/trustchain/block.py
```diff
@@ -26,8 +27,73 @@ class TrustChainBlock(object):
     """
     Container for TrustChain block information
     """
+    Data = namedtuple('Data', ['type',
```
ichorid: :+1: for named tuples!

ipv8/attestation/trustchain/block.py
devos50: I would simply say 'the sequence number of this block in the chain of ...'

PR #569, opened 2019-09-12
WIP: Added latency community

qstokkink:
This PR adds the (highly experimental) latency community. The latency community is not enabled by default. To work with the latency community:

 1. Create a new `Network` class (**very important**) for the `LatencyCommunity`.
 2. Load the `LatencyCommunity` with **only** the `LatencyEdgeWalk` strategy into IPv8.
 3. Fetch your partners from the `community.accepted_proposals` set.

### LatencyEdgeWalk

The `LatencyEdgeWalk` strategy will collect a set number of peers from the bootstrap nodes/trackers, 30 by default. Next, the strategy will attempt to get successive network-unique ping-unique (0.05 seconds similarity) peers from each of these 30 edge roots, up until a given edge length (6 by default). As the edges are being grown, the strategy also provides ping measurements if needed.

After each step in the discovery process the `LatencyEdgeWalk` strategy will update its registered `overlay` by setting the `overlay`'s `possible_peers` attribute to the current list of peers included in the disconnected graph of edges (the `ancestry` datastructure).

Every so often (by default 30 seconds), the strategy performs a cleanup of all open connections to peers which are not included in the ancestry.

At the end of each cycle we distinguish two types of peers:

 - Open connections which cannot be included in a latency-unique edge, we will dub this the `total peers`.
 - Open connections which are part of a latency-unique edge, which we dub the `possible peers`.

### LatencyCommunity

The `LatencyCommunity` is in charge of the matching of peers. The Community predominantly uses the `PeerSelector` to advise any matching with others. The overlay itself periodically (by default every 5 seconds) (a) sends out proposals and keeps a window (30 by default) of optimal matches or (b) breaks the current worst match it has if the amount of matches is equal to the required amount (60 by default). Asynchronously the `LatencyCommunity` may decide to accept or reject incoming proposals.

The `LatencyCommunity` provides the last class of peers:

 - Peers which have bilaterraly agreed to form a matching, we dub these `matched peers`.

### PeerSelector

This class uses kerning to decide the error of including a peer into a given existing set of choices, according to a specified target distribution. The idea is that a programmer gives both (1) a continuous function (`lambda x: preferred_count / x` by default) and (2) a list of x-coordinates (bins) to evaluate the function (`0.001` through `2.001` with steps of `0.05` by default). This function need to be scaled precisely for the kerning, this is handled automatically by `generate_reference` for a given preferred peer count and the current binning distribution. We use Gaussian kernel density estimation.

### Gumby & Jenkins
My personal Jenkins experiment for this PR can be found here: https://jenkins-ci.tribler.org/job/pers/job/latency_noodle_qstokkink/. The source code for the experiment can be found here: https://github.com/qstokkink/gumby/tree/latency_overlay/experiments/latency_overlay.

### Future work
Even though the basic concept works, the CPU utilization should be improved and the matching algorithm will require some further love.

### DIFF
--- a/ipv8/peerdiscovery/latency/cache.py
+++ b/ipv8/peerdiscovery/latency/cache.py
@@ -1,0 +1,60 @@
+from __future__ import absolute_import
+
+from twisted.internet.defer import Deferred
+
+from ...requestcache import NumberCache
+
+
+class ProposalCache(NumberCache):
+    """
+    Cache for keeping track of a made proposal.
+    This cache concludes upon proposal (1) accept, (2) reject or (3) timeout.
+    """
+
+    def __init__(self, overlay, peer, nonce):
+        super(ProposalCache, self).__init__(overlay.request_cache, u"proposal-cache",
+                                            self.number_from_pk_nonce(peer.mid, nonce))
+        self.overlay = overlay
+        self.peer = peer
+
+    @classmethod
+    def number_from_pk_nonce(cls, public_key, nonce):
+        """
+        Create an identifier from a public key and a nonce.
+
+        :param public_key: the counterparty public key
+        :type public_key: str or bytes
+        :param nonce: the nonce for this proposal
+        :type nonce: str or bytes
+        :return: the identifier for the given parameters
+        :rtype: int
+        """
+        number = nonce
+        for c in public_key:
+            number <<= 8
+            number += c if isinstance(c, int) else ord(c)
+        return number
+
+    def on_timeout(self):
+        """
+        When timing out, we remove this proposal from the open proposals.
+
+        :returns: None
+        """
+        try:
+            self.overlay.open_proposals.remove(self.peer)
+        except KeyError:
+            self.overlay.logger.debug("Proposal timed out, but peer already removed.")
+
+
+class StatsRequestCache(NumberCache):
+    """
+    Cache for waiting for a stats response.
+    """
+
+    def __init__(self, overlay):
+        super(StatsRequestCache, self).__init__(overlay.request_cache, u"stats-request", overlay.claim_global_time())
+        self.deferred = Deferred()
+
+    def on_timeout(self):
+        self.deferred.errback(None)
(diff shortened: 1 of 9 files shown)


qstokkink:
Good news:

 - Optimized all magic parameters in 300 peer Gumby experiment to get best performance
 - Unit tested
 - Results as expected for a port-distance-based artificial delay

afbeelding

Bad news:

 - This thing eats CPU for breakfast

synctext:
Solid progress!
How could we measure wild deployed code and establish that it fails/prevails?

qstokkink:
> 
> How could we measure wild deployed code and establish that it fails/prevails?

As this is made to not leak any information about your personal network state, we cannot directly inspect anyone. However, we can test if the protocol is holding up if we create a sybil region and see how often we end up in our own sybil region from a different location. The effectiveness of the algorithm will also greatly depend on the amount of users in different physical locations though..

We could also hook up a network message just for debugging purposes, which would give you the peer counts in the three distinct peer classes this algorithm has. Though this would have a (marginal) impact on the security of the protocol.

synctext:
Stats msg in debug version. That sounds good.

qstokkink:
I'm going to give up on finishing it today. I cannot get the matching algorithm to behave reliably with differing peer latency distributions.

qstokkink:
I'm going to clean up this PR by:

 - Extracting the Network and Community optimizations into its own PR.
 - Rewrite the history for this branch.

Furthermore I'll write a description of how to use this functionality today.

We probably want to reduce CPU usage in a different PR, before using this functionality though.

qstokkink:
Ready for review, let me know if the instructions in the O.P. require supplemental info.

qstokkink:
retest this please

qstokkink:
Marking this as W.I.P. to fix the matching issues within this PR.

qstokkink:
Probably not a good idea to merge this, closing for now.

PR #571, opened 2019-09-17
Network optimizations

qstokkink:
This PR contains optimization of some `Network` methods and also changes `Community` behavior to respond to already known peers upon receiving an introduction request.

### DIFF
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -37,6 +37,10 @@ class Network(object):
         self.reverse_intro_lookup = OrderedDict()
         self.reverse_intro_cache_size = 500
 
+        # Cache of service_id -> [Peer]
+        self.reverse_service_lookup = OrderedDict()
+        self.reverse_service_cache_size = 500
+
     def discover_address(self, peer, address, service=None):
         """
         A peer has introduced us to another IP address.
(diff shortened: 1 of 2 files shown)


qstokkink:
Seems like `pyasn1` is not installed on the Mac.

qstokkink:
retest this please

qstokkink:
retest this please

ipv8/peerdiscovery/network.py
```diff
@@ -183,13 +227,10 @@ def remove_by_address(self, address):
...
-            for i in range(len(self.verified_peers)):
-                if self.verified_peers[i].address == address:
-                    to_remove.insert(0, i)
-                    self.services_per_peer.pop(self.verified_peers[i].mid, None)
-            for index in to_remove:
-                self.verified_peers.pop(index)
+            # Note that the services_per_peer will never be 0, we abuse the lazy `or` to pop the peers from
+            # the services_per_peer mapping if they are no longer included. This is fast.
+            self.verified_peers = [peer for peer in self.verified_peers
+                                   if peer.address != address or self.services_per_peer.pop(peer.mid, None) == 0]
```
xoriole: `self.services_per_peer.pop(peer.mid, None) == 0` 
should this be comparison of the length instead?

qstokkink: No, the reason for its existence is that it will never be `0`:

1. Because of the lazy-or this will only execute for peers which need to be deleted.
2. For each peer to be deleted we pop it out of the `services_per_peer`, but as the return value is never `0` and thus this part of the check is always `False`, it is not re-added/un-deleted (this always produces the same list as `[peer for peer in self.verified_peers if peer.address != address]`).
[2.b. We do it like this because Python can execute mappings faster than explicit loops.]

PR #574, opened 2019-09-19
Binary decoding in JSON

xoriole:
Adds a utility method  `ensure_serializable()` to json_util that decodes binary data sot that the result is compatible with json.dumps(). This utility is used to fix TypeError in `trustchain_endpoint`.

Fixes https://github.com/Tribler/tribler/issues/4845 (Tribler)

### DIFF
--- a/ipv8/REST/json_util.py
+++ b/ipv8/REST/json_util.py
@@ -5,7 +5,7 @@ from collections import Iterable
 
 from six import string_types
 
-__all__ = ['dumps', 'loads']
+__all__ = ['dumps', 'loads', 'ensure_serializable']
 
 
 def _is_undumpable(obj):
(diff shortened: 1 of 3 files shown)


ipv8/REST/json_util.py
```diff
@@ -122,3 +137,25 @@ def load(fp, *args, **kwargs):
     :return: the Python object(s) extracted from the JSON input.
     """
     return json.load(fp, *args, **kwargs)
+
+
+def ensure_serializable(value):
```
qstokkink: This can become **very expensive** if we do this on every dump call. This rewards bad programming and slows down well written code. Please just error out like before: when a programmer feeds in a value which is non-json.

xoriole: Yes, `ensure_serializable` can be an expensive operation but with `decode_bytes=False` set by default should ensure this serializable assurance only happens when it is specifically set to True.

xoriole:
retest this please

qstokkink:
Why did you remove the custom JSONEncoder? If you slim this PR down to just the `JSONEncoder` that converts all `bytes` to `str` using `.decode('utf-8')`, that would solve all of the `bytes` related problems.

The `ensure_serializable` should still go though.

xoriole:
Custom encoder only helps with value objects in case of a dictionary. If there is a binary key in the dict then it will still fail which is the case for the associated issue. For now, I'll close this PR and fix the issue differently without touching `json_util.py`.

PR #575, opened 2019-09-26
Trustchain blocks are encoded when converting to dictionary

xoriole:
Trustchain blocks are encoded when converting to dictionary. This makes sure that the blocks are serializable for json dumping.

Fixes Tribler/tribler#4845

### DIFF
--- a/ipv8/attestation/trustchain/block.py
+++ b/ipv8/attestation/trustchain/block.py
@@ -511,10 +511,18 @@ class TrustChainBlock(object):
         for key, value in self.__dict__.items():
             if key == 'key' or key == 'serializer' or key == 'crypto' or key == '_transaction':
                 continue
-            if isinstance(value, binary_type) and key != "insert_time" and key != "type":
-                yield key, hexlify(value).decode('utf-8')
-            else:
-                yield key, value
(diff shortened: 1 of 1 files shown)


ipv8/attestation/trustchain/block.py
```diff
@@ -511,10 +511,18 @@ def __iter__(self):
         for key, value in self.__dict__.items():
             if key == 'key' or key == 'serializer' or key == 'crypto' or key == '_transaction':
                 continue
-            if isinstance(value, binary_type) and key != "insert_time" and key != "type":
-                yield key, hexlify(value).decode('utf-8')
-            else:
-                yield key, value
+            key = key.decode('utf-8') if isinstance(key, binary_type) else key
+
+            if key == 'transaction' and isinstance(value, dict):
```
qstokkink: You're fighting a battle you cannot win: transactions are JSON objects. This implementation will already error out on the first list of binary types. For any automatic fix, there is a nasty edge case which is not covered.

Instead, the fix should be on another level: the block class should define a transaction format which is proper UTF-8. In fact, the fact that someone apparently has a transaction that is not UTF-8 signifies some major problem in one of the TrustChain wallets/block listeners.

PR #577, opened 2019-09-30
Added relay_early flag to cell payload

egbertbouman:


egbertbouman:
It turns out that I can't make the required changes without breaking compatibility with earlier versions. Since I'm not sure whether this is a good time to break the protocol, I'm closing this PR for now.

PR #597, opened 2019-10-24
Fix puncture without DiscoveryCommunity

### LINKED ISSUE #596: Tracker sending puncture_request with wrong prefix
The tracker plugin rewrites packets with the requested packet prefix:

https://github.com/Tribler/py-ipv8/blob/4aceb300069cde738c00b332e1495f347947a67e/twisted/plugins/tracker_plugin.py#L92

**However**, `create_introduction_response` also sends a `puncture_request` message over with a non-overwritten prefix. This causes the fact that only the `DiscoveryCommunity` will only perform puncturing. We need to support standalone (no `DiscoveryCommunity` loaded) communities as well.

qstokkink:
Fixes #596

### DIFF
--- a/ipv8/community.py
+++ b/ipv8/community.py
@@ -190,7 +190,7 @@ class Community(EZPackOverlay):
         return self._ez_pack(self._prefix, 246, [auth, dist, payload])
 
     def create_introduction_response(self, lan_socket_address, socket_address, identifier,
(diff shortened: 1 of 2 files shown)


qstokkink:
Flake8 can't really handle our Twisted plugin formatting, please ignore. Marking as ready.

PR #622, opened 2019-11-19
Added VariablePayload compilation decorator

### LINKED ISSUE #619: Compile VariablePayload
Right now the `VariablePayload` is easy to use, but around 10x slower than the normal `Payload`. Through some clever use of `type`, `compile` and a class decorator we can "compile" `VariablePayload` definitions at runtime/class-load-time into normal `Payload` definitions, making them much faster.

qstokkink:
Fixes #619 
Sister PR of #623

This just-in-time compilation of `VariablePayload` objects on class load time is much faster (slightly less than a factor of 10).
Look at the source code at your own risk, I tried to document as much as possible.


qstokkink:
I'll make a PR for the `asyncio` branch now.

PR #629, opened 2019-12-09
Fixed peer inequality check in Python 2.7

devos50:
In Python 2.7, there is a slightly different behaviour of the not equal operation (!=) compared to Python 3. In particular, in Python 2.7, statements like peerA == peerB would evaluate to True but statements like peerA != peerB would not evaluate to False, due to the absence of a __ne__ method in the Peer class. This PR fixes this issue and modifies an existing test to test this change.

### DIFF
--- a/ipv8/peer.py
+++ b/ipv8/peer.py
@@ -80,7 +80,12 @@ class Peer(object):
     def __eq__(self, other):
         if not isinstance(other, Peer):
             return False
-        return (self.public_key.key_to_bin() == other.public_key.key_to_bin())
+        return self.public_key.key_to_bin() == other.public_key.key_to_bin()
+
+    def __ne__(self, other):
+        if not isinstance(other, Peer):
+            return True
+        return self.public_key.key_to_bin() != other.public_key.key_to_bin()
 
     def __str__(self):
         return 'Peer<%s:%d, %s>' % (self.address[0], self.address[1], b64encode(self.mid).decode('utf-8'))
--- a/ipv8/test/test_peer.py
+++ b/ipv8/test/test_peer.py
@@ -52,7 +52,8 @@ class TestPeer(unittest.TestCase):
         """
         other = Peer(self.peer.key, self.peer.address)
 
-        self.assertEqual(self.peer, other)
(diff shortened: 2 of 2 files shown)


devos50:
Well... it seems that this breaks the DHT tests. I guess it has something to do with the expected behavior of the `Node` class, which is a subclass of `Peer`.

ipv8/peer.py
```diff
@@ -80,7 +80,12 @@ def __hash__(self):
     def __eq__(self, other):
         if not isinstance(other, Peer):
             return False
-        return (self.public_key.key_to_bin() == other.public_key.key_to_bin())
+        return self.public_key.key_to_bin() == other.public_key.key_to_bin()
+
+    def __ne__(self, other):
+        if not isinstance(other, Peer):
+            return False
```
qstokkink: Shouldn't this return `True`?

devos50: Good catch, thanks!

devos50:
I observed this error in the tests:

```
======================================================================
FAIL: Check if we can bootstrap our peerdiscovery.
  File "/Library/Python/2.7/site-packages/twisted/internet/defer.py", line 1299, in _inlineCallbacks
    result = g.send(result)
  File "/Users/tribler/Documents/workspace2/workspace/ipv8/test_ipv8_PR_unittests_mac/ipv8/test/peerdiscovery/test_community.py", line 79, in test_bootstrap
    self.assertEqual(len(intros), 1)
  File "/Library/Python/2.7/site-packages/twisted/trial/_synctest.py", line 425, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 506, in _baseAssertEqual
    raise self.failureException(msg)
FailTest: 0 != 1
```

qstokkink:
Seems fine, please open the sister PR to the `asyncio` branch and we can merge.

PR #630, opened 2019-12-11
Fix dictionary changed size errors

egbertbouman:


qstokkink:
Is this only an issue on the master branch?

egbertbouman:
@qstokkink Starting from Tribler 7.5, we no longer have a separate thread for making calls to the tunnel community, so I figured it's no longer needed.

qstokkink:
I was thinking the REST API would cause issues, but I guess that also runs on the asyncio event loop. Should be good then 👍

ichorid:
@egbertbouman, does it fix #4955 ?

egbertbouman:
@ichorid Yes, it does.

PR #631, opened 2019-12-11
Fix issue with hidden services tests

egbertbouman:


qstokkink:
Is this only an issue on the `asyncio` branch?

egbertbouman:
@qstokkink I think so (unless twisted was eating the errors).

Issue #632, opened 2019-12-12
builtins.ValueError: Invalid secret key (caused by libnacl 1.7)


devos50:
When running Tribler (or IPv8) with libnacl 1.7, I observe the following stack trace:

```
2019-12-12T10:20:49+0000 [twisted.internet.defer#critical] Unhandled error in Deferred:
2019-12-12T10:20:49+0000 [twisted.internet.defer#critical] 
	Traceback (most recent call last):
	  File "/home/noodle/tribler/Tribler/pyipv8/ipv8/messaging/anonymization/community.py", line 74, in inner_wrapper
	    return func(inner_self, inner_source_address, *combo)
	  File "/home/noodle/tribler/Tribler/pyipv8/ipv8/messaging/anonymization/community.py", line 749, in on_create
	    addCallback(self.should_join_circuit(payload, source_address), determined_to_join)
	  File "/home/noodle/tribler/Tribler/pyipv8/ipv8/util.py", line 76, in addCallback
	    return deferred.addCallbacks(callback, errback)
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 311, in addCallbacks
	    self._runCallbacks()
	--- <exception caught here> ---
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 654, in _runCallbacks
	    current.result = callback(current.result, *args, **kw)
	  File "/home/noodle/tribler/Tribler/pyipv8/ipv8/messaging/anonymization/community.py", line 745, in determined_to_join
	    self.join_circuit(payload, source_address)
	  File "/home/noodle/tribler/Tribler/community/triblertunnel/community.py", line 458, in join_circuit
	    super(TriblerTunnelCommunity, self).join_circuit(create_payload, previous_node_address)
	  File "/home/noodle/tribler/Tribler/pyipv8/ipv8/messaging/anonymization/community.py", line 717, in join_circuit
	    shared_secret, key, auth = self.crypto.generate_diffie_shared_secret(create_payload.key)
	  File "/home/noodle/tribler/Tribler/pyipv8/ipv8/messaging/anonymization/tunnelcrypto.py", line 47, in generate_diffie_shared_secret
	    AUTH = libnacl.crypto_auth(Y, shared_secret)
	  File "/home/noodle/.local/lib/python3.7/site-packages/libnacl/__init__.py", line 873, in crypto_auth
	    raise ValueError('Invalid secret key')
	builtins.ValueError: Invalid secret key
```

[...]

PR #633, opened 2019-12-13
Make TunnelCrypto compatible with libnacl 1.7

egbertbouman:
Fixes https://github.com/Tribler/py-ipv8/issues/632. Sister PR of https://github.com/Tribler/py-ipv8/pull/636.

### DIFF
--- a/ipv8/messaging/anonymization/tunnelcrypto.py
+++ b/ipv8/messaging/anonymization/tunnelcrypto.py
@@ -43,12 +43,12 @@ class TunnelCrypto(ECCrypto):
         Y = tmp_key.key.pk
         shared_secret = libnacl.crypto_box_beforenm(dh_received, y) + libnacl.crypto_box_beforenm(dh_received, key.key.sk)
 
-        AUTH = libnacl.crypto_auth(Y, shared_secret)
+        AUTH = libnacl.crypto_auth(Y, shared_secret[:32])
         return shared_secret, Y, AUTH
 
     def verify_and_generate_shared_secret(self, dh_secret, dh_received, auth, B):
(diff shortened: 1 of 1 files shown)


ipv8/messaging/anonymization/tunnelcrypto.py
```diff
@@ -43,12 +43,12 @@ def generate_diffie_shared_secret(self, dh_received, key=None):
         Y = tmp_key.key.pk
         shared_secret = libnacl.crypto_box_beforenm(dh_received, y) + libnacl.crypto_box_beforenm(dh_received, key.key.sk)
 
-        AUTH = libnacl.crypto_auth(Y, shared_secret)
+        AUTH = libnacl.crypto_auth(Y, shared_secret[:32])
```
qstokkink: This seems equivalent to not appending `libnacl.crypto_box_beforenm(dh_received, key.key.sk)` to the shared secret. Was the remainder of the message never checked by `libnacl` in the old version?

Seems kind of fishy, please provide 4 unit tests with `1.6` -> `1.6`, `1.6` -> `1.7`, `1.7` -> `1.6` and `1.7` -> `1.7` behavior.

egbertbouman: The `shared_secret` is also used for generating session keys, not just for the MAC of the `created` cell.

In `1.6`, the function apparently just drops any data beyond the first 32 bytes. I would prefer calculating the MAC using a different key, but that would break compatibility. Since I'm already planning to break to the protocol for Tribler 7.5, we could just wait until then.

qstokkink: Alright, if you checked it, that's fine. Please make a sister PR to the `asyncio` branch so we can merge this.

PR #637, opened 2019-12-24
Small usability improvements for registering anonymous tasks

egbertbouman:


ipv8/taskmanager.py
```diff
@@ -99,7 +116,7 @@ def register_anonymous_task(self, basename, task, *args, **kwargs):
         """
         Wrapper for register_task to derive a unique name from the basename.
         """
-        return self.register_task(basename + str(id(task)), task, *args, **kwargs)
```
qstokkink: I assume you want to register the same task object twice? [Otherwise `id()` is a better choice.](https://docs.python.org/3/library/functions.html#id)

However, in that case I would still prefer the 100% foolproof non-probabilistic: add random characters to the `id()` as long as the current name is already registered. Basically: why even give it a chance to fail if we can make it guaranteed to work?

egbertbouman: After the move to asyncio I added the ability to register functions as well, instead of just tasks. This avoids having to wrap everything in `ensure_future`. However, this means that if you want to register the same function as an anonymous task more then once (which happens more often then you would think), it won't work.

Anyway, how about just adding a counter? That should work no matter what.

qstokkink: A counter would work, that is fine by me.

PR #645, opened 2019-12-25
Fixed bootstrap test wrong assertion

qstokkink:
Fixes #635

Issue #641, opened 2019-12-25
test_build_proofs failing


devos50:
Observed on Linux (https://jenkins-ci.tribler.org/job/ipv8/job/test_ipv8_PR_unittests_linux/1497/console):

```
test_task_with_exception_ignore (ipv8.test.test_taskmanager.TestTaskManager) ... ERROR:TaskManager:Task resulted in error: division by zero
ok

======================================================================
FAIL: test_build_proofs (ipv8.test.attestation.wallet.irmaexact.test_builder.TestBuilder)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jenkins/.local/lib/python3.7/site-packages/asynctest/case.py", line 297, in run
    self._run_test_method(testMethod)
  File "/home/jenkins/.local/lib/python3.7/site-packages/asynctest/case.py", line 352, in _run_test_method
    result = method()
  File "/home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux/ipv8/test/attestation/wallet/irmaexact/test_builder.py", line 85, in test_build_proofs
    self.assertFalse(proof.Verify(self.testPubK, sig, 10, nonce))
AssertionError: True is not false
```

qstokkink:
This is a source port from the IRMA project: if we "fix" it, we break the compatibility.

I was thinking of skipping these tests, but this does not happen that often.

If this starts occurring more often I'll disable the tests after all. For now, won't fix.

qstokkink:
After a second look, this appears to happen after the PR tests have been set to run on Python 3 for some reason (who did this?). Probably a Python 3 thing, this warrants further investigation.

If it turns out to not be Python 3, but the IRMA protocol this will go back to `wont address`.

qstokkink:
OP also seems to be Python 3. I'll run some stress tests.

qstokkink:
Not just Python 3, also confirmed with `unittest` test runner on Python 2.7:

```
======================================================================
FAIL: test_build_proofs (ipv8.test.attestation.wallet.irmaexact.test_builder.TestBuilder)
test_build_proofs
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/quinten/.local/lib/python2.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/home/quinten/.local/lib/python2.7/site-packages/twisted/internet/utils.py", line 201, in runWithWarningsSuppressed
[...]

PR #646, opened 2019-12-25
Fixed bootstrap test wrong assertion

qstokkink:
Sister PR of #645


qstokkink:
retest this please, #641 is definitely annoying and only happening on Python 3.

PR #647, opened 2019-12-25
Fix IRMA build_proofs test

qstokkink:
Fixes #641

PR #649, opened 2019-12-25
Unite system path modifications as __scriptpath__.py files

### LINKED ISSUE #642: Don't append but insert into sys path
#### Platform #####
All
#### Expected behavior ####
Append paths to system path in `setup.py`.
#### Actual behavior #####
System path is being set up in various modules in the code, resulting in duplicates and a setup that's hard to keep track of.
https://github.com/Tribler/py-ipv8/search?q=sys.path.append&unscoped_q=sys.path.append

An example of a duplicate is `twisted/plugins/ipv8_plugin.py` and `twisted/plugins/trustchain_crawler_plugin.py`, both of which add `twisted` to the system path. Is it possible to gather them all up in setup.py?

As a minor note, appending to the `sys path` means that "an installed package will take precedence over the local package in development (and hair pulling may ensue)." (as the comment to [this](https://stackoverflow.com/questions/1893598/pythonpath-vs-sys-path/1893663#1893663) stackoverflow points out. `sys.path.insert(1,` can overcome it.

Solomon1732:
Instead of modifying system path on a per-module basis it is instead modified in a __scriptpath__.py file that is imported into the desired module.

In addition instead of appending into the system path the modification inserts into it to prevent path overrides.

Fixes #642

### DIFF
--- a/scripts/__scriptpath__.py
+++ b/scripts/__scriptpath__.py
@@ -1,0 +1,4 @@
+import pathlib
+import sys
+
+sys.path.insert(1, pathlib.Path(__file__, "..").resolve())
--- a/scripts/exitnode_ipv8_only_plugin.py
+++ b/scripts/exitnode_ipv8_only_plugin.py
@@ -2,12 +2,18 @@
 This script enables to start IPv8 headless.
 """
 import argparse
-import os
 import signal
 import sys
 from asyncio import all_tasks, ensure_future, gather, get_event_loop, sleep
 
-sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+# Check if we are running from the root directory
+# If not, modify our path so that we can import IPv8
+try:
+    import ipv8
+    del ipv8
+except ImportError:
+    import __scriptpath__  # noqa: F401
+
 
 from ipv8.REST.rest_manager import RESTManager
 from ipv8.configuration import get_default_configuration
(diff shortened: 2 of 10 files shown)


qstokkink:
ok to test

qstokkink:
Making flake8 skip the side-effect import should be fine:

```python
import __scriptpath__  # noqa: F401
```

Solomon1732:
```
FAIL: test_build_proofs (ipv8.test.attestation.wallet.irmaexact.test_builder.TestBuilder)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/asynctest/case.py", line 297, in run
    self._run_test_method(testMethod)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/asynctest/case.py", line 352, in _run_test_method
    result = method()
  File "/Users/tribler/Documents/workspace2/workspace/ipv8/test_ipv8_PR_unittests_mac/ipv8/test/attestation/wallet/irmaexact/test_builder.py", line 85, in test_build_proofs
    self.assertFalse(proof.Verify(self.testPubK, sig, 10, nonce))
AssertionError: True is not false
```
Hmm. Looks like the failure is related to #641

qstokkink:
Yeah I fixed that in #647, but the other Tribler admins appear to have stuff to do during Christmas 😕

PR #655, opened 2019-12-29
Change bare exceptions.

### LINKED ISSUE #650: Bare excepts may lead to unexpected behavior
#### Platform ####
All
#### Expected behavior ####
Catch specific exceptions. Or at most catch `except Exception:`.
#### Actual behavior ####
Bare excepts, whose catching may lead to unexpected behavior (such as catching `RuntimeError` or `SystemExit`).

The locations of the exceptions is provided in a list of links in the paste bin link below
https://gist.github.com/Solomon1732/73e4637aa057e99ae7c860bd9fc4dc88

Solomon1732:
Fixes #650 . Changes the exceptions according to what was specified in #650 . In addition, changed a few %-formatted strings to f-strings and removed a few redundant u-prefix from some strings.

### DIFF
--- a/ipv8/REST/dht_endpoint.py
+++ b/ipv8/REST/dht_endpoint.py
@@ -388,7 +388,7 @@ class DHTBlockPublisher(BlockListener):
                 # Try to add the current chunk to the DHT; if it works, move to the next, otherwise retry
                 try:
                     await self.dht.store_value(self._hashed_dht_key, blob_chunk[0])
-                except:
+                except (RuntimeError, PackError):
                     chunk_attempt += 1
                 else:
                     slice_pointer += self.CHUNK_SIZE
--- a/ipv8/REST/dht_endpoint.py
+++ b/ipv8/REST/dht_endpoint.py
@@ -15,6 +15,7 @@ from .schema import DHTValueSchema, DefaultResponseSchema, schema
 from ..attestation.trustchain.community import TrustChainCommunity
 from ..attestation.trustchain.listener import BlockListener
 from ..attestation.trustchain.payload import DHTBlockPayload
+from ..dht import DHTError
 from ..dht.community import DHTCommunity, MAX_ENTRY_SIZE, MAX_VALUES_IN_FIND
 from ..dht.discovery import DHTDiscoveryCommunity
 from ..keyvault.public.libnaclkey import LibNaCLPK
@@ -388,7 +389,7 @@ class DHTBlockPublisher(BlockListener):
                 # Try to add the current chunk to the DHT; if it works, move to the next, otherwise retry
                 try:
                     await self.dht.store_value(self._hashed_dht_key, blob_chunk[0])
-                except (RuntimeError, PackError):
+                except (DHTError, PackError):
                     chunk_attempt += 1
                 else:
                     slice_pointer += self.CHUNK_SIZE
--- a/ipv8/attestation/trustchain/block.py
+++ b/ipv8/attestation/trustchain/block.py
@@ -9,7 +9,6 @@ from ...database import database_blob
 from ...keyvault.crypto import default_eccrypto
 from ...messaging.deprecated.encoding import decode, encode
 from ...messaging.serialization import PackError, default_serializer
-from ...util import old_round
 
 GENESIS_HASH = b'0' * 32  # ID of the first block of the chain.
 GENESIS_SEQ = 1
@@ -17,6 +16,7 @@ UNKNOWN_SEQ = 0
 EMPTY_SIG = b'0' * 64
 EMPTY_PK = b'0' * 74
 ANY_COUNTERPARTY_PK = EMPTY_PK
+SKIP_ATTRIBUTES = {'key', 'serializer', 'crypto', '_transaction', '_logger'}
 
 
 class TrustChainBlock(object):
@@ -106,7 +106,7 @@ class TrustChainBlock(object):
             # validation
             self.previous_hash = GENESIS_HASH
             self.signature = EMPTY_SIG
-            self.timestamp = int(old_round(time.time() * 1000))
+            self.timestamp = int(time.time() * 1000)
             # debug stuff
             self.insert_time = None
         else:
@@ -125,7 +125,7 @@ class TrustChainBlock(object):
             self.signature = self.signature if isinstance(self.signature, bytes) else bytes(self.signature)
         self.hash = self.calculate_hash()
         self.crypto = default_eccrypto
-        self.logger = logging.getLogger(self.__class__.__name__)
+        self._logger = logging.getLogger(self.__class__.__name__)
 
     @classmethod
     def from_payload(cls, payload, serializer):
@@ -332,7 +332,7 @@ class TrustChainBlock(object):
             try:
                 pck = self.pack(signature=False)
             except PackError as e:
-                self.logger.debug("Failed to pack 'self.pack' (error %s)", e)
+                self._logger.debug("Failed to pack 'self.pack' (error %s)", e)
                 pck = None
             if pck is None or not self.crypto.is_valid_signature(
                     self.crypto.key_from_public_bin(self.public_key), pck, self.signature):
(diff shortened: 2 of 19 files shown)


Solomon1732:
I am at lost as to why this additional `Unite sys-path modifications as __scriptpath__.py ` commit appears in all my commits. I'll remove it.

devos50:
ok to test

devos50:
add to whitelist

ipv8/attestation/trustchain/community.py
```diff
@@ -286,10 +286,7 @@ def sign_block(self, peer, public_key=EMPTY_PK, block_type=b'unknown', transacti
         """
         peer = Peer(payload.public_key, source_address)
         block = self.get_block_class(payload.type).from_payload(payload, self.serializer)
-        try:
-            await self.process_half_block(block, peer)
-        except:
-            pass
+        await self.process_half_block(block, peer)
```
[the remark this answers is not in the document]
Solomon1732: Ah, right. It raises `RuntimeError`. Will change it.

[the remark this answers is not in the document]
Solomon1732: Wait, now I remember why I removed it. @qstokkink said in #650 
> Message handling is already in a general try .. except block, why is this even here?

Therefor it was deemed safe to remove.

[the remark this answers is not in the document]
Solomon1732: If it were to be logged, which severity should it be? INFO?

Solomon1732:
For some reason me adding logging [here](https://github.com/Solomon1732/py-ipv8/blob/f_no_bare_excepts/ipv8/attestation/trustchain/block.py#L335) causes it to fail the tests. Why?

devos50:
The error in the tests:

```
======================================================================
FAIL: Check that the iterator of a Block has all of the required keys without duplicates.
  File "/home/jenkins/.local/lib/python3.7/site-packages/asynctest/case.py", line 297, in run
    self._run_test_method(testMethod)
  File "/home/jenkins/.local/lib/python3.7/site-packages/asynctest/case.py", line 352, in _run_test_method
    result = method()
  File "/home/jenkins/ci/workspace/ipv8/test_ipv8_PR_unittests_linux_py36/ipv8/test/attestation/trustchain/test_block.py", line 795, in test_iter
    self.assertSetEqual(set(block_keys), expected_keys)
AssertionError: Items in the first set but not the second:
'logger'
```

The `__iter__` method in `block.py` goes over all the attributes of the block and prints them. Since you've added `self.logger` as attribute to the class, it will also try to serialise this attribute, which it fails to do. You should add the `logger` as an exception in the `__iter__` method 👍

Solomon1732:
Oh, that's why. Thank you 😄 How do I add it as an exception? 😅 Should I add it as `getlogger(__name__)` instead of `getlogger(__class__.__name__)`?

Solomon1732:
Apparently not

Solomon1732:
Nope. Doesn't work.

Solomon1732:
I'm blind. There's already an `__iter__` method. My bad.

ipv8/attestation/trustchain/community.py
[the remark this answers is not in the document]
Solomon1732: Added as INFO level logging.

ipv8/attestation/trustchain/block.py
```diff
@@ -504,7 +507,7 @@ def __iter__(self):
         :return: generator to iterate over all properties of this block
         """
         for key, value in self.__dict__.items():
-            if key == 'key' or key == 'serializer' or key == 'crypto' or key == '_transaction':
```
qstokkink: Could you bind the set `{'key', 'serializer', 'crypto', '_transaction', '_logger'}` to a module level variable? It's faster.

ipv8/messaging/anonymization/community.py
```diff
@@ -763,11 +763,17 @@ def on_created(self, source_address, payload, _):
             self.directions[request.from_circuit_id] = EXIT_NODE
             self.remove_exit_socket(request.from_circuit_id)
 
-            self.send_cell([relay.peer], u"extended", ExtendedPayload(relay.circuit_id,
-                                                                      payload.key,
-                                                                      payload.auth,
-                                                                      payload.candidate_list_enc))
-        elif self.request_cache.has(u"retry", payload.circuit_id):
+            self.send_cell(
```
qstokkink: I'm not a fan of this formatting style, please change it back.

ipv8/messaging/anonymization/tunnel.py
```diff
@@ -72,9 +72,9 @@ def could_be_udp_tracker(data):
     @staticmethod
     def could_be_dht(data):
         try:
-            if len(data) > 1 and data[0:1] == 'd' and data[-1:] == 'e':
```
qstokkink: Oh, you found a bug. `data` is a bytes object, so it should be compared to `b'd'` and `b'e'` (something we missed when porting to Python 3).

Other than that, the old code was correct (slicing over a `bytes` produces `bytes`, indices produce `int`):

```python
a = b"data"
print(a[0] == b"d")
>> False
print(a[0:1] == b"d")
>> True
```

ipv8/REST/dht_endpoint.py
```diff
@@ -388,7 +388,7 @@ def should_sign(self, block):
                 # Try to add the current chunk to the DHT; if it works, move to the next, otherwise retry
                 try:
                     await self.dht.store_value(self._hashed_dht_key, blob_chunk[0])
-                except:
+                except (RuntimeError, PackError):
```
egbertbouman: The `store_value` function throws a `DHTError` instead of a `RuntimeError`.

Solomon1732: Ah, oops. Will fix 👍

ipv8/messaging/anonymization/tunnel.py
Solomon1732: Ohhhh, ok. Gotcha

ipv8/messaging/anonymization/community.py
Solomon1732: I did it like this because the indentation was a flake8 violation for some reason, but will do.

ipv8/attestation/trustchain/block.py
Solomon1732: Will do.

ipv8/messaging/anonymization/tunnel.py
Solomon1732: Done

ipv8/REST/dht_endpoint.py
Solomon1732: Fixed

ipv8/attestation/trustchain/block.py
Solomon1732: Done.

ipv8/messaging/anonymization/community.py
Solomon1732: Reverted

ipv8/attestation/trustchain/block.py
```diff
@@ -16,6 +17,7 @@
 EMPTY_SIG = b'0' * 64
 EMPTY_PK = b'0' * 74
 ANY_COUNTERPARTY_PK = EMPTY_PK
+SKIP_ATTRIBUTES = frozenset({'key', 'serializer', 'crypto', '_transaction', '_logger'})
```
qstokkink: Why not just a normal `set`? `frozenset` is actually slower: https://stackoverflow.com/q/36555214 . Also, if any programmer subclasses the block, he cannot add to the `SKIP_ATTRIBUTES` set (without `setattr` magic).

Solomon1732: Ah. I thought mutability was not desired. Now that I know better I will change it 👍

Solomon1732: Done.

PR #663, opened 2020-01-17
Filter peers by community in RandomWalk strategy

MattSkala:
Currently, the `RandomWalk` discovery strategy tries to send an introduction request to a random verified peer in each step, not considering if the requested community is actually supported by that peer. This PR introduces filtering, so only peers that are members of the community are contacted.


qstokkink:
ok to test

PR #672, opened 2020-01-19
Added basic tests for the DataChecker class

devos50:
While I was debugging the libtorrent data checker, I wrote this basic test, so I thought I could as well add it to the IPv8 code base 👍 

Note that these tests only verify libtorrent DHT packets. We should probably also add some tests for UTP data later.

### DIFF
--- a/ipv8/test/messaging/anonymization/test_datachecker.py
+++ b/ipv8/test/messaging/anonymization/test_datachecker.py
@@ -1,0 +1,28 @@
+from binascii import unhexlify
+
+from ...base import TestBase
+from ....messaging.anonymization.tunnel import DataChecker
+
+
+class TestDataChecker(TestBase):
+
+    def setUp(self):
+        super(TestDataChecker, self).setUp()
+
+    def test_could_be_dht_correct(self):
+        """
+        Check if a valid DHT packet is correctly identified.
+        """
+        pkt = b"64313a6164323a696432303a3b2cb14348257b7a481f7010ca7c519b8bef5086393a696e666f5f6861736832303af84b51" \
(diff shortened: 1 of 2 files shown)


ipv8/test/messaging/anonymization/test_datachecker.py
```diff
@@ -0,0 +1,21 @@
...
+from ...base import TestBase
+from ....messaging.anonymization.tunnel import DataChecker
+
+
+class TestDataChecker(TestBase):
+
+    def setUp(self):
+        super(TestDataChecker, self).setUp()
+
+    def test_is_allowed(self):
```
qstokkink: Thanks for the test. To avoid this test becoming huge, I'd rather have it split into units though. Something like this:

```suggestion
    def test_could_be_dht_correct(self):
        """
        Check if a valid DHT packet is correctly identified.
        """
        pkt = b"64313a6164323a696432303a3b2cb14348257b7a481f7010ca7c519b8bef5086393a696e666f5f6861736832303af84b51" \
        b"f0d2c3455ab5dabb6643b4340234cd036e65313a71393a6765745f7065657273313a74323ae037313a76343a4c54010131" \
        b"3a79313a7165"
        self.assertTrue(DataChecker.is_allowed(unhexlify(pkt)))
        
        
    def test_could_be_dht_incorrect(self):
        """
        Check if an ivalid DHT packet is correctly identified.
        """
        pkt = b"63313a6164323a696432303a3b2cb14348257b7a481f7010ca7c519b8bef5086393a696e666f5f6861736832303af84b51" \
        b"f0d2c3455ab5dabb6643b4340234cd036e65313a71393a6765745f7065657273313a74323ae037313a76343a4c54010131" \
        b"3a79313a7165"
        self.assertFalse(DataChecker.is_allowed(unhexlify(pkt)))
```

devos50:
@qstokkink done 👍

PR #677, opened 2020-01-24
Fix block broadcast in TrustChain

MattSkala:
Currently, the block is being broadcasted to random peers in the network. This PR filters only peers participating in TrustChain community.

Additionally, service discovery was broken in tests due to a type error, which originally caused some tests to fail after this change.


xoriole:
ok to test

PR #678, opened 2020-01-24
Fix decrementing TTL when received block broadcast

MattSkala:
Currently, TTL is decremented by 2 whenever a block is received. This PR fixes it and adds a test that fails before this change.


xoriole:
ok to test

qstokkink:
Could you also fix this for `received_half_block_pair_broadcast`?

MattSkala:
Sure, I missed that one. I fixed it and updated the test.

PR #687, opened 2020-02-21
Fix TrustChain genesis block invariant validation

MattSkala:
The `is_genesis` condition in `update_block_invariant` prevented the sub-conditions to ever be evaluated true.

The tests were not failing only because the validation result was still invalid due to an invalid signature.


devos50:
ok to test

PR #689, opened 2020-02-24
Added ez_send shortcut

### LINKED ISSUE #688: Add ez_send to EZPackOverlay
Right now all overlay designers have to define the following to send a single payload to a peer (example taken from https://github.com/Tribler/tribler/blob/d66da4cd15474e9c0582e697694a95e03ecda9ac/src/tribler-core/tribler_core/modules/metadata_store/community/remote_query_community.py#L75-L77):

```python
self.endpoint.send(
    peer.address, self.ezr_pack(SelectResponsePayload.msg_id, SelectResponsePayload(request.id, data))
)
```

By only imposing a `msg_id` defined in the `Payload` class, we can create shortcuts for this type of usage like this:

```
self.ez_send(peer, SelectResponsePayload(request.id, data))
self.ez_senda(address, SelectResponsePayload(request.id, data))
```

The amount of payloads as arguments can be variable (but only the message id of the LAST message will be used).

qstokkink:
Fixes #688

### DIFF
--- a/ipv8/lazy_community.py
+++ b/ipv8/lazy_community.py
@@ -139,6 +139,37 @@ def lazy_wrapper_unsigned_wd(*payloads):
 
 class EZPackOverlay(Overlay):
 
+    def ez_send(self, peer, *payloads, **kwargs):
+        """
+        Send a Payload instance (with a defined `msg_id` field) to a peer.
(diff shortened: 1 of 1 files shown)


qstokkink:
@ichorid this one is for you.

PR #705, opened 2020-03-17
Used prefix mapping for endpoint

qstokkink:
This prefers to use prefix mapping instead of always looping through all listeners.

Also fixed a bug where the `AttestationCommunity` had a module-level lock instead of an instance-level lock.

### DIFF
--- a/ipv8/messaging/interfaces/endpoint.py
+++ b/ipv8/messaging/interfaces/endpoint.py
@@ -2,6 +2,7 @@ import abc
 import logging
 import socket
 import struct
+import threading
 
 try:
     # Especially on Android netifaces may fail.
@@ -16,9 +17,12 @@ class Endpoint(metaclass=abc.ABCMeta):
     Interface for sending messages over the Internet.
     """
 
-    def __init__(self):
+    def __init__(self, prefixlen=22):
         self._logger = logging.getLogger(self.__class__.__name__)
         self._listeners = []
+        self._prefix_map = {}
+        self.prefixlen = prefixlen
+        self.listener_update_lock = threading.RLock()
 
     def add_listener(self, listener):
         """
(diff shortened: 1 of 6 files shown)


ipv8/test/mocking/endpoint.py
```diff
@@ -34,8 +33,7 @@ def send(self, socket_address, packet):
         if socket_address in internet:
             # For the unit tests we handle messages in separate asyncio tasks to prevent infinite recursion.
             ep = internet[socket_address]
-            for listener in ep._listeners:
-                get_event_loop().call_soon(ep._deliver_later, listener, (self.wan_address, packet))
```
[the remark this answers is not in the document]
qstokkink: True, but only because it was removed from the `Endpoint`, which is now reinstated.

ipv8/messaging/interfaces/endpoint.py
```diff
@@ -28,19 +32,44 @@ def add_listener(self, listener):
         """
         if not isinstance(listener, EndpointListener):
             raise IllegalEndpointListenerError(listener)
-        self._listeners.append(listener)
+        with self.listener_update_lock:
+            self._listeners.append(listener)
+            for prefix in self._prefix_map:
+                self._prefix_map[prefix].append(listener)
```
egbertbouman: Why add the listener to both `self._listeners` and to `self._prefix_map`?

qstokkink: Speed upon receiving a packet.

It takes longer to add and remove, but when a packet is received it instantly has all the listeners it needs to deliver the packet to.

qstokkink: Oh, and this will use more memory.

qstokkink: Basically, we sacrifice memory and complexity for a marginal speed improvement.

PR #708, opened 2020-03-18
WIP: Used bytearray for packets internally

qstokkink:
This is *probably* faster.

This will need extensive profiling to see whether the overhead of converting to and from `bytearray` to `bytes` is worth it.


qstokkink:
This is pretty much equivalent in performance to the old implementation. Closing this PR.

PR #710, opened 2020-03-20
Fixed binary data in TC tx err 500

### LINKED ISSUE #709: 500 error when trying to get trustchain blocks
## General
When you try to get the most recent trustchain blocks via the REST API (http://localhost:14411/trustchain/recent) the server response with a 500 Internal Server Error and the error message:
```
File "/home/niels/Documents/ipv8/pyipv8/ipv8/messaging/deprecated/encoding.py", line 210, in _a_decode_unicode
    return offset + count, stream[offset:offset + count].decode("UTF-8")
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 0-1: invalid continuation byte
```
When trying to receive the blocks of a specific user (http://localhost:14411/trustchain/users/pub_key/blocks) the server also responds with a 500 Internal Server Error and the error message:
```
File "/home/niels/Documents/ipv8/pyipv8/ipv8/messaging/deprecated/encoding.py", line 210, in _a_decode_unicode
    return offset + count, stream[offset:offset + count].decode("UTF-8")
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 0-1: invalid continuation byte
```
As far as I understand it goes wrong when decoding the `transaction` part/key of a block on this line: https://github.com/Tribler/py-ipv8/blob/master/ipv8/attestation/trustchain/block.py#L513

## Steps to reproduce
1. Follow the "Enrollment/Attestation flow" section of the [Attestation Tutorial](https://py-ipv8.readthedocs.io/en/latest/basics/attestation_tutorial/#enrollment-attestation-flow)
2. Go to the above urls and confirm the response is a 500 Internal Server Error

qstokkink:
Fixes #709.

### DIFF
--- a/ipv8/messaging/deprecated/encoding.py
+++ b/ipv8/messaging/deprecated/encoding.py
@@ -212,6 +212,16 @@ def _a_decode_unicode(stream, offset, count, _):
         raise ValueError("Invalid stream length", len(stream), offset + count)
 
 
+def _a_decode_unicode_safe(stream, offset, count, _):
+    """
+    'a3sba\x80',3,3 --> 6,u'ba\x80'
+    """
+    if len(stream) >= offset + count:
+        return offset + count, "".join([chr(c) for c in stream[offset:offset + count]])
+    else:
+        raise ValueError("Invalid stream length", len(stream), offset + count)
+
+
 def _a_decode_bytes(stream, offset, count, _):
     """
     'a3bfoo',3,3 --> 6,'foo'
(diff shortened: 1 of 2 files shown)


n9iels:
The 500 error is solved. I get a JSON output with the following `transaction`:
```
"transaction": {
    "hash": "_^\u00ac\u001d;0\u00f5OP\u00cdo\u00fcVW\u00c6`\u00c8Q\u00f8\u000e",
    "name": "my_attribute",
    "date": 1584701156.413731,
    "metadata": {}
},
```
Is that correct?

qstokkink:
> Is that correct?

Correct, in a sense that it is the same as before (`ISO-8859-1` encoding): yes.

However, we should probably change the database format (specifically the `"hash"`) such that it is JSON compatible (`UTF-8` encoding, probably through further `base64` or `hex` encoding).

qstokkink:
Note that old blocks would still not be UTF-8, only newer blocks after I make that change.

PR #712, opened 2020-03-20
READY: Add REST for multiple Trustchains

qstokkink:
Right now, if multiple Trustchains were loaded, the REST API would simply pick one.
Which could occur if we (a) load the Tribler Trustchain together with the IPv8 identity or (b) allow for multiple pseudonyms in the IPv8 identity.

This PR unifies all loaded Trustchain instances under the REST API.

### DIFF
--- a/ipv8/REST/trustchain_endpoint.py
+++ b/ipv8/REST/trustchain_endpoint.py
@@ -1,4 +1,4 @@
-from binascii import unhexlify
+from binascii import hexlify, unhexlify
 
 from aiohttp import web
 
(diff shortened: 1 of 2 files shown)


ipv8/REST/trustchain_endpoint.py
```diff
@@ -65,8 +65,9 @@ def initialize(self, session):
             offset = int(request.query['offset'])
 
         return Response({
-            "blocks": [dict(block) for block in
-                       self.trustchain.persistence.get_recent_blocks(limit=limit, offset=offset)]
+            "blocks": [dict(block)
```
devos50: This would simply append all blocks from all TrustChains in a single list right? If the same peer is re-used across different TrustChain communities, it is impossible for an application using this REST API to distinguish blocks created in the different communities by the same peer.

Maybe we should add the TrustChain community as a REST resource to the URL? It's not very pretty and convenient but at least, it allows us to distinguish blocks in different communities.

qstokkink: I am torn on this decision.

Making it a REST resource would mean our out-of-the-box usage requires two calls instead of one (1st get the list of communities and find the right one and 2nd get the blocks). But it RESTful.

Alternatively, by making it an optional URL argument you would have out-of-the-box use for most use-cases that have a single Trustchain subclass loaded. But this is ugly and counter-REST.

qstokkink: [Moving this back to WIP as either of these should be implemented]

qstokkink: @egbertbouman had a suggestion (which I will follow): redirects. Meaning we will use the RESTful resource urls, but automatically redirect from `/trustchain/recent`.

qstokkink:
@devos50 implemented your suggestion

qstokkink:
In light of #733, Trustchain is probably disappearing from IPv8. I will close this PR.

PR #713, opened 2020-03-21
Fix RequestCache to use identifiers for tasks

ichorid:


ichorid:
Nah, it is not worth it to change this critical code with such petty detail

PR #715, opened 2020-03-23
Added managed futures to RequestCache

qstokkink:


ipv8/requestcache.py
```diff
@@ -20,6 +20,22 @@ def __init__(self, request_cache, prefix, number):
         self._prefix = prefix
         self._number = number
 
+        self._managed_futures = []
+
+    def register_future(self, future):
```
devos50: Can't we use the `TaskManager` class for this, which contains similar primitives to wrap futures and cancel them?

qstokkink: Yes we can and no we shouldn't. Then we would have to register TaskManagers (NumberCaches) in TaskManagers (RequestCache). You would be doing double work.

You could make an argument to register all the Futures in each NumberCache in the RequestCache separately, in case there is an intermittent crash between canceling Futures. I believe that's overkill though.

devos50: Oh, this change is is `NumberCache`, which is not a subclass of `TaskManager`. Then it's fine for me 👍

PR #720, opened 2020-03-24
Fix sqlite text type

devos50:
When running IPv8 with Python 2.7, it stored several text element as `TEXT` type. In Python 3, however, we are using `bytes` to fetch from the database, causing SQL queries to return incorrect results.

This PR bumps the version of the TrustChain database and converts all `TEXT` elements to the `BLOB` type. This should fix the SQL queries that are using `bytes` in their queries, and requires no changes to the existing SQL queries.


qstokkink:
retest this please

qstokkink:
Please merge (or not) after you finished your testing.

devos50:
The error is still present, although it is far less common now. I will not merge this one yet and investigate further.

PR #731, opened 2020-04-12
[WIP] The Noodle protocol

devos50:
Opening the PR already to have CI integration.

### DIFF
--- a/ipv8/attestation/noodle/block.py
+++ b/ipv8/attestation/noodle/block.py
@@ -1,0 +1,598 @@
+import logging
+import orjson as json
+import time
+from binascii import hexlify, unhexlify
+from collections import namedtuple
+from hashlib import sha256
+
+from .payload import HalfBlockPayload
+from ...database import database_blob
+from ...keyvault.crypto import default_eccrypto
+from ...messaging.deprecated.encoding import decode, encode
+from ...messaging.serialization import PackError, default_serializer
+
+
+GENESIS_HASH = b'0' * 32  # ID of the first block of the chain.
+GENESIS_SEQ = 1
+UNKNOWN_SEQ = 0
+EMPTY_SIG = b'0' * 64
+EMPTY_PK = b'0' * 74
+ANY_COUNTERPARTY_PK = EMPTY_PK
+SKIP_ATTRIBUTES = {'key', 'serializer', 'crypto', '_transaction', '_logger'}
+
+
+class NoodleBlock(object):
+    """
+    Container for Noodle block information
+    """
+    Data = namedtuple('Data', ['type',
+                               'transaction',
+                               'public_key',
+                               'sequence_number',
+                               'link_public_key',
+                               'link_sequence_number',
+                               'previous_hash',
+                               'signature',
+                               'timestamp',
+                               'insert_time'])
+    """
+    Data struct to initialize a NoodleBlock.
+
+
+        **[0] type:** The block type name, as utf-8 string or utf-8 encoded bytes.
+
+        *> type:* str or bytes
+
+        **[1] transaction:** Metadata dictionary, consisting of binary type strings.
+
+        *> type:* dict
+
+        **[2] public_key:** The serialized public key of the initiator of this block (binary data).
+
+        *> type:* bytes (Py3) or str (Py2)
+
+        **[3] sequence_number:** The the sequence number of this block in the chain of the intiator of this block.
+
+        *> type:* int
+
+        **[4] link_public_key:** The serialized public key of the counterparty of this block (binary data).
+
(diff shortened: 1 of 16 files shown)


qstokkink:
@grimadas @devos50 

Even though this is not done yet and I see lesser problems such as code duplication, some general remarks:

 - [**Mortal sin** 💀] Please do not make feature branches on the main repository, use a fork for this. You have permanently tainted the Git history, inflating it for all eternity for everyone who pulls the IPv8 repository. This leads to inflation of the git files, causing stupid stuff like the Tribler repo which has a bigger history than codebase and no way to get rid of it. Also, frankly, it's just very annoying to be pulling in a feature branch every time you clone or pull the IPv8 repository.

- [**Red flag** 🚩] Adding dependencies. Especially `networkx` is a very nasty dependency, which I have finally worked out after a year and it is being casually reintroduced. This makes the code slower and any compiled binaries bigger and harder to work with. Just like `orjson`, you really don't need these external packages, they're just nice to have. Both of these dependencies have no business being in the main repository. If you want to use them for your own projects, or in the Tribler repository, be my guest though -- introduce them there.

- [**Red flag** 🚩] Changing the `Network` class. Generally if you need to change it, you're doing it wrong. This is the most fragile and most critical component of IPv8 and changes to it should not be taken lightly. It is a performance killer and the main source of memory leaks.

Concluding:
- The mortal sin should never be repeated.
- The red flags we can talk about, but changes in those departments warrant separate pull requests and very well motivated reasons for introducing these changes.

devos50:
@qstokkink thanks for your feedback!

I agree with the red flags, however, this is the code as it is shipped in the experimental release. I am aware that we are adding dependencies, which we should avoid as much as possible of course. I see opportunities to get rid of these new dependencies.

This code will most likely undergo several iterations of improvements/refactoring efforts before it is ready for review/merge.

Regarding the mortal sin, I disagree. I made the Noodle protocol an explicit branch of the IPv8 main repository for the following two reasons:
- It is a future and a fundamental mechanism to enhance one of the core features of IPv8, TrustChain. Since more people work on this feature, the code attached to this feature should be in a shared git repository and not buried on one of our forks.
- Extending upon my first point, there was confusion during the development process of the Noodle community between @grimadas and me. Specifically, it was unclear what the 'current' state of the code was, which moved between repositories multiple times. By hosting the code here, I aim to address this issue since it is now clear for everyone where the latest version of the code can be found.

For the above reasons, I do not consider this practice as a mortal sin (unless well-established software practices prove me wrong).

qstokkink:
I will not repeat my original point, but let me address yours:

> Since more people work on this feature, the code attached to this feature should be in a shared git repository and not buried on one of our forks.

There is nothing preventing you from working together on a fork. In fact, it is what git was made for.

> By hosting the code here, I aim to address this issue since it is now clear for everyone where the latest version of the code can be found.

Being unable to coordinate a team is not an excuse for doing that which cannot be undone (see my original point). 

> unless well-established software practices prove me wrong

Read this (collection of excerpts from literature):
https://www.sciencedirect.com/topics/computer-science/development-branch

Important highlights:
 - Keep it on a fork.
 - Submit a clean version for PR.

devos50:
> There is nothing preventing you from working together on a fork. In fact, it is what git was made for.

There's also nothing preventing us from making a feature branch on the main repository.

Except from the [Gitflow workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow), which is a standardized software pattern:

> In addition, feature branches can (and should) be pushed to the central repository. This makes it possible to share a feature with other developers without touching any official code. Since master is the only “special” branch, storing several feature branches on the central repository doesn’t pose any problems. Of course, this is also a convenient way to back up everybody’s local commits. The following is a walk-through of the life-cycle of a feature branch.

> Being unable to coordinate a team is not an excuse for doing that which cannot be undone (see my original point).

I will migrate the Noodle development to my fork, but you should then address the other, existing branches, which I think have even less right to exist on the IPv8 main repository.

devos50:
Also, to prevent mortal sins in the future, please write some contribution guidelines and publish them.

qstokkink:
> storing several feature branches on the central repository

To be clear: I have no problem with storing clean features, only storing development branches. I also have no issue with opening a PR to a development branch on a fork. The `crawler`, `experimental_gossip_primitive`, `nuclear_backdoor` and `python2-1.10` all contain features which are incompatible with the `master` branch, but finished. I asked @grimadas to remove the `db-experimental` branch already.

> Also, to prevent mortal sins in the future, please write some contribution guidelines and publish them.

Normal contributors do not have admin rights. But I'll make a contributing guide nonetheless 👍 .

Issue #733, opened 2020-04-20
ID 2.0


qstokkink:
The identity layer has remained virtually unchanged for more than 2 years in IPv8, but is in need of some maintenance. Some initial design decisions require work to improve security and functionality, though they break the disk format and protocol:

 - [x] Add REST API access key (@egbertbouman made this for Tribler already - and will provide this PR). #773
 - [x] Refactor REST API design (URL parameters are not pretty 😃 ). #742 #772
 - [x] Add DAG Trustchain derivative for selective disclosure of metadata. #754 #758
 - [x] Add pseudonym chains --completely devoid of all metadata-- for storage of attribute commitments for ZKPs. #738
 - [x] Add pseudonym REST API bindings for easy pseudonym control. #759
 - [x] Add additional advanced controls for attestation REST API. #772
 - [x] Update documentation. #779

Once this is all complete, Trustchain will probably be moved into the `Tribler` repository. The identity solution may be moved into its own new repository.

synctext:
Forking software is something that is quite a red flag for me. 

Something that is virtually unchanged for more than 2 years is clearly stable and low maintenance, on the other hand. Can you do an ID2.0 upgrade both the Tribler usage and identity usage?
Your list seems like solid positive maintenance upgrade.

> The identity solution may be moved into its own new repository.

Lets discuss matters first, before any changes are made. I have a preference for mono-repo development. Why not simply another branch in the repo?

qstokkink:
> Something that is virtually unchanged for more than 2 years is clearly stable and low maintenance, on the other hand. Can you do an ID2.0 upgrade both the Tribler usage and identity usage?

I agree and I would not have posted this issue if it was not a pressing concern. Stable as it may be, our identity layer has security concerns regarding disclosure of metadata. This can still be somewhat controlled by patching the current Trustchain functionality for production use. However, with Trustchain also being **forked already** for currency use, with **more aggressive spreading** of blocks in the fundamental protocol (where blocks contain metadata in the case of identity), now is the time to fix all of this. We have reached the breaking point for both use cases on the same chain type.

> Why not simply another branch in the repo?

I'd like to avoid shipping dead code in a repository. IPv8 is being shipped for both the Tribler and identity use cases. Tribler does not use the identity part of IPv8 at all, inflating its binary size and having its interpreter load code that will never be used. That said, this is not a pressing issue, just some code organization: the identity refactoring has the utmost priority now.

synctext:
[...]

PR #734, opened 2020-04-20
Fixed attestation tutorial documentation

qstokkink:
The 10 second timeout was fixed months ago.


devos50:
Thanks for the fix! (one of our software engineering projects is going to follow the tutorial this week).

qstokkink:
I noticed: I might also make another PyPi release to this end soon - so they have a stable version to work with before I starting moving around all of the identity stuff for #733.

PR #738, opened 2020-04-27
Added token chaining datastructure

qstokkink:
This is the first part of #733, adding chains of tokens devoid of metadata.

All code is typed and documented.

To guide reviews, this is the visualization of the datastructure:

tokenchain_arch

Some of the guarantees and design principles:

 - A "view" of a (partial) TokenChain can be (re)constructed and verified without knowledge of the public key or the contents.
 - The signatures and the genesis block pointer can be verified with a public key (not part of the public serialized "view").
 - Given a token and content, it can be shown whether the content belongs to the token. Given the token, it cannot be derived what content belongs to it (hash).
 - The serialization can take any key format with a known signature length. The serialization is bound to 32-byte/256-bit hashes.

In future work, the content will be filled by cryptographic commitments and metadata will point to the tokens.

### DIFF
--- a/ipv8/attestation/tokenchain/token.py
+++ b/ipv8/attestation/tokenchain/token.py
@@ -1,0 +1,155 @@
+from binascii import hexlify
+from hashlib import sha3_256
+from struct import unpack
+from typing import Optional, Tuple, TypeVar
+
+from ...keyvault.crypto import ECCrypto
+from ...keyvault.keys import PrivateKey, PublicKey
+
+
+TokenType = TypeVar('TokenType', bound='Token')
+
+
+class Token(object):
+    """
+    A double pointer, pointing to a previous token and content.
+    The contents belonging to the content pointer can be added to a Token object.
+
+    Tokens do not and should not contain an index (but occupy one in a chain).
+    Tokens do not and should not contain a reference to the public key (but are signed by one).
+    """
+
+    def __init__(self,
+                 previous_token_hash: bytes,
+                 content: Optional[bytes] = None,
+                 content_hash: Optional[bytes] = None,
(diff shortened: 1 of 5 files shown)


ipv8/attestation/tokenchain/token.py
```diff
@@ -0,0 +1,155 @@
...
+                 signature: Optional[bytes] = None) -> None:
+        """
+        Create a new Token object, always specify the hash of the preceding Token.
+
+        For a Token belonging to yourself, give the content and private key and the content hash and signature will
+        be automatically created.
+        For a Token of someone else or your own reloaded Token, give the content hash and the signature.
+
+        :param previous_token_hash: a string pointing to the hash of the preceding Token.
+        :param content: to content to tokenize.
```
devos50, line 36: *the

qstokkink:
> 
> 
> I've thoroughly reviewed this PR. Although I am not entirely sure how this exactly going to be integrated with the current identity solution, it is worth pointing out that this code is "dead" currently. I would prefer to only merge code that is called in some way or another if it would be the Tribler repository. However, I believe that your next PR will most likely address this concern 👍
> 
> Anyway, I could only find one typo.

Thanks for the review, I fixed the typo.

Also: you are correct, I am slowly and incrementally adding the individual components to separate the Trustchain-As-An-Identity solution from the Trustchain-As-A-Currency implementation. There wil probably be another +- 4 PRs before this is done. After that, we can make a clean separation of the Trustchain implementations.

qstokkink:
@devos50 could you review my typo fix and approve 😃 ?

qstokkink:
Thanks 👍

PR #740, opened 2020-04-30
Make the host binding for the REST api configurable

n9iels:
## General
Currently the `port` for the REST API is configurable but the host is not. This causes issues when you like to bind the REST API to a specific interface or IP. For example, when using IPv8 inside a Docker container the host must be set to `0.0.0.0` in order to make communication possible. When bind to `127.0.0.1` (the current default) communication with the outside world is not possible.

## Changes
This PR adds a seconds parameter `host` to the `start()` method of `rest_manager.py`. The default value is still `127.0.0.1`.

### DIFF
--- a/ipv8/REST/rest_manager.py
+++ b/ipv8/REST/rest_manager.py
@@ -32,7 +32,7 @@ class RESTManager:
         self.session = session
         self.site = None
 
-    async def start(self, port=8085):
+    async def start(self, port=8085, host='127.0.0.1'):
         """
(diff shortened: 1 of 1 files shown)


egbertbouman:
ok to test

egbertbouman:
Seems fine. Please do remember that the REST API server uses plain HTTP and therefore is generally a bad idea to bind to 0.0.0.0. Although your case is different.

Maybe log a warning if the host is not local?

PR #745, opened 2020-05-17
Various testing infrastructure improvements

### LINKED ISSUE #744: create_test_coverage_report still using Twisted
The `create_test_coverage_report.py` script still uses Twisted. This should be switched to `asyncio`.

qstokkink:
- Removes unused endpoint tests.
- Fixes #744.
- Added test requirements to `setup.py`.
- Added universal Python-based test invocation script (to remove operating specific `run_all_tests_*` scripts).

The new universal test script is also multiprocessed, which brings the test time down from 54 seconds to 18 seconds.

I will follow up with another PR for the `run_all_tests.py` script, integrating it properly into the documentation. First I want to do some preliminary field testing (if it turns out it doesn't work *everywhere*, I'll remove it again).

### DIFF
--- a/create_test_coverage_report.py
+++ b/create_test_coverage_report.py
@@ -3,11 +3,10 @@ import os
 import shutil
 import sys
 from io import StringIO
+from unittest import TextTestRunner, defaultTestLoader
+from unittest.suite import TestSuite
 
 import coverage
-
-from twisted.trial.reporter import VerboseTextReporter
-from twisted.trial.runner import TestLoader
 
 if __name__ != '__main__':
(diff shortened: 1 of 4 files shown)


devos50:
The mac machine seems to be offline. I will look at it tomorrow when I'm at the office.

qstokkink:
@devos50 thanks 👍

I think I'll mark this as ready for review (when the Mac test passes). I'll slowly experiment with integrating the new test runner script and if everything works I'll do the final change of integrating it into the documentation.

qstokkink:
Ready for review, here's an animated gif to celebrate:

output

PR #748, opened 2020-05-30
Avoid loosing peer_flags on disconnect

egbertbouman:


egbertbouman:
@qstokkink It's not as bad as it sounds. However, with the new HTTP logic in Tribler, losing the peer_flags of a circuit means we may end up creating more circuits then needed. Luckily, the HTTP stuff isn't released yet.

Issue #752, opened 2020-06-05
Defer I/O to a separate process


qstokkink:
IPv8 has as many threads as it has branches: only one.

This single thread approach also means that any nasty operations (like file I/O) on the main thread block all other events (most notably packets will be dropped by the socket).

Today I decided to investigate the impact of deferring file I/O to a separate process on the *responsiveness* of periodic calls. The experiment setup consists of two asynchronous calls:
 - A measuring call attempting to sleep for 0.01 seconds continuously.
 - A call making 3 batches of file writes. Each batch contains 500 files and uses a different writing strategy.

The strategies are as follows:
 1. Await an `asyncio.gather` of all 500 file writes performed by a different process.
```python
await asyncio.gather(write1(), write2(), write3(), ... )
```
 2. Sequentially await all 500 files writes performed by a different process.
```python
await write1()
await write2()
await write3()
...
```
 3. Run all 500 file writes in the call itself.
```python
with open(file1, "wb") as f:
    f.write("something")
with open(file2, "wb") as f:
    f.write("something")
...
```

The measuring call measures the time it takes beyond the 0.01 seconds for control to return to it, which I called *drift*. A drift of 0.01 seconds would mean that the requested 0.01 second sleep took 0.02 seconds. I plotted the drift as the writing call went through the three strategies - white shading signifies the waiting time between starting and completing the 500 file writes:

![clock_drift](https://user-images.githubusercontent.com/3630389/83891646-cbb24600-a74d-11ea-9ffb-7f66eaa223ec.png)

### Findings

 - Never ever use `gather` to batch file I/O offloading.
 - If you want speed and don't care about other event handling, run everything on the main thread using the normal file I/O.
 - If you want to remain responsive to other events, sequentially offload (i.e. `await` each write on demand) your file I/O to another process. This is slower however.

### Conclusion
Offloading I/O to a separate process has its place and time. As IPv8 is event-based, it would pay off to offload I/O to another process under heavy load to remain responsive to events. Future work consists of checking if this holds for all I/O - most importantly the database I/O.

egbertbouman:
From what I understand in strategy 1 you create 500 tasks more or less at the same time, so more drift seems to be expected. In strategy 2 you create 1 task at a time, so obviously there is less drift. Besides strategy 3, where you block the `asyncio` thread, perhaps you could also try making the write calls non-async and using an executor:
```python
await loop.run_in_executor(None, write1)
````

In combination with `gather` (untested):
```python
[...]

PR #753, opened 2020-06-06
Added REST endpoint for measuring core drift

qstokkink:
Related to #752

This adds an Endpoint that stores the last 100 core tick drift measurements.
It can be enabled and disabled with a put request (it starts off disabled).

### DIFF
--- a/ipv8/REST/health_endpoint.py
+++ b/ipv8/REST/health_endpoint.py
@@ -1,0 +1,114 @@
+import collections
+import time
+
+from aiohttp import web
+
+from aiohttp_apispec import docs, json_schema
+
+from marshmallow.fields import Boolean, Float
+
+from .base_endpoint import BaseEndpoint, HTTP_BAD_REQUEST, HTTP_NOT_FOUND, Response
+from .schema import DefaultResponseSchema, schema
+from ..peerdiscovery.discovery import DiscoveryStrategy
+
+
+class DriftMeasurementStrategy(DiscoveryStrategy):
+
+    def __init__(self, core_update_rate):
+        super(DriftMeasurementStrategy, self).__init__(None)
+        self.last_measurement = time.time()
+        self.history = collections.deque(maxlen=100)
+        self.core_update_rate = core_update_rate
+        self.overlay = type('FakeOverlay', (object,), {'get_peers': lambda: []})
+
+    def take_step(self):
+        with self.walk_lock:
+            this_time = time.time()
+            self.history.append((this_time, max(0.0, this_time - self.last_measurement - self.core_update_rate)))
+            self.last_measurement = this_time
+
+
+class HealthEndpoint(BaseEndpoint):
+    """
+    This endpoint manages measurements of non-functional requirements.
+    """
+
+    def __init__(self):
+        super(HealthEndpoint, self).__init__()
+        self.strategy = None
+        self.enabled = False
+
+    def setup_routes(self):
+        self.app.add_routes([web.get('/drift', self.retrieve_drift), web.put('/enable', self.enable_measurements)])
+
+    def enable(self):
+        if not self.session:
+            return False
+        if not self.enabled:
+            self.strategy = DriftMeasurementStrategy(self.session.walk_interval)
+            with self.session.overlay_lock:
+                self.session.strategies.append((self.strategy, -1))
+            return True
+        return True
+
+    def disable(self):
+        if not self.session:
+            return False
+        if self.enabled:
+            with self.session.overlay_lock:
+                self.session.strategies = [s for s in self.session.strategies
+                                           if not isinstance(s[0], DriftMeasurementStrategy)]
+            return True
+        return True
+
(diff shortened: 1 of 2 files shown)


qstokkink:
@egbertbouman could you have a look at this, it's my first `aiohttp_apispec` endpoint.

ipv8/REST/health_endpoint.py
```diff
@@ -0,0 +1,112 @@
...
+                "schema": schema(DriftResponse={"measurements": [
+                    schema(Measurement={
+                        "timestamp": Float,
+                        "drift": Float
+                    })
+                ]})
+            },
+            400: {
+                "schema": DefaultResponseSchema,
+                "examples": {'': {"success": False, "error": "Core drift disabled."}}
```
egbertbouman, line 78: If you prefer not to give a name to the example, you can rewrite this as
```python
"example": {"success": False, "error": "Core drift disabled."}
```

ipv8/REST/health_endpoint.py
```diff
@@ -0,0 +1,112 @@
...
+            'in': 'body',
+            'name': 'enable',
+            'description': 'Whether to enable or disable measuring.',
+            'type': 'boolean',
+            'required': True
+        }],
+        responses={
+            200: {
+                "schema": DefaultResponseSchema,
+                "examples": {'true': {"success": True, "error": ""}}
```
egbertbouman, line 101: Since 'true' is not very descriptive anyway, this can be rewritten as:
```python
"example": {"success": True, "error": ""}
```

ipv8/REST/health_endpoint.py
```diff
@@ -0,0 +1,112 @@
...
+            400: {
+                "schema": DefaultResponseSchema,
+                "examples": {'Incorrect parameters': {"success": False, "error": "incorrect parameters"}}
+            }
+        }
+    )
+    async def enable_measurements(self, request):
+        enabled = await request.json()
+        status = self.enable() if enabled else self.disable()
+        return Response({"success": status, "error": "" if status else "Session not initialized."})
```
egbertbouman, line 112: Maybe drop the `error` field if there is no error? The `DefaultResponseSchema` already allows for this.

ipv8/REST/health_endpoint.py
```diff
@@ -0,0 +1,112 @@
...
+    async def retrieve_drift(self, _):
+        if self.strategy is None:
+            return Response({"success": False, "error": "Core drift disabled."}, status=HTTP_NOT_FOUND)
+        with self.strategy.walk_lock:
+            return Response({"measurements": [{"timestamp": e[0], "drift": e[1]} for e in self.strategy.history]})
+
+    @docs(
+        tags=["Health"],
+        summary="Enable or disable measurements.",
+        parameters=[{
```
egbertbouman, line 91: While a plain string is considered valid JSON (although in the past this was different), in our other endpoints we tend to put the request parameters in a dict. Maybe not pretty, but it is consistent:
```python
    @json_schema(schema(EnableHealthRequest={
        'enable*': (Boolean, 'Whether to enable or disable measuring.'),
    }))
```

qstokkink:
@egbertbouman thanks! I implemented your suggestions (and verified everything still works). Could you verify I didn't do anything stupid and formally submit your review?

After this is merged we can have some fun with the Tribler debug panel :)

ipv8/REST/health_endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+        with self.strategy.walk_lock:
+            return Response({"measurements": [{"timestamp": e[0], "drift": e[1]} for e in self.strategy.history]})
+
+    @docs(
+        tags=["Health"],
+        summary="Enable or disable measurements.",
+        responses={
+            200: {
+                "schema": DefaultResponseSchema,
+                "examples": {"success": True}
```
egbertbouman, line 94: You need to change `examples` to `example`, otherwise 'success' will be considered the name of the example (instead of a JSON field).

ipv8/REST/health_endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    )
+    @json_schema(schema(EnableHealthRequest={
+        'enable*': (Boolean, 'Whether to enable or disable measuring.'),
+    }))
+    async def enable_measurements(self, request):
+        enabled = await request.json()
```
egbertbouman, line 102: Since 'enabled' is now a field in a JSON dict, you'll to something like:
```python
parameters = await request.json()
enabled = parameters.get('enabled')
```
Also, considering you've marked the field as a required field in the docs, you'll have to verify that it exists (or don't make the field required).

egbertbouman: This reminds me that I still have to look into the `validation_middleware` so that `marshmallow` schemas will be checked automatically.

qstokkink:
@egbertbouman incorporated (and working)! Would you mind giving her another review?

PR #759, opened 2020-06-15
Integrated ID 2.0 pseudonym management

qstokkink:
This PR performs integration of the new ID Trustchain derivative with the identity overlay for #733.

In the process of testing I also automated the tutorial (which I will probably make a PR for in the future):
https://gist.github.com/qstokkink/a76a91ec00f8e5c496e0d21c91af6217

Concretely this PR;

 - Updates the attestation tutorial documentation.
 - Ports the old REST API to use the new backend.
 - Ports the old community to no longer subclass the currency Trustchain.
 - Ports the old identity tests to the new format.
 - Adds a convenience method to the ID 2.0 token datastructure for getting a single chain.
 - Internally uses unicode strings instead of flip-flopping to and from bytes for metadata.
 - Fixes race conditions in the REST tests.

*Roadmap*
This is the most intensive operation of #733, but ends the big backend work. After this the REST API bindings should be modernized and #733 is done.

### DIFF
--- a/doc/basics/attestation_tutorial.rst
+++ b/doc/basics/attestation_tutorial.rst
@@ -64,7 +64,6 @@ Fill your ``main.py`` file with the following code (runnable with ``python3 main
            configuration['logger']['level'] = "ERROR"
            configuration['keys'] = [
                {'alias': "anonymous id", 'generation': u"curve25519", 'file': u"ec%d_multichain.pem" % i},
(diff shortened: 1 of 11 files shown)


qstokkink:
retest this please

qstokkink:
Temporarily back to WIP to rebase onto the latest master.

PR #768, opened 2020-06-20
Added Asyncio debug endpoints

egbertbouman:
This PR:
* Renames HealthEndpoint to AsyncioEndpoint
* Adds task and debug endpoints
* Fixes an issue with drift measurements not disabling when asked

### DIFF
--- a/ipv8/REST/asyncio_endpoint.py
+++ b/ipv8/REST/asyncio_endpoint.py
@@ -1,15 +1,27 @@
 import collections
+import logging
 import time
+from asyncio import all_tasks, current_task, get_event_loop
 
 from aiohttp import web
 
 from aiohttp_apispec import docs, json_schema
 
-from marshmallow.fields import Boolean, Float
+from marshmallow.fields import Boolean, Float, Integer, String
 
 from .base_endpoint import BaseEndpoint, HTTP_BAD_REQUEST, HTTP_NOT_FOUND, Response
 from .schema import DefaultResponseSchema, schema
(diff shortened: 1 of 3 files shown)


ipv8/REST/asyncio_endpoint.py
```diff
@@ -0,0 +1,240 @@
...
+                ]})
+            },
+            400: {
+                "schema": DefaultResponseSchema,
+                "example": {"success": False, "error": "at least Python 3.7 is required"}
+            }
+        }
+    )
+    async def get_asyncio_tasks(self, _):
+        if sys.version_info[1] < 7:
```
qstokkink, line 157: This is a project requirement already, we don't support anything below Python 3.7. Misconfiguring the OS is HTTP 500 worthy i.m.o. (and saves adding this if-statement).

egbertbouman: It's there because some of our users are running Tribler under Python 3.6. Back when we released v7.5.0-exp2, we had several reports related to Python 3.6 compatibility.

However, if you're announcing that 3.6 is not supported, I'll stop trying to keep things compatible.

egbertbouman: Nevermind, I'll just remove it 🤷

qstokkink: This is actually in the `setup.py`: https://github.com/Tribler/py-ipv8/blob/master/setup.py#L36

(which we should update to also include Python 3.8 now that I look at it)

Issue #770, opened 2020-06-23
NoneType' object has no attribute 'number'


devos50:
As reported by a user running Tribler 7.5.0 (trace 33791):

```
Traceback (most recent call last):
File "src\tribler-gui\tribler_gui\event_request_manager.py", line 112, in on_read_data
RuntimeError: 'NoneType' object has no attribute 'number'
--LONG TEXT--
Traceback (most recent call last):
File "ipv8\dht\community.py", line 388, in _contact_node
File "ipv8\dht\community.py", line 383, in _send_find_request
AttributeError: 'NoneType' object has no attribute 'number'
--CONTEXT--
{'message': 'Task exception was never retrieved', 'exception': AttributeError("'NoneType' object has no attribute 'number'"), 'future': <Task finished coro=<DHTCommunity._contact_node() done, defined at ipv8\dht\community.py:386> exception=AttributeError("'NoneType' object has no attribute 'number'")>}
```

qstokkink:
Seems like a DHT error, @egbertbouman do you want to take this one?

devos50:
I saw this one popping up very briefly during Tribler shutdown.

PR #771, opened 2020-06-23
Fix deadlock while shutting down RequestCache

egbertbouman:
This PR fixes a deadlock that occurs if:
* The request cache is in the middle of shutting down, but still holds the lock.
* At the same time a task wants to add something to the `RequestCache`, so it tries to get the lock.

I've noticed this in Tribler while unloading communities, so this should fix that issue as well.
Personally, I can't think of any situation in which the locks would be required, so we may have to think about removing them altogether.

This PR also:
* Fixes https://github.com/Tribler/py-ipv8/issues/770
* Fixes a memory leak caused by the exit socket `TaskManager` not being shutdown.
* Includes a bunch of smaller fixes.

### DIFF
--- a/ipv8/test/dht/test_community.py
+++ b/ipv8/test/dht/test_community.py
@@ -1,5 +1,5 @@
 import time
-from asyncio import TimeoutError, wait_for
+from asyncio import TimeoutError, ensure_future, wait_for
 
 from ..base import TestBase
 from ..mocking.ipv8 import MockIPv8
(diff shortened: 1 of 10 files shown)


devos50:
retest this please

ipv8/test/dht/test_community.py
```diff
@@ -201,6 +201,12 @@ async def test_rate_limit(self):
         with self.assertRaises(TimeoutError):
             await wait_for(self.nodes[0].overlay.ping(node1), 0.1)
 
+    async def test_unload_while_contacting_node(self):
```
qstokkink, line 204: [I know this PR is still WIP, but:] please assert *something* in this unit test.

egbertbouman: Sure, since you asked *so* nicely.

PR #772, opened 2020-06-24
Added v2 ID 2.0 REST API

### LINKED ISSUE #742: Verification output endpoint show verifications that are in progress
The endpoint `attestation?type=verification_output` shows verifications that are still in progress. The issue is that those verifications always have a `match` value of `0`. So when the endpoint is called during the verification process the result might indicate that the verification faild, but it is still in progress and a few ms later the `match` value could be different.

I noticed this behavior when writing integration test. The time between `verify`, `allow_verify` and `verification_output` is extremely short since they are right after each other. In more "normal" use cases I never encountered this issue.

### LINKED ISSUE #774: RuntimeWarning: coroutine 'Overlay.unload' was never awaited
Found this one in the Mac tests:

```
test_create_circuit_multiple_calls (ipv8.test.messaging.anonymization.test_community.TestTunnelCommunity) ... /Users/tribler/Documents/workspace2/workspace/ipv8/test_ipv8_mac/ipv8/REST/attestation_endpoint.py:248: RuntimeWarning: coroutine 'Overlay.unload' was never awaited
  self.session.unload_overlay(self.identity_overlay)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

qstokkink:
This PR:

 - Adds middleware (CommunicationManager/CommunicationChannel) for dealing with ID 2.0 pseudonyms.
 - Adds a new `identity` REST endpoint to replace `attestation` (and a lot of tests 😃).
 - Adds the factory method `produce_anonymized_endpoint` to `IPv8` to facilitate clean testing.
 - Fixes the name of the `temporary_directory` of `TestBase`.
 - Fixes the infinite loop on the main thread in the (old) attestation REST API.
 - Adds the ability to specify a JSON body to the REST test base in `make_request`.
 - Saves the launch configuration of IPv8 into `IPv8.configuration`.
 - Fixes #742
 - Fixes #774

Remaining for the roadmap of #733 after this PR are: updating the documentation and adding an access token for the API.

### DIFF
--- a/ipv8/attestation/communication_manager.py
+++ b/ipv8/attestation/communication_manager.py
@@ -1,0 +1,262 @@
+import asyncio
+import json
+import os
+import typing
+
+from .identity.community import IdentityCommunity, create_community
+from .identity.manager import IdentityManager
+from .wallet.community import AttestationCommunity
+from ..keyvault.crypto import ECCrypto
+from ..util import succeed
+
+
+class CommunicationChannel(object):
+
+    def __init__(self, attestation_community, identity_community):
+        super(CommunicationChannel, self).__init__()
+        self.attestation_overlay = attestation_community
+        self.identity_overlay = identity_community
+
+        self.attestation_requests = {}
+        self.verify_requests = {}
+        self.verification_output = {}
+        self.attestation_metadata = {}
+
+        self.attestation_overlay.set_attestation_request_callback(self.on_request_attestation)
+        self.attestation_overlay.set_attestation_request_complete_callback(self.on_attestation_complete)
+        self.attestation_overlay.set_verify_request_callback(self.on_verify_request)
+
+    @property
+    def public_key_bin(self):
+        return self.identity_overlay.my_peer.public_key.key_to_bin()
+
+    @property
+    def peers(self):
+        return self.identity_overlay.get_peers()
+
+    @property
+    def schemas(self):
+        return list(self.attestation_overlay.schema_manager.formats.keys())
+
+    def on_request_attestation(self, peer, attribute_name, metadata):
+        """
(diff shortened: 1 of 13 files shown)


devos50:
retest this please

qstokkink:
retest this please

qstokkink:
🤔 the PR testers are not liking _something_ in this PR.. they are almost immediately dying in the middle of a test run:

```
+ python3 --version
Python 3.7.3
+ export PYTHONPATH=.
+ nosetests3 ./ipv8/test

..............Build step 'Execute shell' marked build as failure
```

This might be a 3.7 vs 3.8 thing?

ipv8/attestation/communication_manager.py
```diff
@@ -0,0 +1,258 @@
...
+    def attest(self, peer, attribute_name, value):
+        outstanding = self.attestation_requests.pop((peer, attribute_name))
+        outstanding[0].set_result(value)
+
+    def import_blob(self, attribute_name, id_format, metadata, value):
+        metadata.update({"id_format": id_format})
+        self.attestation_overlay.dump_blob(attribute_name, id_format, value, metadata)
+
+    def allow_verification(self, peer, attribute_name):
+        outstanding = self.verify_requests.pop((peer, attribute_name))
```
egbertbouman, line 164: What if verification is not allowed, is there some kind of timeout for this? Otherwise, the message handler that awaits this `Future` could end up waiting indefinitely.

qstokkink: A user should be able to answer at any point in time he likes, so there is no timeout. However, having a "disallow" may indeed be a good idea.

qstokkink:
@egbertbouman done!

PR #776, opened 2020-06-25
Small fixes

### LINKED ISSUE #769: Add Python 3.8 to supported versions
IPv8 has no conflicts with Python 3.8: we should add it to the supported versions in `setup.py`.

### LINKED ISSUE #775: run_all_tests.py mixes print statements from different processes
As `run_all_tests.py` uses multiprocessing, combining the logging output from different processes may result in writes on the same line in the global log, causing ugly artifacts like this:

```
test_find_nodes (ipv8.test.dht.test_community.TestDHTCommunity) ... ok [7.8 ms]
test_dht_lookup_no_counterparty (ipv8.test.messaging.anonymization.test_hiddenservices.TestHiddenServices) ... test_get_set_del (ipv8.test.dht.test_trie.TestNode) ... ok [1.19 ms]
test_longest_prefix (ipv8.test.dht.test_trie.TestNode) ... ok [0.88 ms]
```

This can be resolved by not allowing half-lines in the local logs of a `CustomLinePrint`, only full lines should be inserted:

```
    def write(self, __text: str) -> int:
        wtime = time.time()
        if self.raw_lines and not self.raw_lines[-1][2].endswith('\n'):
            self.raw_lines[-1] = (self.raw_lines[-1][0], self.raw_lines[-1][1], self.raw_lines[-1][2] + __text)
        else:
            self.raw_lines.append((wtime, self.prefix, __text))
        return self.delegated.write(__text)
```

qstokkink:
Fixes #775, fixes #769.

### DIFF
--- a/run_all_tests.py
+++ b/run_all_tests.py
@@ -61,7 +61,10 @@ class CustomLinePrint(io.StringIO):
 
     def write(self, __text: str) -> int:
         wtime = time.time()
-        self.raw_lines.append((wtime, self.prefix, __text))
+        if self.raw_lines and not self.raw_lines[-1][2].endswith('\n'):
+            self.raw_lines[-1] = (self.raw_lines[-1][0], self.raw_lines[-1][1], self.raw_lines[-1][2] + __text)
+        else:
+            self.raw_lines.append((wtime, self.prefix, __text))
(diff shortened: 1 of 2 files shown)


qstokkink:
retest this please

PR #790, opened 2020-06-30
Added convenience add strategy method

qstokkink:
This PR adds a convenience method to add strategies/overlays to IPv8. This makes sure programmers don't have to worry about acquiring a lock.

### DIFF
--- a/ipv8_service.py
+++ b/ipv8_service.py
@@ -167,6 +167,12 @@ else:
                     else:
                         await sleep(self.walk_interval)
 
+        def add_strategy(self, overlay, strategy, target_peers):
+            with self.overlay_lock:
+                if overlay not in self.overlays:
+                    self.overlays.append(overlay)
+                self.strategies.append((strategy, target_peers))
+
         def unload_overlay(self, instance):
             with self.overlay_lock:
                 self.overlays = [overlay for overlay in self.overlays if overlay != instance]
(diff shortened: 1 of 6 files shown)


ipv8_service.py
```diff
@@ -167,6 +167,11 @@ async def on_tick(self):
                     else:
                         await sleep(self.walk_interval)
 
+        def add_strategy(self, overlay, strategy, target_peers):
```
egbertbouman, line 170: I'm wondering why we can only add a single strategy for a particular overlay. In Tribler, we have 8 communities of which 4 have more than 1 overlay. So, in its current form, its use is limited.

Also, the names don't seem to match: we have `add_strategy` to add an overlay/strategy, and then have `unload_overlay` to undo the operation.

qstokkink: In contrast to what the use of the _IPv8 framework_ suggests: the _`IPv8` object_ secretly does not manage overlays. it manages strategies. What actually _manages_ the overlays themselves are the endpoints. The strategies only serve to generate events/messages periodically with a "weak" reference to the overlays. We could (and possibly should) remove the `overlays` datastructure from the `IPv8` object.

The reason why we use overlays for the other methods, is that it is more natural in most program flows to keep a reference to the overlay, because it implements what the overlay programmer is interested in (the overlay).

It is also perfectly fine to add multiple strategies for the same overlay.

qstokkink:
retest this please

ipv8_service.py
qstokkink: I added an if statement to the `overlays.append` to make sure the `overlays` do not have duplicates.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

devos50:
retest this please

ipv8_service.py
egbertbouman: I'm sure developers using this method will take comfort in the fact *secretly* there is a reason for the inconsistent naming.

At least with the if statement we can add multiple strategies without making a complete mess of everything.

I reluctantly approve :shrug:

PR #793, opened 2020-07-02
Filter relay_candidates based on public_key

egbertbouman:


ipv8/messaging/anonymization/community.py
```diff
@@ -645,6 +644,8 @@ def on_cell(self, source_address, data):
         try:
             cell.decrypt(self.crypto, circuit=circuit, relay_session_keys=self.relay_session_keys.get(circuit_id))
         except CryptoException:
+            self.logger.debug("Could not decrypt cell from circuit %d (sender %s, receiver %s)",
```
qstokkink, line 647: I'm fine with merging this. However, I suggest to use the error message from the `CryptoException` itself, which has more information (see https://github.com/Tribler/py-ipv8/blob/master/ipv8/messaging/anonymization/payload.py).

egbertbouman: This exception will also happen during normal operating (right after a restart) and didn't want to spam the log too much. But on second thought, you're probably right. I'll change it.

PR #808, opened 2020-07-09
Don't load database on CommunicationManager start

qstokkink:
Fixes https://github.com/Tribler/tribler/issues/5450

### DIFF
--- a/ipv8/attestation/communication_manager.py
+++ b/ipv8/attestation/communication_manager.py
@@ -196,10 +196,18 @@ class CommunicationManager(object):
         else:
             if working_directory is None:
                 working_directory = ipv8_instance.configuration.get("working_directory", ".")
-            self.identity_manager = IdentityManager(working_directory if working_directory == ":memory:"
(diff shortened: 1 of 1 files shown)


qstokkink:
> 
> 
> I tested this and it seems to work with Tribler.
> 
> Which bring me to another question: I think it would be preferable to set the `working_directory` correctly, even when lazy loading the database. If there is any call in Tribler somehow that triggers the loading of the identity framework, the database is still created in the wrong location.

Yes this exists: you should set the IPv8 object's `working_directory` field for this.

qstokkink:
You can find that here by the way: https://github.com/Tribler/py-ipv8/blob/master/ipv8/configuration.py#L256-L263

qstokkink:
> Ok, could you make sure that the working_directory is correctly set in Tribler then? Also, could you update the IPv8 pointer on the release-7.5 branch?

Will do 👍

qstokkink:
@devos50 https://github.com/Tribler/tribler/pull/5462

PR #814, opened 2020-07-15
Removed log statement

devos50:
The log statement that I removed is clogging the Tribler logs and is uninformative in general.


devos50:
retest this please

devos50:
Note: the IPv8 unit tests now run under Python 3.8.4 on Windows 64-bit.

PR #819, opened 2020-07-18
Removed check_num_packets from TunnelExitSocket

egbertbouman:
After doing some tunnel performance monitoring, it turns out the circuits are often destroyed well before their time. Unfortunately, this is having a significant negative impact on performance.

Luckily we recently fixed an issue with the `DestroyPayload` not setting the `reason` field, so we can now see why this is happening. As it turns out the circuits are frequently destroyed because of [`check_num_packets`](https://github.com/Tribler/py-ipv8/blob/10836c5f70f0d381a6c808b0d4c24f5e07451d77/ipv8/messaging/anonymization/tunnel.py#L208). Given the fact that this restriction is rather arbitrary to begin with, I'd like to get rid of it.

### DIFF
--- a/ipv8/configuration.py
+++ b/ipv8/configuration.py
@@ -71,7 +71,6 @@ default = {
                     'max_time': 10 * 60,
                     'max_time_inactive': 20,
                     'max_traffic': 250 * 1024 * 1024,
-                    'max_packets_without_reply': 50,
                     'dht_lookup_interval': 30
                 }
             },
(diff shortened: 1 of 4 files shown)


qstokkink:
Can the `TunnelExitSocket.ips` attribute also be removed?

egbertbouman:
@qstokkink I forgot about that one. I just removed it.

PR #831, opened 2020-08-06
Added speed test functionality to TunnelCommunity

egbertbouman:
This PR:

- Removes `PEER_FLAG_EXIT_ANY`. From now on we will only use more specific flags: `PEER_FLAG_EXIT_BT/IPV8`.
- Removes `CIRCUIT_TYPE_IPV8`.
- Adds an `exit_flags` argument to `create_circuit` that allows the caller to specify which capabilities the circuit should have.
- Adds speed test functionality to the `TunnelCommunity`. It supports both upload and download tests. End-to-end circuits are also supported. By implementing special messages for the test (test-request/response), we are able to do a speed test without involving an outside source (e.g. a BitTorrent swarm). By adding a new peer flag (`PEER_FLAG_SPEED_TEST`) we extend the tunnel protocol without breaking compatibility.

### DIFF
--- a/ipv8/messaging/anonymization/community.py
+++ b/ipv8/messaging/anonymization/community.py
@@ -186,10 +186,9 @@ class TunnelCommunity(Community):
         for circuit_id in list(self.exit_sockets.keys()):
             self.remove_exit_socket(circuit_id, 'unload', remove_now=True, destroy=DESTROY_REASON_SHUTDOWN)
 
-        # First we unload the community, then we shutdown the RequestCache. This prevents calls to
-        # RequestCache.add after RequestCache.shutdown is called (which will return None after shutdown).
-        await super(TunnelCommunity, self).unload()
         await self.request_cache.shutdown()
+
+        await super(TunnelCommunity, self).unload()
 
     def _generate_circuit_id(self):
         circuit_id = random.getrandbits(32)
@@ -788,8 +787,7 @@ class TunnelCommunity(Community):
 
         self.logger.info("Extending circuit, got candidate with IP %s:%d from cache", *extend_candidate.address)
 
-        self.request_cache.add(CreateRequestCache(self, to_circuit_id, circuit_id,
-                                                  candidate, extend_candidate))
+        self.request_cache.add(CreateRequestCache(self, to_circuit_id, circuit_id, candidate, extend_candidate))
 
         self.send_cell(extend_candidate, "create",
                        CreatePayload(to_circuit_id, self.my_peer.public_key.key_to_bin(), payload.key))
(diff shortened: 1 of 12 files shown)


ichorid:
How to run the test? I mean, should it be executed by hand? Will it use the live network? What is the output?

egbertbouman:
@ichorid Yes, it uses the live network. There is a REST endpoint that allows you to run a test. Since it's in the Swagger docs, you can use the Swagger UI to run it and read about what it returns.

ipv8/messaging/anonymization/tunnel.py
```diff
@@ -186,6 +184,17 @@ def datagram_received(self, data, source):
         else:
             self.logger.warning("Dropping forbidden packets to exit socket with circuit_id %d", self.circuit_id)
 
+    def is_allowed(self, data):
+        is_bt = DataChecker.could_be_bt(data)
+        is_ipv8 = DataChecker.could_be_ipv8(data)
+
+        if not (is_bt and PEER_FLAG_EXIT_BT in self.overlay.settings.peer_flags) \
```
qstokkink, line 192: Could you confirm this conditional is correct?

`is_bt = True` and `self.overlay.settings.peer_flags = [PEER_FLAG_EXIT_BT]`, leads to a return value of `False`.

ipv8/messaging/anonymization/community.py
```diff
@@ -928,3 +933,42 @@ async def dht_peer_lookup(self, mid, peer=None):
             await self.dht_provider.peer_lookup(mid, peer)
         else:
             self.logger.error("Need a DHT provider to connect to a peer using the DHT")
+
+    def send_test_request(self, circuit, request_size=0, response_size=0):
+        cache = self.request_cache.add(TestRequestCache(self, circuit))
```
qstokkink, line 938: I don't think anyone will send a `send_test_request` while shutting down, but just for completeness: `request_cache.add()` can return `None`.

An easy way to avoid this would be:

```python
cache = TestRequestCache(self, circuit)
self.request_cache.add(cache)
```

ipv8/messaging/anonymization/tunnel.py
egbertbouman: If `is_bt and PEER_FLAG_EXIT_BT in self.overlay.settings.peer_flags` is `True`, the if-statement will not be executed and `is_allowed` will return `True`. So that should be OK.

Having said that, this really needs a unit test. I'll add that.

ipv8/messaging/anonymization/community.py
egbertbouman: Normally, I would fix this by calling `RequestCache.unload` until after the community is unloaded. However, in this case, the `RequestCache.add` is not called from within a message handler. So, you're right. I'll change it.

ipv8/messaging/anonymization/tunnel.py
qstokkink: Ah right, I see. Thanks, must be suffering a heatstroke here 🥵

egbertbouman:
The community should be able to deal with this. The `RetryRequestCache` does callback the community, but those calls shouldn't be a problem.

Having said that, it's a bad idea to increase complexity for a slightly lower LoC count. I'll change it in both `TunnelCommunity` and `HiddenTunnelCommunity`.

Issue #615, opened 2019-11-12
IPv8 v2.0, IPv6 support and network interface refactoring


qstokkink:
*Title is slightly suggestive: only one change is needed for all three*

Right now, `Peer` objects have a most up to date address (`.address`). In the future we will want to support multiple network interfaces (requiring multiple sockets, which probably need to be separately punctured).

The biggest change to our current codebase will be the refactoring of `Peer` objects to support a set/list of interface specifications. The `Peer` object will have to be in charge of implementing a `send()` and `puncture()` method (opening at least one communication channel), this is currently handled by the `Community` class.

Pros:
 - Overlay programmers have an opaque identity messaging layer, while still having control over the byte format of each message. No more worrying about addresses.
 - It just works, all of the network magic is in the IPv8 core.
 - More efficient: right now we just register any IPv4 we find as walkable and hope there is some user on the other side. If we bundle the possible IPv4 addresses under one key, one address working excludes the need of checking the other addresses.
 - We can store historical addresses per peer and attempt to reconnect to the last known interface.

Cons:
 - Need to open a socket for each interface (IPv4, IPv6, NFC?, ..)
 - Breaking change to all existing code.
 - Even though you really shouldn't need it in the first place, IPv4 addresses can no longer be seen as semi-stable identifiers.
 - A lot of work.

qstokkink:
I think the road map should be:

1. Refactor existing code to use `Peer.send` (and internally update the endpoint accordingly).
2. Create a new version of `Community` logic, that allows for multiple interfaces. This second step is a hard wire-protocol break for any overlays that use it - as the `IntroductionRequest` and `IntroductionResponse` are tied to IPv4. Some `Communities` will inevitably be left behind, like the `DiscoveryCommunity` that we have kept going for so long.

qstokkink:
After first attempt today, priority is as follows:

 - [ ] Don't use the IPv4 address in the DHT community.

qstokkink:
Once we have this abstraction layer, it would probably make sense to use [Scapy](https://pypi.org/project/scapy/).

[It supports many of the interfaces we would like to use](https://github.com/secdev/scapy/tree/master/scapy/layers), for example:

 - IPv6
 - DHCP
 - Bluetooth
 - TFTP
 - USB
 - ..and much more.

This should also solve our complex NAT local service discovery (#703).

qstokkink:
[...]

PR #832, opened 2020-08-06
Added multiple Peer interfaces support

qstokkink:
Related to #615

This PR:

 - Adds support for multiple addresses per `Peer` in the `Peer` and `Network` classes.

This is backwards compatible with the old style of using `.address`.

### DIFF
--- a/ipv8/peer.py
+++ b/ipv8/peer.py
@@ -2,6 +2,7 @@ from base64 import b64encode
 from collections import deque
 from struct import unpack
 from time import time
+from typing import Any, Optional, Tuple
 
 from .keyvault.crypto import default_eccrypto
 from .keyvault.keys import Key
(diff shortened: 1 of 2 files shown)


ipv8/peer.py
```diff
@@ -24,11 +24,44 @@ def __init__(self, key, address=("0.0.0.0", 0), intro=True):
...
 
+    @property
+    def address(self) -> Tuple[str, int]:
+        """
+        Deprecated way to retrieve the IPv4 address.
+
+        Use the ``.addresses`` dictionary instead!
+        """
+        from .messaging.interfaces.udp.endpoint import UDPv4Address
+        return self.addresses.get(UDPv4Address, self.addresses.get(tuple, ("0.0.0.0", 0)))
```
egbertbouman, line 42: Are we sure we want to start mixing `tuple`/`UDPv4Address` addresses? I imagine that this could lead to unexpected problems (e.g. in `Network`).

qstokkink: I intentionally made `UDPv4Address` a `collections.namedtuple` for this reason, so there are no compatibility issues. Since the last PR we already use `UDPv4Address` from the Endpoint itself and perform this mixing.

We only need this particular Peer logic to deal with Community logic that recreates addresses as `tuple` instances and sets the Peer `address` manually. Thankfully, there are hardly any instances of Communities manually creating addressses and inserting them. Changing the Community logic will be a next PR though.

PR #838, opened 2020-08-07
Undeprecated Peer.address

qstokkink:
Related to #615

This PR:

 - Adds `DirtyDict`, a dictionary subclass that sets a `dirty` flag when it is modified.
 - Updates `Peer.address` to no longer be deprecated.
 - Updates `Peer.address` to be restricted to tuples of `(str, int)`.
 - Updates `Peer.addresses` to be read-only for performance of `Peer.address`.

This implementation focuses on speed instead of maintainability as `.address` is called very often.

### DIFF
--- a/ipv8/peer.py
+++ b/ipv8/peer.py
@@ -2,15 +2,51 @@ from base64 import b64encode
 from collections import deque
 from struct import unpack
 from time import time
-from typing import Any, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple, Type
 
 from .keyvault.crypto import default_eccrypto
 from .keyvault.keys import Key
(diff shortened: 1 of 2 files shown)


qstokkink:
Profiled with wild runs, new implementation is just over 3x faster:

```
NEW (percall: 2.95857 microseconds):

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2028    0.003    0.000    0.006    0.000 /home/quinten/Documents/py-ipv8/ipv8/peer.py:81(address)
     
        
OLD (percall: 9.6313 microseconds):

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     6645    0.046    0.000    0.064    0.000 /home/quinten/Documents/py-ipv8/ipv8/peer.py:34(address)
```

PR #839, opened 2020-08-10
Add message ID to tunnel payload classes

egbertbouman:
This PR:
- Adds `Payload.msg_id` to the tunnel payload classes, so we can use `add_message_handler`.
- Adds `add_cell_handler` for messages that are wrapped in a `Cell`.
- Fixes some cases in which control packets were not included in `bytes_up/bytes_down`.
- Adds `vp_compile` decorator wherever possible to speed up message serialization.
- Renames `tc_lazy_wrapper_unsigned` to `unpack_cell`.


egbertbouman:
@qstokkink Thanks!

Issue #841, opened 2020-08-10
Get rid of Python 2 ord() compatibility


qstokkink:
We have a quite a few instances of `ord()` being used on `bytes` objects, which was a Python2/3 compatibility measure.

In Python 3 we no longer have to use 1 character `bytes` slices and call `ord()`. Instead, getting the item at the specific character position is automatically an `int`.

An example:

https://github.com/Tribler/py-ipv8/blob/4f073075ab594b3a203d49757b29a56425b60b34/ipv8/attestation/wallet/primitives/attestation.py#L32

This could be replaced with:

```python
out |= hashed[i]
```

PR #843, opened 2020-08-12
Removed unnecessary ord calls

egbertbouman:
Fixes https://github.com/Tribler/py-ipv8/issues/841.

### DIFF
--- a/ipv8/attestation/trustchain/caches.py
+++ b/ipv8/attestation/trustchain/caches.py
@@ -24,7 +24,7 @@ class IntroCrawlTimeout(NumberCache):
         """
         charlist = []
         for i in range(len(peer.mid)):
-            charlist.append(ord(peer.mid[i:i + 1]))
+            charlist.append(peer.mid[i])
         return reduce(lambda a, b: ((a << 8) | b), charlist, 0)
 
     @property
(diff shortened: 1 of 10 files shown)


qstokkink:
Would you also be willing to fixup the remainder of `M2CryptoPK.verify`? It seems to be a bit.. broken.

egbertbouman:
@qstokkink Sure, I'll take a look.

qstokkink:
Thanks 👍. Specifically the `str` (`"\x00"`) + `bytes` object concatenations don't seem like they could possibly be correct.

egbertbouman:
It's fixed now and I've added some unit tests to test signing/verifying.

I did notice that even though `M2CryptoSK` inherits from `M2CryptoPK` and therefore has a `verify` method, it does not work. You first have to get the `M2CryptoPK` by calling `pub()`, and then it does work. `LibNaCLSK` doesn't suffer from this.

qstokkink:
> I did notice that even though M2CryptoSK inherits from M2CryptoPK and therefore has a verify method, it does not work. You first have to get the M2CryptoPK by calling pub(), and then it does work. LibNaCLSK doesn't suffer from this.

I just checked this out.. might have something to do with them not calling their `super(...).__init__`? We'll have to look at it, but let's not include that in this PR.

qstokkink:
Follow-up: the `verify` issue has nothing to do with `super(...).__init__` apparently.

qstokkink:
..and everything with the `try .. except` hiding the error 😃 I'll make a fix for this.

PR #850, opened 2020-08-14
Switch decode_map to list

qstokkink:
This PR:

 - Fixes `deprecated_message_names` indexing on `chr` and being looked-up with `int`.
 - Updates `decode_map` to be a list instead of a dictionary.

This _should_ be faster and more memory efficient. I'll produce some perfomance results soon, if this is not faster I'll close this PR again.


qstokkink:
This is actually faster, though not by much. The average time spent in the `on_packet` function has been reduced from 118 microseconds down to 106 microseconds (10% faster).

qstokkink:
The `list` switch will mostly be helpful if we ever want to port this low-level stuff to C.

PR #866, opened 2020-09-19
Added pytest support

### LINKED ISSUE #864: The deadlock checker in the IPv8 tests is not compatible with pytest
It seems that the custom deadlock detection logic in IPv8 does not play nicely with `pytest`. I am using the `pytest-flakefinder` tool to detect flaky tests in the Tribler core tests. This tool runs each test a specific number of times across different parallel worker processes. This work great for the tests that are exclusively using `pytest`. However, we have a few tests that are inheriting from IPv8, specifically the overlay tests for the popularity community and tunnel community.

Now, when running these tests with `pytest`, the worker process is killed after around 10 seconds into the `TestTriblerTunnelCommunity` tests. Setting the `MAX_TEST_TIME` to a higher value fixes this problem, however, it feels like a workaround and not a dedicated fix. It would be helpful if there is a feature to disable the deadlock checker for these Tribler tests.

### LINKED ISSUE #865: The temporary_directory method fails when the system random seed is fixed
The `temporary_directory` method in the tests fails when one seeds the system `random` with a fixed value. I found this when running the IPv8 overlay tests in the Tribler repository with `pytest-randomly`, which seeds `random`. Specifically, different tests running in parallel will then use the same directory and fail the test.

For example:

```
name = '/home/jenkins/ci/workspace/pers/pytest_linux_flake_devos50/src/TestRemoteQueryCommunity_temp_R006OC1EKX'
mode = 511, exist_ok = False

    def makedirs(name, mode=0o777, exist_ok=False):
        """makedirs(name [, mode=0o777][, exist_ok=False])
    
        Super-mkdir; create a leaf directory and all intermediate ones.  Works like
        mkdir, except that any intermediate path segment (not just the rightmost)
        will be created if it does not exist. If the target directory already
        exists, raise an OSError if exist_ok is False. Otherwise no exception is
        raised.  This is recursive.
    
        """
        head, tail = path.split(name)
        if not tail:
            head, tail = path.split(head)
        if head and tail and not path.exists(head):
            try:
                makedirs(head, exist_ok=exist_ok)
            except FileExistsError:
                # Defeats race condition when another thread created the path
                pass
            cdir = curdir
            if isinstance(tail, bytes):
                cdir = bytes(curdir, 'ASCII')
            if tail == cdir:           # xxx/newdir/. exists if xxx/newdir exists
                return
        try:
>           mkdir(name, mode)
E           FileExistsError: [Errno 17] File exists: '/home/jenkins/ci/workspace/pers/pytest_linux_flake_devos50/src/TestRemoteQueryCommunity_temp_R006OC1EKX'

/usr/lib/python3.7/os.py:221: FileExistsError
```

Maybe it is a better idea to use a `Random` instance to ensure seed uniqueness in the `temporary_directory` method?

qstokkink:
Fixes #864
Fixes #865 

This PR:

 - Updates `TestBase.setUpClass` to detect `pytest` and disable the class lock-up detection.
 - Updates `TestBase.temporary_directory` to use `uuid` instead of 10x `random.choice`.

Note: the body of the if-statement for the first change is the same as before.

### DIFF
--- a/ipv8/test/base.py
+++ b/ipv8/test/base.py
@@ -1,10 +1,9 @@
 import os
-import random
 import shutil
-import string
 import sys
 import threading
 import time
+import uuid
 from asyncio import all_tasks, get_event_loop, sleep
 
 import asynctest
@@ -155,7 +154,7 @@ class TestBase(asynctest.TestCase):
         await self.deliver_messages()
 
     def temporary_directory(self):
-        rndstr = '_temp_' + ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
+        rndstr = '_temp_' + uuid.uuid4().hex
         d = os.path.abspath(self.__class__.__name__ + rndstr)
         self._tempdirs.append(d)
         os.makedirs(d)
--- a/ipv8/test/base.py
+++ b/ipv8/test/base.py
@@ -63,39 +63,41 @@ class TestBase(asynctest.TestCase):
     def setUpClass(cls):
         TestBase.__lockup_timestamp__ = time.time()
 
-        def check_loop():
-            while time.time() - TestBase.__lockup_timestamp__ < cls.MAX_TEST_TIME:
-                time.sleep(2)
-                # If the test class completed normally, exit
-                if not cls.__testing__:
-                    return
-            # If we made it here, there is a serious issue which we cannot recover from.
-            # Most likely the threadpool got into a deadlock while shutting down.
-            import traceback
-            print("The test-suite locked up! Force quitting! Thread dump:", file=sys.stderr)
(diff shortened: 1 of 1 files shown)


qstokkink:
I'm sure this still doesn't offer full support for pytest, but let me know if this is sufficient for your needs @devos50 .

devos50:
Thanks, I'll test it out!

devos50:
[Seems to work!](https://jenkins-ci.tribler.org/job/pers/job/pytest_linux_flake_devos50/8/console) There are still two failures left but they are unrelated to IPv8.

Thanks 👍

devos50 (approved):

qstokkink:
Cool, let me know if you run into anything else.

PR #878, opened 2020-10-02
Replaced master_peer with a 20-byte community_id

egbertbouman:
This PR:
* Replaces the `master_peer` of every `Community` with a byte string. In the past, this was used by `Dispersy` as part of a mechanism to remotely shutdown a `Community`. However, since we no longer have this feature, we can easily replace this with a byte string.
* Removes `cast_to_chr` and `char_to_unicode` that were used while porting from Python 2 to 3.
* Possibly fixes intermittent "unregistered address" error during `EdgeWalk` tests.

### DIFF
--- a/doc/basics/overlay_tutorial.rst
+++ b/doc/basics/overlay_tutorial.rst
@@ -106,6 +106,7 @@ To do this, fill your ``main.py`` file with the following code:
 
 .. code-block:: python
 
+    import os
     from asyncio import ensure_future, get_event_loop
 
     from pyipv8.ipv8.community import Community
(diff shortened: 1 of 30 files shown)


qstokkink (changes requested):
Please update https://github.com/Tribler/py-ipv8/blob/master/doc/basics/overlay_tutorial.rst to reflect your changes.

egbertbouman:
@qstokkink I forgot about the docs. It's fixed now.

qstokkink (approved):
Nice 👍 This has been on my wishlist for a while.

Issue #872, opened 2020-09-25
Add identity test Jenkins job


qstokkink:
The identity tutorial (https://py-ipv8.readthedocs.io/en/latest/basics/identity_tutorial/) describes a set of steps to achieve attribute attestation and verification. We could add a Jenkins job that verifies the described behavior.

qstokkink:
For reference, I copied over the scripts I use for manual verification to Gist:

 - [Attestation test scripts ](https://gist.github.com/qstokkink/6f5bec769dae981a5eef67e57e6741a3)
 - [Deprecated Attestation test scripts](https://gist.github.com/qstokkink/33d988fb81053f6349aa35ddbec65a01)
 - [Other documentation tests](https://gist.github.com/qstokkink/32488e9a250649333b95aaa10886c4cb)

These should at some point be converted to automated Jenkins jobs (properly matching the outputs to what the tutorial claims). The 'other' documentation checks could possibly be automatically inserted into the documentation automagically through RST embedding.

qstokkink:
The new attestation test script also still needs to integrate the advanced identity tools, given here: https://py-ipv8.readthedocs.io/en/latest/further-reading/advanced_identity/

devos50:
You could consider integrating these jobs in [our validation experiment pipeline](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiments/).

qstokkink:
That seems like a good idea. We _should_ be able to make a configurable MultiJob that takes a `branch` argument, which we can then use to both test feature branches in IPv8 PRs and then also use for the IPv8 version shipped with Tribler (in that validation pipeline).

That will probably take some Jenkins-fu to correctly set-up though 😄

qstokkink:
Before we tackle this, all code exmples should be first extracted into separate files: https://www.sphinx-doc.org/en/1.5/markup/code.html#includes
Otherwise we will need to maintain a duplicate per example.

PR #886, opened 2020-10-15
Extracted documentation examples into Python files

qstokkink:
First step for #872.

This PR:

 - Updates the documentation to include examples from stand-alone files, instead of embedding them.
 - Updates the serialization documentation to prefer `ez_send` over `ezr_pack`.

This is the "easy" part of #872. In a following PR I will add new functionality to run these examples and check for errors.


qstokkink:
Ready for review: the remaining Pylint errors are due to the documentation expecting a different import structure.

egbertbouman (approved):

PR #892, opened 2020-10-17
Updated DHTCommunity serialization

egbertbouman:
This PR:
* Adds the ability to register new packers with the `Serializer`.
* Adds `ListOf` that allows for almost any packer (except for `NestedPayload` and `Raw`) to be turned into a list.
 Now you can do stuff like `serializer.add_packer('infohashes', ListOf(DefaultStruct('20s')))`.
Besides its use in the `DHTCommunity`, I'd also like to use it in the `TunnelCommunity` payloads and get rid of `encode`/`decode`.
* Adds `ipv4` packer as an alternative to `4SH`. The advantage of this packer is that it deals with IP address serialization and avoids the use of `fix_` functions.
* Removes global time from DHT payloads  Since we're breaking the tunnel protocol by changing the payout mechanism in Tribler, this is a good time to update the DHT protocol.
* Removes `send_message` from the `DHTCommunity` and uses `ez_send` instead.
* Improved the way the `DHTCommunity` deals with the possibility of `RequestCache.add` returning `None`.

### DIFF
--- a/doc/reference/serialization.rst
+++ b/doc/reference/serialization.rst
@@ -84,7 +84,7 @@ Datatypes
 
 Next to the unsigned integer and unsigned short data types, the default Serializer has many more data types to offer.
 The following table lists all data types available by default, all values are big-endian and most follow the default Python ``struct`` format.
-A ``Serializer`` can be extended with arbitrary ``struct`` formats by calling ``serializer.add_packing_format(name, format)`` (for example ``serializer.add_packing_format("I", ">I")``).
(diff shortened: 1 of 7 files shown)


qstokkink (changes requested):
Very cool, but could you please also update the documentation here:

https://github.com/Tribler/py-ipv8/blame/1da00158130390adbac99f6e9a958869828c1a89/doc/reference/serialization.rst#L87

qstokkink:
Addendum: and please make sure to add the two new formats to the table below the line I referenced too.

egbertbouman:
Docs are now updated, and I also added the missing `q` format.

qstokkink:
Cool, thanks 👍 The omitted `q` was my bad. Now, for the final trick:

qstokkink:
validate

qstokkink (approved):
All green 🙆‍♂️ Thanks!

egbertbouman:
Great, thanks! Now I can have fun with variable length lists.

PR #893, opened 2020-10-20
Added static typing support

### LINKED ISSUE #879: Provide top-level types file
Over the past few months I have been opportunistically adding typing to the codebase. However, one of the issues I have experienced with this approach is that to type functions, one needs to import the actual object. [This can lead to circular imports](https://www.stefaanlippens.net/circular-imports-type-hints-python.html) and beyond-top-level-import errors.

At this point in time, I see the ideal solution for this to be _conditional imports_ that are _only performed when checking types_: [PEP 563 Runtime annotation resolution and TYPE_CHECKING](https://www.python.org/dev/peps/pep-0563/#runtime-annotation-resolution-and-type-checking). If we provide all of the classes in IPv8 with an import-safe definition in a top-import-level `types.py`, we could have a circular-import-safe normal-typing-compatible way to statically type-check our code.

Practically, this should materialize as follows:

`ipv8/types.py`

```python
import typing

if typing.TYPE_CHECKING:
    # Absolute imports:
    from ipv8_service import IPv8
    from ipv8.database import Database
     ... et cetera ...
```

`ipv8/foo/bar.py`

```python
import ..types

def foo(ipv8: IPv8):
    return True
```

We can implement this incrementally and we should definitely start with a small part of the codebase first and evaluate this approach.

qstokkink:
Fixes #879

This PR:

 - Adds a top-level file for typing, reducing the amount of imports only used for typing (possibly creating circular dependencies).
 - Adds support for running Mypy on the codebase (using `mypy ipv8` from the root folder).
 - Fixes Mypy errors (fixing errors includes adding missing types in some cases).

Note that this PR **does not** add typing to the entire codebase: it only fixes what was already there.

## Lessons learned (and applied)

 - Use PEP 563 for class and static methods. For example:

```python
from __future__ import annotations  # Included PEP 563 in Python 3.7

class A:
    @classmethod
    def create(cls) -> A:  # This wont work without the future import, A is not defined/loaded yet
        return cls
```

 - Use PEP 484 and PEP 563 to avoid circular imports and unnecessary/typing-only imports:

```python
# Top-level definitions in types.py
if typing.TYPE_CHECKING:  # PEP 563
    # This may be a circular import, but Mypy doesn't care for its static analysis
    from ipv8.database import Database
else:
    # This is what is actually used at running time
    Database = 'ipv8.database.Database'  # PEP 484
```

 - Mypy gets confused sometimes. Our Pylint configuration doesn't like specific Mypy ignores though, use this instead:

```python
a = 1
a_str = str(a)  # Preferred solution

# The following is a type conflict, sometimes this cannot be avoided
# NOTE: Don't use specific ignores like "type:ignore[override]", it throws off Pylint.
a = str(a)  # type:ignore
```

 - Subtypes are automatically inferred and may need adjustment in the function body:

```python
class A:
    pass
class B1(A):
    pass
class B2(A):
    pass
a: A = B1()  # Need type hint here, otherwise type will be inferred as B1, causing a conflict on the next line
a = B2()  # No type hint needed here as "a" was already introduced
```

 - Mypy and `OrderedDict` don't get along, use `OrderedDict` without subscript.

 - Mypy needs static type definitions. To play nicely with VariablePayloads, you can define valueless class attributes with types:

```python
@vp_compile
class MyPayload(VariablePayload):
    msg_id = 1
    format_list = ['I', 'H', 'varlenH', 'bits']
    names = ['a', 'b', 'c', 'd']

    a: int
    b: int
    c: bytes
    d: typing.List[bool]
```

### DIFF
--- a/ipv8/attestation/communication_manager.py
+++ b/ipv8/attestation/communication_manager.py
@@ -9,12 +9,13 @@ from .identity.manager import IdentityManager
 from .wallet.community import AttestationCommunity
 from ..keyvault.crypto import ECCrypto
 from ..messaging.anonymization.hidden_services import HiddenTunnelCommunity
+from ..types import IPv8, Peer
 from ..util import succeed
 
 
 class CommunicationChannel(object):
 
-    def __init__(self, attestation_community, identity_community):
+    def __init__(self, attestation_community: AttestationCommunity, identity_community: IdentityCommunity):
         super(CommunicationChannel, self).__init__()
         self.attestation_overlay = attestation_community
         self.identity_overlay = identity_community
@@ -29,18 +30,19 @@ class CommunicationChannel(object):
         self.attestation_overlay.set_verify_request_callback(self.on_verify_request)
 
     @property
-    def public_key_bin(self):
+    def public_key_bin(self) -> bytes:
         return self.identity_overlay.my_peer.public_key.key_to_bin()
 
     @property
-    def peers(self):
+    def peers(self) -> typing.List[Peer]:
         return self.identity_overlay.get_peers()
 
     @property
-    def schemas(self):
+    def schemas(self) -> typing.List[str]:
         return list(self.attestation_overlay.schema_manager.formats.keys())
 
-    def on_request_attestation(self, peer, attribute_name, metadata):
+    def on_request_attestation(self, peer: Peer, attribute_name: str,
+                               metadata: typing.Dict[str, str]) -> asyncio.Future:
         """
         Return the measurement of an attribute for a certain peer.
         """
@@ -49,7 +51,8 @@ class CommunicationChannel(object):
         self.attestation_metadata[(peer, attribute_name)] = metadata
         return future
 
-    def on_attestation_complete(self, for_peer, attribute_name, attribute_hash, id_format, from_peer=None):
+    def on_attestation_complete(self, for_peer: Peer, attribute_name: str, attribute_hash: bytes, id_format: str,
+                                from_peer: Peer = None) -> None:
         """
         Callback for when an attestation has been completed for another peer.
         We can now sign for it.
@@ -65,7 +68,7 @@ class CommunicationChannel(object):
             self.identity_overlay.add_known_hash(attribute_hash, attribute_name, for_peer.public_key.key_to_bin(),
                                                  metadata)
 
-    def on_verify_request(self, peer, attribute_hash):
+    def on_verify_request(self, peer: Peer, attribute_hash: bytes) -> asyncio.Future:
         """
         Return the measurement of an attribute for a certain peer.
         """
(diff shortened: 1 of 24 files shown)


ipv8/attestation/communication_manager.py
```diff
@@ -29,18 +34,19 @@ def __init__(self, attestation_community, identity_community):
         self.attestation_overlay.set_verify_request_callback(self.on_verify_request)
 
     @property
-    def public_key_bin(self):
+    def public_key_bin(self) -> bytes:
         return self.identity_overlay.my_peer.public_key.key_to_bin()
 
     @property
-    def peers(self):
+    def peers(self) -> typing.List[Peer]:
```
devos50, line 41: Why not use `List` here? I think that is more explicit and clear. The fact that it's in the `typing` library should not be relevant here.

qstokkink: This is faster* and `typing` was already imported anyway. Here's the proof:

```python
print(timeit.timeit("import typing; a = typing.List[str]", number=1000000))
> 2.016579663999437
print(timeit.timeit("import typing; from typing import List; a = List[str]", number=1000000))
> 4.145822678998229
```

Is `List` more readable than `typing.List`? Yes**. However, I don't believe it's such a dramatic improvement that it warrants the 2x slow-down though.

*Because it happens at module-load time and not at running time (e.g. in a loop).
**Sort of, `typing.List` leaves less room for ambiguity.

---

All of that said, this is a discussion about aesthetics, personal taste and microsecond optimizations. I implore you not to start this discussion here and let this one slide 😃

qstokkink:
validate

qstokkink:
validate

qstokkink:
validate

qstokkink:
validate

qstokkink:
🤔 Seems like something is interfering with the validation.

qstokkink:
validate

qstokkink:
validate

qstokkink:
Nvm. that was my bad 😄

qstokkink:
validate

qstokkink:
validate

qstokkink:
Status: deprecated attestation tutorial validation check is failing. Not changed in this PR and working on my machine. Assuming some sort of socket error or leaked process, I'll validate again in +- 30 minutes.

qstokkink:
validate

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
validate

qstokkink:
validate

qstokkink:
Tribler ran for 2 hours without error on this IPv8 PR code, all the checks pass: ready for review.

xoriole (approved):

PR #895, opened 2020-10-20
Renamed MockIPv8.unload to MockIPv8.stop

devos50:
This is consistent with the real IPv8 object, which does not have an unload but a stop method.

This PR:

 - Renames MockIPv8.unload to MockIPv8.stop
 - Fixes a typo in the DHT community

### DIFF
--- a/ipv8/test/REST/rest_base.py
+++ b/ipv8/test/REST/rest_base.py
@@ -61,7 +61,7 @@ class MockRestIPv8(object):
         await self.rest_manager.start(0)
         self.rest_port = self.rest_manager.site._server.sockets[0].getsockname()[1]
 
-    async def unload(self):
+    async def stop(self, stop_loop=True):
         await self.rest_manager.stop()
         self.endpoint.close()
         for overlay in self.overlays:
--- a/ipv8/dht/discovery.py
+++ b/ipv8/dht/discovery.py
@@ -139,7 +139,7 @@ class DHTDiscoveryCommunity(DHTCommunity):
             self.logger.debug('Storing peer %s (key %s)', node, hexlify(payload.target))
             self.store[payload.target].append(node)
 
-        self.ez_send(node, StorePeerResponsePayload(payload.indentifier))
+        self.ez_send(node, StorePeerResponsePayload(payload.identifier))
 
     @lazy_wrapper(StorePeerResponsePayload)
     def on_store_peer_response(self, peer, payload):
--- a/ipv8/test/base.py
+++ b/ipv8/test/base.py
@@ -51,7 +51,7 @@ class TestBase(asynctest.TestCase):
     async def tearDown(self):
         try:
             for node in self.nodes:
-                await node.unload()
+                await node.stop()
             internet.clear()
         finally:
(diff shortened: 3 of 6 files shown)


devos50:
retest this please

qstokkink:
Why was this retested?

devos50:
Because Jenkins was down :)

ipv8/test/REST/rest_base.py
```diff
@@ -61,7 +61,7 @@ async def start_api(self):
         await self.rest_manager.start(0)
         self.rest_port = self.rest_manager.site._server.sockets[0].getsockname()[1]
 
-    async def unload(self):
+    async def stop(self, **_):
```
qstokkink, line 64: Please remove the `**_` instances.

qstokkink (changes requested):

ipv8/test/REST/rest_base.py
devos50: The original `stop` method does have a `stop_loop` keyword argument:

```
async def stop(self, stop_loop=True):
```

Therefore, if I would use the `MockIPv8` in a Tribler session and implement the requested change, it would be called with this keyword argument, and failing.

qstokkink: That is a fair use case, but then you are breaking with the function signature. Please use `stop_loop` instead of `_` then.

qstokkink: Actually, if this is used as an `IPv8` drop-in replacement, it should really be a `IPv8` subclass.. but that's for another PR.

devos50: Implemented, at the cost of two pylint errors.

qstokkink:
validate

devos50:
Wait, is this one expected?!

```
ERROR:DHTDiscoveryCommunity:Exception occurred while handling packet!
  File "/home/jenkins/ci/workspace/ipv8/validate_documentation/pyipv8/ipv8/community.py", line 331, in on_packet
    result = handler(source_address, data)
  File "/home/jenkins/ci/workspace/ipv8/validate_documentation/pyipv8/ipv8/lazy_community.py", line 38, in wrapper
    return func(self, Peer(auth.public_key_bin, source_address), *unpacked)
  File "/home/jenkins/ci/workspace/ipv8/validate_documentation/pyipv8/ipv8/dht/discovery.py", line 142, in on_store_peer_request
    self.ez_send(node, StorePeerResponsePayload(payload.indentifier))
AttributeError: 'StorePeerRequestPayload' object has no attribute 'indentifier'
```

qstokkink:
Uh oh, we'll have to fix that.

qstokkink:
@devos50 do you want to fix that in this PR or do you want to wait and rebase on another PR that fixes this?

devos50:
Fixed it in this PR 👍

qstokkink:
Thanks 👍

qstokkink:
validate

devos50:
Seems to pass now 👍

qstokkink (approved):

PR #910, opened 2020-10-30
Made packet handling fragile in TestBase

qstokkink:
Related to https://github.com/Tribler/tribler/issues/5683

This PR:

 - Adds fragile packet handling in `TestBase`.
 - Updates `TestBase` to set itself as the `asyncio` exception handler.

Notable caveats:

 - With the way `asyncio` works, we can only sweep up errors after a test run has completed. This means that compound errors may cause the root of the issue to be shadowed. For example the situation `message handler has exception and fails to execute some code -> class invariant is violated elsewhere -> exception` would seem like a case of `class invariant is violated elsewhere -> exception` instead of `message handler has exception`. Chronologically speaking: the last exception would be triggered first.
 - Any node `i` inserted after `initialize()` will need to have `patch_overlays(i)` called manually.
 - ~~Multiple failures may overwrite each others exception. Only one exception can cause the `TestCase` to fail at a time, so preceeding failures may be hidden.~~ Solved.

Notable "don't"s:

 - Don't overwrite `_cb_exception` and cause an exception there. This will lead to silent failures (this is what happened before) - this is not run in the main loop!

### DIFF
--- a/ipv8/test/base.py
+++ b/ipv8/test/base.py
@@ -4,7 +4,8 @@ import sys
 import threading
 import time
 import uuid
-from asyncio import all_tasks, get_event_loop, sleep
+from asyncio import all_tasks, ensure_future, get_event_loop, iscoroutine, sleep
+from functools import partial
 
 import asynctest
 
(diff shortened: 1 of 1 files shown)


qstokkink:
Quickly back to WIP: I can solve one of the downsides.

ichorid:
:+1: 

Well, I guess there is nothing Quinten can't achieve if he is intellectually provoked enough :wink:

ichorid (approved):

Issue #908, opened 2020-10-29
Trustchain move to anydex-core


qstokkink:
Our old Trustchain implementation is moving to the anydex-core repository, the files that will need changes are:

 - [x] `ipv8_service.py`
 - [x] `doc/index.rst`
 - [x] `doc/basics/crawler_service_tutorial.rst`
 - [x] `doc/further-reading/trustchain.rst`
 - [x] `doc/reference/configuration.rst`
 - [x] `ipv8/attestation/trustchain/` remove entirely
 - [x] `ipv8/attestation/wallet/community.py`
 - [x] `ipv8/messaging/deprecated` remove entirely
 - [x] `ipv8/REST/attestation_endpoint.py`
 - [x] `ipv8/REST/root_endpoint.py`
 - [x] `ipv8/REST/trustchain_endpoint.py`
 - [x] `ipv8/test/attestation/trustchain/` remove entirely
 - [x] `ipv8/test/mocking/ipv8.py`
 - [x] `ipv8/test/test_configuration.py`
 - [x] `scripts/exitnode_ipv8_only_plugin.py`
 - [x] `scripts/trustchain_crawler_plugin.py`
 - [x] `systemd/trustchain-crawler.service`

devos50:
You can probably also remove [the encoding implementation](https://github.com/Tribler/py-ipv8/blob/master/ipv8/messaging/deprecated/encoding.py) since I will use JSON serialization for transactions within blocks instead (if this encoding is not used anywhere else in IPv8).

devos50:
I merged TrustChain into AnyDex in https://github.com/Tribler/anydex-core/pull/55. When we are sure that we are not making another 7.5.x release, you can safely remove TrustChain from IPv8 👍

qstokkink:
Nice 👍 I will probably strip this out for the big 3.0 release then, when we also add IPv6 support.

qstokkink:
As this touches a lot of files, I'll tag this as "on hold" until #615 is resolved.

qstokkink:
Instead of waiting for #615, this issue will be `on hold` until the Tribler 7.6.0 official release (officially ending the downstream dependency on the deprecated Trustchain protocol).

qstokkink:
This issue is now out of lockdown as Tribler 7.6.0 has been released.

devos50:
> out of lockdown

It's a bit sad that this issue gets earlier out of lockdown than the Tribler developers 😢

PR #911, opened 2020-11-02
Add address packer to default Serializer

egbertbouman:
This PR:
* Adds the `Address` packer to the default `Serializer`. Unfortunately, to maintain compatibility with the `TunnelCommunity`, it's using a `varlenH` for domain names (wasting an extra byte).
* Updates `CellPayload` to catch the correct exception while doing crypto.
* Updates `PexCommunity` to stop using encode/decode. Now that `TrustChain` is the only community still using it, this functionality can be removed in the future (see https://github.com/Tribler/py-ipv8/issues/908).
Note that the `PexCommunity` does a +1 on the `community_id`, instead of using `version_id`. This is because the trackers aren't looking at the first 2 bytes of the prefix, which could lead to being introduced to old `PexCommunities`. This would make peer discovery take longer.

### DIFF
--- a/ipv8/messaging/anonymization/community.py
+++ b/ipv8/messaging/anonymization/community.py
@@ -155,7 +155,6 @@ class TunnelCommunity(Community):
 
     def get_serializer(self):
         serializer = super().get_serializer()
-        serializer.add_packer('address', Address())
         serializer.add_packer('flags', Flags())
         return serializer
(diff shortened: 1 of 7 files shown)


qstokkink (approved):
Props for removing deprecated functionality in the `PexCommunity` 👍

PR #920, opened 2020-11-12
Add Sentry support to the exit node script

drew2a:
This PR:

 - Adds Sentry support for the exit node script

### Prerequisites

```bash
 python -m pip install sentry-sdk
```

### Configuration

nix:
```bash
EXPORT SENTRY_URL=<sentry_url>
```

windows:
```windows shell
SETX SENTRY_URL <sentry_url>
```

Changes in ansible roles: https://github.com/Tribler/tribler-ansible-roles/pull/1

PS: cool PR template 👍

### DIFF
--- a/scripts/exitnode_ipv8_only_plugin.py
+++ b/scripts/exitnode_ipv8_only_plugin.py
@@ -1,5 +1,22 @@
 """
 This script enables to start IPv8 headless.
+
+For error reporting Sentry is used.
+If sentry-sdk is not installed or SENTRY_URL is not defined, then no reports
+will be sent.
+
+Prerequisites:
+    python -m pip install sentry-sdk
+
+Configuring Sentry:
+    nix:
+    ```
+        EXPORT SENTRY_URL=<sentry_url>
+    ```
+    windows:
+    ```
+        SETX SENTRY_URL <sentry_url>
+    ```
 """
 import argparse
 import os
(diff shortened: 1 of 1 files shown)


devos50:
ok to test

drew2a:
Flake8 and Pylint want the different import order:
https://github.com/Tribler/py-ipv8/pull/920/commits/8f3a7db3379165e8cfbc392bd868cd933f2e73df
https://github.com/Tribler/py-ipv8/pull/920/commits/0c11660c3b7053c750b1d0f043bb07cb9684d182

That's why I've added `# noqa: I100`.
If you have a better solution, please feel free to suggest.

qstokkink:
> If you have a better solution, please feel free to suggest.

This is the "correct" import order (or at least, the consistent one).

As IPv8 and all of its scripts are to be run by everyone and not just the Tribler team, forcefully tying them into centralized reporters is not acceptable. Please make the sentry import optional instead, something like:

```python
try:
    import sentry_sdk
except ImportError:
    sentry_sdk = None

 ...

if __name__ == "__main__":
    if sentry_sdk:
        sentry_sdk.init(os.environ.get('SENTRY_URL'), traces_sample_rate=1.0)
```

drew2a:
If no `SENTRY_URL` is defined, then no reports are sent, that's why the original code still works both with centralized error reporters and without them.

Maybe I should add additional comments that will describe this behavior.

But, yes I force everyone to install `sentry-sdk` now. 
I think this approach is ok because it makes code more clear than adding "try-except" in the import section. 

Otherwise, if we will have two or more "optional" dependencies in the future, the code can look really horrible.

Don't get me wrong, I'm ready to change the original code to the "try-except" version, just consider it as a bad practice.

drew2a:
@qstokkink fixed.

qstokkink (approved):
Nice, thanks 👍

Issue #928, opened 2020-11-19
Update Network.snapshot to work with all addresses


qstokkink:
The `Network.snapshot()` method is marked as deprecated due to it only handling IPv4 addresses and (at the time) no generalized address packing code being available.

With the advent of the new `Serializer` format for addresses (`address`) this method can be updated to support all kinds of addresses.

This will, however, break backwards compatibility with the old format. Specifically, this will affect Tribler's `TriblerTunnelCommunity`.

PR #929, opened 2020-11-20
Fix for introducing peers to outdated addresses

egbertbouman:
This PR fixes an issue that causes trackers (and other nodes) to introduce peers to outdated addresses. This is due to outdated peer information getting stuck in the `service_cache`. It's a bit difficult to explain what happens, but it boils down to this:

* Peer A *first* sends an intro-req to Peer B for community X and *then* for community Y.
* Peer A updates its address.
* Peer A sends another intro-req to Peer B for communities X & Y.

Now in community X, peer B will use the correct addresses when sending intro-responses. However, in community Y peer B will still use the old address.

I've intentionally used the same instance in the cache as the one in `verified_peers`, to prevent future issues (like losing IPv6 addresses).

### DIFF
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -76,7 +76,11 @@ class Network(object):
                 self.services_per_peer[peer.mid] |= set(services)
             for service in services:
                 service_cache = self.reverse_service_lookup.get(service, [])
-                self.reverse_service_lookup[service] = list(set(service_cache + [peer]))
+                # Ensure that the peer instance in the cache is the same one as in verified_peers.
+                if peer in service_cache:
+                    service_cache.remove(peer)
+                service_cache.append(self.get_verified_by_public_key_bin(peer.public_key.key_to_bin()) or peer)
+                self.reverse_service_lookup[service] = service_cache
 
     def add_verified_peer(self, peer):
         """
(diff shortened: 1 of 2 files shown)


qstokkink:
> It's a bit difficult to explain what happens

No need for further explanation, I understand what is happening 👍 Your PR also inspired me to create #930.

egbertbouman:
I was just reading the issue. I agree it can be dangerous to work with multiple peer instances. Personally, I think it's best to not create duplicates at all if possible (assuming there is no significant negative performance impact). This should be easy to implement since we just have to reuse the peers that are already in `Network`.

qstokkink:
> I think it's best to not create duplicates at all if possible

Completely agree, but:

> assuming there is no significant negative performance impact

You hit the nail on the head there. I avoided this like the plague before, as it is implemented like this:

https://github.com/Tribler/py-ipv8/blob/e378ff1389d32625d23cf4286dfa1fd69d09041a/ipv8/peerdiscovery/network.py#L198-L208

That said, most of the design of `Network` is 4 years old. We can improve many of these things now, `verified_peers` is now a set for instance -- this should allow for more efficient lookups.

Concluding, also in light of #928, it is about time the `Network` gets some TLC. I'll probably take care of some of these things (performance, readability, typing, namedtuples, etc.) after the IPv6 stuff is merged, because I don't want to screw with anything that might get in the way of that.

qstokkink (approved):

PR #931, opened 2020-11-23
Added IPv6 support to DHTCommunity

egbertbouman:
This PR partially addresses https://github.com/Tribler/py-ipv8/issues/615 and adds IPv6 support to the `DHTCommunity`.

IPv6 support works as follows: 
All nodes in the network are identified by an ID that is partially based on the IP address that is used while sending DHT messages. Since we can now receive messages over multiple interfaces, I've added a `RoutingTable` and `Storage` object for each interface. Which of these objects is used while processing a query is decided based on the query's source address. This ensures that all nodes in the `RoutingTable` will be of the same address type, which also means that a response with nodes can only be of the same address type as was used when sending the request. When `find_values` or `find_nodes` is called, the `DHTCommunity` will perform a crawl over all available endpoints.


qstokkink:
validate

qstokkink (approved):

PR #932, opened 2020-11-25
Add overlay and walk strategy loading from a function

drew2a:
This PR:

Adds the ability to load overlay and walk strategy from "function class representation" to have both explicit imports and lazy loading:

``` python
def GigaChannelCommunity():
    from tribler_core.modules.metadata_store.community.gigachannel_community import GigaChannelCommunity
    return GigaChannelCommunity

 ...

@overlay(GigaChannelCommunity)
class GigaChannelCommunityLauncher(IPv8CommunityLauncher):
    pass
```
 
Reference: https://github.com/Tribler/tribler/pull/5746

### DIFF
--- a/ipv8/loader.py
+++ b/ipv8/loader.py
@@ -5,6 +5,7 @@ Use if you have a IPv8 manager (a "session") that needs fine-grained control ove
 """
 
 import logging
+import types
 
 from .keyvault.crypto import default_eccrypto
 from .peer import Peer
@@ -100,6 +101,17 @@ def precondition(str_condition):
     return decorator
 
 
+def _get_class(class_or_function):
+    """
+    Return the class by processing incoming argument either as a function or as
+    a class.
+
+    :param class_or_function: the class, or the function that represents this class.
+    """
+    return class_or_function() if isinstance(class_or_function, types.FunctionType) \
+        else class_or_function
+
+
 def overlay(str_module_or_class, str_definition=None):
     """
     Specify, as strings, a module and Community class object defined therein to lazy-load.
@@ -118,6 +130,14 @@ def overlay(str_module_or_class, str_definition=None):
         class A(CommunityLauncher):
            ...
 
+        def my_community_class():
+            from my_module.some_submodule import MyCommunityClass
+            return MyCommunityClass
+
+        @overlay(my_community_class)
+        class A(CommunityLauncher):
+           ...
+
     :param str_module_or_class: either the module to load or a Community class.
     :param str_definition: either the class definition to load or None if str_module_or_class is not a string.
     """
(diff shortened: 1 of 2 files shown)


qstokkink:
ok to test

ipv8/loader.py
```diff
@@ -118,6 +119,15 @@ class A(CommunityLauncher):
...
            ...
 
+        def MyCommunityClass():
+            from my_module.some_submodule import MyCommunityClass
+            return MyCommunityClass
+           ...
+
+        @overlay(MyCommunityClass)
+        class A(CommunityLauncher):
+            pass
```
qstokkink, line 129: This new example redefines `MyCommunityClass` and uses `...` differently. This issue is repeated at line 173. Please use something close (or equal) to:

```suggestion
        def my_community_class():
            from my_module.some_submodule import MyCommunityClass
            return MyCommunityClass

        @overlay(my_community_class)
        class A(CommunityLauncher):
            ...
```

ipv8/util.py
```diff
@@ -38,3 +40,8 @@ async def call_async(*args, **kwargs):
 
 def strip_sha1_padding(s: bytes):
     return s[12:] if s.startswith(b'SHA-1\x00\x00\x00\x00\x00\x00\x00') else s
+
+
+def get_class(class_or_function):
```
qstokkink, line 45: This function is not used outside of `loader.py`, please relocate it to that file.

qstokkink (changes requested):
Thanks for your contribution. Please fix the three issues with this PR so it can be merged.

ipv8/util.py
drew2a: Relocated :)

ipv8/loader.py
drew2a: 👌

ipv8/loader.py
```diff
@@ -160,6 +181,15 @@ class A(CommunityLauncher):
...
            ...
 
+       def my_strategy_class():
+            from my_module.some_submodule import MyStrategyClass
+            return MyStrategyClass
+           ...
+
+        @walk_strategy(my_strategy_class)
+        class A(CommunityLauncher):
+            pass
```
qstokkink, line 191: Looking good 👍 , please don't forget the second half of the request for changes: consistent use of the `...`:

```suggestion
       def my_strategy_class():
            from my_module.some_submodule import MyStrategyClass
            return MyStrategyClass

        @walk_strategy(my_strategy_class)
        class A(CommunityLauncher):
            ...
```

After that, please squash your commits (please contact me if you need help and I'll give you the appropriate git commands - this should take about 30 seconds).

Then, finally, this should be completely done.

qstokkink (changes requested):

drew2a:
@qstokkink looks good now)

ipv8/loader.py
```diff
@@ -118,6 +130,15 @@ class A(CommunityLauncher):
...
            ...
 
+        def my_community_class():
+            from my_module.some_submodule import MyCommunityClass
+            return MyCommunityClass
+
+        @overlay(my_community_class)
+        class A(CommunityLauncher):
+            pass
+           ...
```
qstokkink, line 140: The issue at line 190 was resolved, please make sure to tackle this one too.

qstokkink (approved):
Beautiful, thanks for addressing all the change requests 👍

PR #938, opened 2020-12-04
Fixed new style punctures sent to old style peers

### LINKED ISSUE #937: Old-style peers getting sent new-style punctures
On the following line we create a puncture request which we send to the peer being introduced in an introduction response. This causes the introduced peer to send a puncture to the peer we are receiving the introduction request from.

https://github.com/Tribler/py-ipv8/blob/d56389f190c364e68e21c5e4ec4d70878f11efdb/ipv8/community.py#L291-L292

If the peer we received the introduction request from is old style and the introduced peer is new style, this will cause the new style introduction to send a new style puncture to the old style peer that sent the introduction request. This code should read as follows (confirmed fix):

```python
packet = self.create_puncture_request(lan_socket_address, socket_address, identifier, prefix=prefix, 
                                      new_style=new_style) 
```

qstokkink:
Fixes #937

This PR:

 - Fixes introduction responses suggesting to new-style introduced peers that the old-style peer they need to puncture is new-style based on the introduced peer's new-style setting instead of the setting of the peer they need to puncture.

### DIFF
--- a/ipv8/community.py
+++ b/ipv8/community.py
@@ -289,7 +289,7 @@ class Community(EZPackOverlay):
 
         if introduced:
             packet = self.create_puncture_request(lan_socket_address, socket_address, identifier, prefix=prefix,
-                                                  new_style=new_style_intro)
+                                                  new_style=new_style)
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #942, opened 2020-12-10
Network class polishing

### LINKED ISSUE #935: Optimize Network set lookups
We have several instances of lookups by set hash in the `Network` class. These all look like this example:

https://github.com/Tribler/py-ipv8/blob/e060a37a0ddad5f3c061305bb56ebb107c494340/ipv8/peerdiscovery/network.py#L202-L212

Even though for-loops are fast, we can (and probably should) optimize these lookups. Note that `set` objects don't have a `find()` or `get()` method. I'll document the three fastest lookups I found here.

---

Each experiment uses 100000 repetitions with a set of 300 elements (all integers in `[0, 300)`). All code fragments operate of a given set (`container`), to find an object (`search`) that has a field (`a`, `search.a = randint(0, 299)` for each repetition) to match to the field of the set elements, if no match is found then the result (`found`) should be `None`.

### for loop (current)
```python
found = None
for o in container:
    if o.a == search.a:
        found = o
        break
```
<table>
<tr><td>Total time</td><td>1.6288s</td></tr>
<tr><td>Time per search</td><td>16.288μs</td></tr>
<tr><td>Speedup vs. for-loop</td><td>1.0000</td></tr>
</table>

### set operators
```python
found = container - (container ^ {search})
found = found.pop() if found else None
```
<table>
<tr><td>Total time</td><td>0.7954s</td></tr>
<tr><td>Time per search</td><td>7.954μs</td></tr>
<tr><td>Speedup vs. for-loop</td><td>2.0478</td></tr>
</table>

*Note: `{search} & container` or `container & {search}` will not work. Python will give you the element from the smallest set back - our "fake" search object.*

### map hijack
```python
def find(x):
    if x == search:
        find.out = x
find.out = None
map(find, container)
found = find.out
```
<table>
<tr><td>Total time</td><td>0.0517s</td></tr>
<tr><td>Time per search</td><td>0.517μs</td></tr>
<tr><td>Speedup vs. for-loop</td><td>31.5023</td></tr>
</table>

---

So far, the "map hijack" is the fastest. If you have a faster lookup method, please let me know 😃

qstokkink:
Fixes #935

This PR:

 - Adds automatic resolution of known peers in `lazy_community.py` `lazy_wrapper*` decorators.
 - Adds typing information to `Network`.
 - Adds `verified_by_public_key_bin` to `Network` for quicker lookups per public key.
 - Updates `Network` attribute comments to be attribute docstrings (you can now press `Ctrl+Q` on them 😃).
 - Updates `Network.services_per_peer` to use the full public key material instead of an mid.
 - Updates `TestNetwork` to use `bytes` as overlay ids.

Notable deviation from proposed fix in #935: decided not to go with best performance, but simple one line lookups (thanks @kozlovsky). This class is cryptic and fagile enough, we don't need even more mystical constructs.

### DIFF
--- a/ipv8/lazy_community.py
+++ b/ipv8/lazy_community.py
@@ -83,7 +83,9 @@ def lazy_wrapper(*payloads: Type[Payload]):
                 raise PacketDecodingError("Incoming packet %s has an invalid signature" %
                                           str([payload_class.__name__ for payload_class in payloads]))
             # PRODUCE
-            return func(self, Peer(auth.public_key_bin, source_address), *unpacked)
+            peer = (self.network.verified_by_public_key_bin.get(auth.public_key_bin)
+                    or Peer(auth.public_key_bin, source_address))
+            return func(self, peer, *unpacked)
         return wrapper
     return decorator
 
@@ -117,7 +119,9 @@ def lazy_wrapper_wd(*payloads: Type[Payload]):
                                           str([payload_class.__name__ for payload_class in payloads]))
             # PRODUCE
             output = unpacked + [data]
-            return func(self, Peer(auth.public_key_bin, source_address), *output)
+            peer = (self.network.verified_by_public_key_bin.get(auth.public_key_bin)
+                    or Peer(auth.public_key_bin, source_address))
+            return func(self, peer, *output)
         return wrapper
     return decorator
 
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -1,52 +1,83 @@
 import logging
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from threading import RLock
+from typing import Dict, Iterable, List, Optional, Set
 
 from ..messaging.serialization import default_serializer
+from ..types import Address, Overlay, Peer
+
+
+MID = bytes
+PublicKeyMat = bytes
+Service = bytes
+ServiceSet = Set[Service]
+WalkableAddress = namedtuple('WalkableAddress', ['introduced_by', 'services', 'new_style'])
 
 
 class Network(object):
 
     def __init__(self):
-        # All known IP:port addresses, mapped to (introduction peer, services, new_style)
-        self._all_addresses = {}
-        # All verified Peer objects (Peer.address must be in _all_addresses)
-        self.verified_peers = set()
+        self._all_addresses: Dict[Address, WalkableAddress] = {}
+        '''All known IP:port addresses, mapped to (introduction peer, services, new_style).'''
+
+        self.verified_peers: Set[Peer] = set()
+        '''All verified Peer objects (Peer.address must be in _all_addresses).'''
+
+        self.verified_by_public_key_bin: Dict[PublicKeyMat, Peer] = {}
+        '''Map of known public keys to Peer objects.'''
+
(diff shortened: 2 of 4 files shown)


ipv8/lazy_community.py
```diff
@@ -35,7 +35,9 @@ def wrapper(self, source_address, data):
                 raise PacketDecodingError("Incoming packet %s has an invalid signature" %
                                           str([payload_class.__name__ for payload_class in payloads]))
             # PRODUCE
-            return func(self, Peer(auth.public_key_bin, source_address), *unpacked)
+            peer = self.network.verified_by_public_key_bin.get(auth.public_key_bin,
+                                                               Peer(auth.public_key_bin, source_address))
```
kozlovsky, line 39: The creation of a Peer object is not exactly cheap, and it happens on each lookup. If the speed is important here, it is better to create a Peer instance after the unsuccessful lookup only:
```python
peer = self.network.verified_by_public_key_bin.get(auth.public_key_bin) or \
       Peer(auth.public_key_bin, source_address))
```
or
```python
peer = self.network.verified_by_public_key_bin.get(auth.public_key_bin)
if peer is None:
    peer = Peer(auth.public_key_bin, source_address))
```

ipv8/peerdiscovery/network.py
```diff
@@ -1,140 +1,175 @@
...
-    def is_new_style(self, address):
-        return self._all_addresses.get(address, ('', None, False))[2]
+    def is_new_style(self, address: Address) -> bool:
+        """
+        Check if an address supports new-style introduction requests and responses.
+
+        :param address: the address to check for.
+        :returns: True iff the address is both known and known to support new-style introductions.
+        """
+        return self._all_addresses.get(address, WalkableAddress(b'', None, False)).new_style
```
kozlovsky, line 64: The same here, it is possible to avoid the creation of WlakableAddress if it is already exists:
```python
return (self._all_addresses.get(address) or WalkableAddress(b'', None, False)).new_style
```

ipv8/lazy_community.py
qstokkink: Thanks 👍, good point: I'll change that.

qstokkink:
validate

ipv8/peerdiscovery/network.py
```diff
@@ -1,141 +1,175 @@
...
+
+        self.services_per_peer: Dict[PublicKeyMat, ServiceSet] = {}
+        '''Map of advertised services (set) per peer'''
+
+        self.service_overlays: Dict[Service, Overlay] = {}
+        '''Map of service identifiers to local overlays'''
+
         self.reverse_ip_cache_size = 500
+        self.reverse_ip_lookup = OrderedDict()
+        '''Cache of IP:port -> Peer. This is not a normal dict, as the amount of addresses can grow infinitely!'''
```
kozlovsky, line 46: The comment looks a bit unclear. `OrderedDict` does not provide some infinite memory to allow infinite growth :)
Probably it means that the old entries will be removed from the dict using `popitem(False)` which are missing from the usual dict?

qstokkink: Reworded, should be a bit more clear now.

qstokkink:
validate

kozlovsky (approved):

PR #944, opened 2020-12-11
Remove UDPIPv6 from default endpoints

### LINKED ISSUE #943: Remove IPv6 default setting
The default endpoint settings of IPv8 are currently as follows:

https://github.com/Tribler/py-ipv8/blob/37b19238443dcdae426f67d6e9013027c6f248a8/ipv8_service.py#L72-L77

These settings were adopted optimistically in anticipation of a #615 resolution. However, IPv8 will crash if used with these default settings. In order to allow an intermediate release (before IPv8 v3.0) this should be reverted to:

```python
self.endpoint = DispatcherEndpoint(["UDPIPv4"], 
                                    UDPIPv4={'port': configuration['port'], 
                                             'ip': configuration['address']}) 
```

qstokkink:
Fixes #943

This PR:

 - Removes `UDPIPv6` from the default endpoints loaded by `IPv8`.

### DIFF
--- a/ipv8_service.py
+++ b/ipv8_service.py
@@ -69,12 +69,9 @@ else:
             if endpoint_override:
                 self.endpoint = endpoint_override
             else:
-                self.endpoint = DispatcherEndpoint(["UDPIPv6", "UDPIPv4"],
+                self.endpoint = DispatcherEndpoint(["UDPIPv4"],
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #954, opened 2021-01-21
Updated tracker script

egbertbouman:
This PR:
* Updates the tracker script such that it will stop gossiping peers in a particular community if they have reached the `max_peers` limit. There is no change in protocol and works by letting the tracker send `introduction-requests` to peers. If a peer doesn't respond with an `introduction-response` or the `peer_limit_reached` bit is set, the tracker will no longer introduce that particular peer to others within the community in which the original `introduction-request` was sent.
* Updates the tracker script to keep a larger cache of peers that are looking to be bootstrapped. The current drop time of 120s is replaced by a keep-alive mechanism (as discussed in the first point) that periodically checks if the peer is still responding.
* Adds a test that periodically queries live trackers and plots the received introduction address information.
* Adds a simple API server for debugging purposes that returns `{service_id: [peers]}` (disabled by default, and optionally protected with an API key and TLS).
* Fixes an issue in `Network` that allowed for multiple `Peer` instances to exist with the same public key.
* Adds some docs on the basics of IPv8 peer discovery.


qstokkink:
validate

qstokkink (approved):
Looks good 👍 

I apparently broke the validation tests quite badly in my PR, please ignore it (moved to #956).

PR #955, opened 2021-01-26
Modularized Community bootstrapping

### LINKED ISSUE #946: Add external bootstrap server inject script
We can use bootstrap servers from external parties in our stack. We already have lazy resolution of DNS names, external mechanisms can use the exact same method.

Note that external bootstrap servers have different protocols, so these will have to be modularized.

Proposed design:

 - A bootstrap script per external protocol in a (new) `ipv8.bootstrap` module.
 - Add a bootstrap manager in this (new) `ipv8.bootstrap` module.
 - Extend `Community.bootstrap()` to call on the (new) bootstrap manager with a `service_id` (the community prefix).
 - Add bootstrap module configuration to the `/ipv8/configuration.py` file.

Possible hurdles:

 - Bootstrapped peers need to reuse the port they received from the external bootstrap server or face bouncing off the NAT or firewall layers. Without resolution, this means the same port will serve two protocols (which can lead to nasty errors).
 - [Thanks @devos50] These capabilities most likely require additional libraries/dependencies to be added. We want each module to have conditional imports.

The interface custom scripts should fill, would be the following:

```python
class Bootstrapper(ABC):
    """
    A script to connect to external bootstrapping resources, using external hardware.

    When not to crash:
     - Loading this module file when missing dependencies.
     - Initializing a subclass when the service is unavailable (return ``False``).
     - Failing to retrieve peers for a service (return an empty iterable, e.g. ``[]``).

    When to crash:
     - Initializing this module subclass when missing dependencies.
    """

    @abstractmethod
    def initialize(self, overlay: Overlay) -> bool:
        """
        Start connecting to this bootstrapping service. Don't perform any network traffic in ``__init__``!

        :param overlay: the network overlay to initialize for.
        :returns: whether the initialization was successful.
        """
        pass

    @abstractmethod
    async def get_addresses(self, overlay: Overlay, timeout: float) -> Iterable[Address]:
        """
        Return some IPv8 addresses (if available) from this bootstrapping service.
        These addresses should be walkable (not blocked by a NAT or firewall).

        :param overlay: the network overlay to get peers for.
        :param timeout: the maximum time we wish to wait until we get any result (i.e. an empty list).
        :returns: the addresses for the given service_id.
        """
        pass

    @abstractmethod
    def keep_alive(self, overlay: Overlay) -> None:
        """
        Periodically called to keep this bootstrap connection alive.

        :param overlay: the network overlay to keep alive.
        """
        pass

    @abstractmethod
    def blacklist(self) -> Iterable[Address]:
        """
        Returns the blacklisted addresses for this Bootstrapper.
        """
        pass
```

qstokkink:
Fixes #946

This PR:

 - Adds a `Bootstrapper` interface for generic bootstrapping.
 - Adds a `DispersyBootstrapper` implementation for the default IPv8 bootstrapping.
 - Adds bootstrapping configuration options to the overlay configuration entries.
 - Adds a `bootstrapper` definition for `loader.py` loading.
 - Removes hard-coded bootstrapping logic from `Community`.

Notable changes:
 - To inspect the IPv8 bootstrap ips use `DISPERSY_BOOTSTRAPPER['init']['ip_addresses']` instead of `_DEFAULT_ADDRESSES`.
 - `Community` instances start without any bootstrapping implementation, cleaning up many tests, Tribler and Gumby.

This PR requires changes in Tribler and Gumby (I'll take care of these).

### DIFF
--- a/ipv8/community.py
+++ b/ipv8/community.py
@@ -7,13 +7,11 @@ Community instance.
 @organization: Technical University Delft
 @contact: dispersy@frayja.com
 """
-import logging
 import sys
 from asyncio import ensure_future, iscoroutine
 from binascii import hexlify
+from functools import partial
 from random import choice, random
-from socket import error, gethostbyname
-from threading import Thread
 from time import time
 from traceback import format_exception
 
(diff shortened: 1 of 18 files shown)


qstokkink:
Ready for review. The remaining Pylint error is due to the import magic in `ipv8_service.py`.

ipv8/community.py
```diff
@@ -180,40 +139,23 @@ def on_deprecated_message(self, source_address, data):
...
-    def _ensure_blacklisted(self, address):
+        if not self._bootstrappers_initialized:
+            self.bootstrappers = [bootstrapper for bootstrapper in self.bootstrappers if bootstrapper.initialize(self)]
+            self._bootstrappers_initialized = True
+        for bootstrapper in self.bootstrappers:
+            task = self.register_anonymous_task(f"bootstrap {repr(bootstrapper)}", bootstrapper.get_addresses,
+                                                self, 60.0)
+            task.add_done_callback(self.received_bootstrapped_addresses)
+
+    def received_bootstrapped_addresses(self, addresses):
```
egbertbouman, line 150: I would be surprised if this is working. If I remember correctly, the done callback takes a `Future` as an argument. Casting a finished `Future` to a `list` always seems to yield an empty `list`.

ipv8/community.py
```diff
@@ -180,40 +139,23 @@ def on_deprecated_message(self, source_address, data):
...
-            return
-        self.logger.debug("Bootstrapping %s, current peers %d", self.__class__.__name__, len(self.get_peers()))
-        self.last_bootstrap = time()
-        for socket_address in _DEFAULT_ADDRESSES:
-            self._ensure_blacklisted(socket_address)
-            self.walk_to(socket_address)
-
-    def _ensure_blacklisted(self, address):
+        if not self._bootstrappers_initialized:
+            self.bootstrappers = [bootstrapper for bootstrapper in self.bootstrappers if bootstrapper.initialize(self)]
```
egbertbouman, line 143: It may be useful to have `initialize` async as well. If this can be done cleanly.

ipv8/community.py
qstokkink: Woops, that should indeed be `addressses.result()`.

ipv8/community.py
qstokkink: Implemented 👍

qstokkink:
@egbertbouman I implemented your suggestion and fix, please review again.

Tribler is confirmed to still work (with a patch).

egbertbouman (approved):
Looks good!

PR #960, opened 2021-01-28
Added DiscoveryStrategies to OverlaysEndpoint

egbertbouman:
This PR:

* Adds a list of `DiscoveryStrategies` for each overlay in `OverlaysEndpoint`. This could be useful for debugging.
* Fixes a broken image in the peer discovery docs (attempt 2 :face_with_head_bandage:).
* Removed link to crawler_service_tutorial.rst from index.rst


qstokkink (approved):
Could you use this `overlay.max_peers` information to finally take out the peer hard-coding from [the Tribler debug tab](https://github.com/Tribler/tribler/blob/devel/src/tribler-gui/tribler_gui/debug_window.py#L332-L342)?

egbertbouman:
@qstokkink That would be nice. Currently, all communities just seem to use the default of 30. So, we may want to change the value of `max_peers`  for certain communities. I don't think that should be a problem, but I'll have to test it to be sure.

PR #964, opened 2021-02-09
Changed the name of CommunityLauncher

egbertbouman:
Right now when calling `CommunityLoader.set_launcher` IPv8 imports the community (which happens when calling `CommunityLauncher.get_name`). However, since the `precondition` can fail, this import may be unneeded. This requires us to have the dependencies of the community installed, even if it's disabled.

As a workaround, I suggest changing `CommunityLauncher.get_name` to simply return the name of the launcher itself, rather than that of the community.

### DIFF
--- a/ipv8/loader.py
+++ b/ipv8/loader.py
@@ -343,7 +343,7 @@ class CommunityLauncher:
 
         :rtype: str
         """
-        return self.get_overlay_class().__name__
+        return self.__class__.__name__
 
     def not_before(self):
         """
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

qstokkink (approved):
Looks like a good resolution.

Please merge when you're prepared to make the required changes to Tribler and Gumby*.

*The `@after`/`not_before()` entries will have to be changed there.

PR #965, opened 2021-02-10
Add TunnelSettings.from_dict classmethod

kozlovsky:
This PR adds `TunnelSettings.from_dict` classmethod. This method is useful, because sometimes it is necessary to adjust settings in constructor of a class inherited from `TunnelCommunity` before passing them to `TunnelCommunity.__init__`, and it is easier to convert settings from dict to `TunnelSettings`  before modifying them:

```python
class SomeTunnelCommunity(TunnelCommunity):
    def __init__(self, *args, **kwargs):
        settings = kwargs.pop('settings', TunnelSettings())
        if isinstance(settings, dict):
            settings = TunnelSettings.from_dict(settings)  # <-- new method
            settings.peer_flags.add(SOME_FLAG)
        super(SomeTunnelCommunity, self).__init__(*args, settings=settings, **kwargs)
```

### DIFF
--- a/ipv8/messaging/anonymization/community.py
+++ b/ipv8/messaging/anonymization/community.py
@@ -81,6 +81,13 @@ class TunnelSettings(object):
         # Maximum number of relay_early cells that are allowed to pass a relay.
         self.max_relay_early = 8
 
+    @classmethod
+    def from_dict(cls, d):
+        result = cls()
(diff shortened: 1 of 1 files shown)


egbertbouman (approved):
Looks good!

qstokkink (approved):

PR #967, opened 2021-02-23
Update hidden services peer discovery

egbertbouman:
This PR fixes various issues with the `PexCommunity` and peer discovery in general. Also, it contains several tweaks that should increase the success rate of creating end-to-end circuits. And finally, I added some additional debug information to the tunnels and overlays endpoints.

### DIFF
--- a/ipv8/community.py
+++ b/ipv8/community.py
@@ -254,7 +254,7 @@ class Community(EZPackOverlay):
         if introduced:
             packet = self.create_puncture_request(lan_socket_address, socket_address, identifier, prefix=prefix,
                                                   new_style=new_style)
-            self.endpoint.send(introduction_wan if introduction_lan == ("0.0.0.0", 0) else introduction_lan, packet)
+            self.endpoint.send(introduction.address, packet)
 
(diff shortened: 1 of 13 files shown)


ipv8/community.py
```diff
@@ -254,7 +254,7 @@ def create_introduction_response(self, lan_socket_address, socket_address, ident
         if introduced:
             packet = self.create_puncture_request(lan_socket_address, socket_address, identifier, prefix=prefix,
                                                   new_style=new_style)
-            self.endpoint.send(introduction_wan if introduction_lan == ("0.0.0.0", 0) else introduction_lan, packet)
+            self.endpoint.send(introduction.address, packet)
```
devos50, line 257: Would this fix https://github.com/Tribler/anydex-core/issues/64? 🤔

egbertbouman: Yes, it should. That's how I discovered the bug.

devos50: Great, thanks! I was unable to find out where it originates from so I just disabled that test 😂

qstokkink:
validate

qstokkink (approved):

PR #968, opened 2021-02-26
Fixed estimated WAN address being set to a LAN address

egbertbouman:
Right now it's possible for any peer to overwrite the WAN address of a `Community` with a LAN address. It just needs to set the `IntroductionResponse.destination_address` to a LAN address that could be part of our LAN. (That's easy to do since we're gossiping our own LAN address.) This problem shouldn't occur when `netifaces` is not installed, as `address_is_lan` works differently without `netifaces`.

If the walker continues to walk, this problem would resolve itself right after the next introduction-response. But if the walker stops that address would remain in `my_estimated_wan`. That's the problem that I'm seeing on our exit nodes: eventually, most nodes seem to start putting LAN addresses on the DHT. Maybe some students are playing with our protocol.

Admittedly, a similar problem would arise with WAN addresses. That's why we used to have WAN voting. I think we should consider re-introducing some of this logic.

### DIFF
--- a/ipv8/community.py
+++ b/ipv8/community.py
@@ -344,7 +344,7 @@ class Community(EZPackOverlay):
 
     def on_introduction_response(self, peer, dist, payload):
         if (isinstance(payload.destination_address, UDPv4Address)
-                and not self.address_is_lan(payload.destination_address[0])):
+                and not self.address_in_lan_subnets(payload.destination_address[0])):
             self.my_estimated_wan = payload.destination_address
         self.my_peer.address = payload.destination_address
 
--- a/ipv8/messaging/interfaces/endpoint.py
+++ b/ipv8/messaging/interfaces/endpoint.py
@@ -251,15 +251,17 @@ class EndpointListener(metaclass=abc.ABCMeta):
         """
         Checks if the given ip address is either our own address or in one of the subnet defined for local network usage
         :param address: ip v4 address to be checked
-        :return: True if the adrress is a lan address, False otherwise
+        :return: True if the address is a lan address, False otherwise
         """
         if address == self.get_lan_address_without_netifaces():
             return True
-        else:
-            lan_subnets = (("192.168.0.0", 16),
-                           ("172.16.0.0", 12),
-                           ("10.0.0.0", 8))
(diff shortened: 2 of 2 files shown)


synctext:
> Right now it's possible for any peer to overwrite the WAN address

So a _remote IPv4 alteration_ attack. Sounds bad. Thank you for diving into this at that late hour. Wild idea...As exit nodes are running on fixed IPv4 addresses, feel free to make this a hard-coded WAN. Internet voting on external WAN should not be able to overrule cmdline input I guess.

qstokkink:
The idea behind removing the voting was that, regardless of what a (possibly malicious or erroneous) majority thinks and  regardless of the NAT configuration, as you flip-flop between guesses for your WAN address _eventually_ you get the "right" WAN address and perform some connection step on some subnet you are connectable on. Therefore, I'd much prefer your solution to self-correct by walking over reintroducing voting.

That said, addresses should be treated as ephemeral in the first place: ideally no overlay should depend on any form of address, only the public key. Depending on IPv4 will only get worse if we also have Bluetooth, IPv6, etc. (which is actually quite a realistic concern for the Kotlin port).

Lastly: at some point we should also set up some physical complex NAT setup to test all of this on. Perhaps I am completely wrong about the flip-flopping approach. At this point its mostly speculation and no hard measurements.

egbertbouman:
@qstokkink I agree that self-correcting is the preferred solution. However, the problem with self-correcting right now is that we eventually stop walking (until peers go offline). Personally, I think if `my_estimated_wan` is there, we should do what we can to make sure it's correct. Otherwise, there is no point in having it in the first place.

Regarding your remark about not using addresses. Let me put that in the "future work" pile for tunnels/DHT :wink:

Anyway, this PR fixes my issue with the exit nodes. So, if you prefer to leave it as-is, that's fine by me.

qstokkink:
Agreed and, until proven faulty/wrong, I'd indeed prefer to leave it as-is.

> Regarding your remark about not using addresses. Let me put that in the "future work" pile for tunnels/DHT 😉

Ultimately, I think it will depend on the Kotlin/superapp implementation how pressing this concern is. It might be fine to push this into the far future. We'll have to see how that develops.

qstokkink (approved):

PR #973, opened 2021-03-27
Updated IRMA enroll script

qstokkink:
This PR:

 - Adds the current Nijmegen IRMA public key.
 - Fixes the `enroll_script.py` to be compatible with the latest version of the Nijmegen public key and IRMA version.
 - Fixes some Python 2 -> 3 port issues with the `enroll_script.py`.

For now this is all I feel like fixing. Perhaps in a future PR I'll add the keys to the `default_identity_formats.py`.

Note: because you need a Dutch DigiD to test this (you're welcome to do so) I can't really make any automated unit tests for this.


qstokkink:
The remaining Pylint errors are mostly due to the fact that Pylint doesn't know about `PyQtWebEngine` (and I firmly believe we shouldn't install that on our test server).

egbertbouman (approved):

PR #975, opened 2021-04-03
Added UDP broadcast bootstrapper

### LINKED ISSUE #972: Add broadcast-based LAN bootstrapping
The following documents a prototype to discover processes using UDP broadcasts which we can inflate to be a full-fledged IPv8 bootstrapper. Works most of the time (not always!). Sometimes the setup works but it can only be performed unidirectionally. However, most importantly, this approach does not require elevated privileges.

In order to transform this into a bootstrapper, all processes/machines should both listen and send at the same time (using non-blocking receives). Furthermore, this should also filter out IPv8 traffic (e.g., my printer also decided to respond to this traffic). Lastly, there should be some sane way to limit traffic once peers have been found (instead of continuously spamming 65535 ports all the time using broadcasts).

### Setup (both processes)

```python
from socket import *
s = socket(AF_INET, SOCK_DGRAM)
s.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
s.setsockopt(SOL_SOCKET, SO_BROADCAST, 1)
s.bind(('', 0))
```

### Listener

```python
m, addr = s.recvfrom(1024)  #  Sender managed to successfully contact us
print(m, addr)
s.sendto(b'a response!', addr)  #  See if the connection is bidirectional
```

### Sender

```python
# Try to find the listener (fire and forget):
for p in range(65535):
    try:
        if p != s.getsockname()[1]:  # Avoid sending to self for testing (remove check in final product)
            s.sendto(b'Please send me a response!', ('255.255.255.255', p))
    except OSError:
        pass  # Windows doesn't really care, Ubuntu throws an exception on "illegal" ports
m, addr = s.recvfrom(1024) #  Listener managed to respond to us: puncture success!
print(m, addr)
```

qstokkink:
Fixes #972 (this PR follows the approach presented in this issue)

This PR:

 - Adds a bootstrapper based on UDP broadcasts.
 - Adds `unload()` to bootstrapper interface.

This is my test script (which I run on separate machines):

```python
import os
from asyncio import ensure_future, get_event_loop

from pyipv8.ipv8.community import Community
from pyipv8.ipv8.configuration import Bootstrapper, BootstrapperDefinition, ConfigBuilder, Strategy, WalkerDefinition
from pyipv8.ipv8_service import IPv8


class MyCommunity(Community):
    community_id = b'testudpbroadcast\x00\x00\x00\x00'

    def started(self):
        async def print_peers():
            print("I am:", self.my_peer, "\nI know:", [str(p) for p in self.get_peers()])

        self.register_task("print_peers", print_peers, interval=5.0, delay=0)


async def start_communities():
    for i in [1]:
        builder = ConfigBuilder().clear_keys().clear_overlays()
        builder.add_key("my peer", "medium", f"ec{i}.pem")
        builder.add_overlay("MyCommunity",
                            "my peer",
                            [WalkerDefinition(Strategy.RandomWalk, 10, {'timeout': 3.0})],
                            [BootstrapperDefinition(Bootstrapper.UDPBroadcastBootstrapper, {})],
                            {},
                            [('started',)])
        await IPv8(builder.finalize(), extra_communities={'MyCommunity': MyCommunity}).start()


ensure_future(start_communities())
get_event_loop().run_forever()
```

### DIFF
--- a/ipv8/bootstrapping/bootstrapper_interface.py
+++ b/ipv8/bootstrapping/bootstrapper_interface.py
@@ -53,3 +53,9 @@ class Bootstrapper(ABC):
         """
         Returns the blacklisted addresses for this Bootstrapper.
         """
+
+    @abstractmethod
+    def unload(self) -> None:
+        """
+        Stop and unload all the resources used by this Bootstrapper.
+        """
--- a/ipv8/bootstrapping/dispersy/bootstrapper.py
+++ b/ipv8/bootstrapping/dispersy/bootstrapper.py
@@ -67,3 +67,6 @@ class DispersyBootstrapper(Bootstrapper):
 
     def blacklist(self) -> Iterable[Address]:
         return self.ip_addresses
+
+    def unload(self) -> None:
+        pass
--- a/ipv8/bootstrapping/udpbroadcast/bootstrapper.py
+++ b/ipv8/bootstrapping/udpbroadcast/bootstrapper.py
@@ -1,8 +1,8 @@
 import logging
+from asyncio import BaseTransport, DatagramProtocol, get_event_loop
 from asyncio.futures import Future
 from binascii import hexlify
 from socket import AF_INET, SOCK_DGRAM, SOL_SOCKET, SO_BROADCAST, SO_REUSEADDR, socket
-from threading import Thread
 from typing import Coroutine, Iterable, Optional, Union
 
 from ..bootstrapper_interface import Bootstrapper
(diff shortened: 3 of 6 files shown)


qstokkink:
The remaining Pylint error is due to the creative import system used in `ipv8_service.py`, which cannot be fixed. Ready for review.

qstokkink:
By the way, regarding why this is not enabled by default:

 1. In most cases you prefer the Internet over your local network. The extra bandwidth expenditure is not worth it in most cases, for the off chance that a peer in your local network even exists.
 2. As local peers will be very fast to respond and fill out all the `MAX_PEER` slots before peers from the Internet can respond, partitioning your local network off from the rest of the Internet is very easy. This can be dangerous.

egbertbouman (commented):
I can't find anything obviously wrong with this PR, but I do have some concerns/thoughts:

* With this PR, calls to a community can come from different threads. Perhaps you could consider executing `walk_to`/`on_packet` from the `asyncio` thread (e.g., by removing the extra thread and letting `asyncio` handle the socket management). Just to avoid potential issues.

* Since this PR adds an additional socket, I assume the addresses of this socket will be added to `Peer.addresses`. If `UDPBroadcastBootstrapper` is used in combination with `DispersyBootstrapper`, I imagine IPv8 could end up talking to a certain peer over multiple sockets.

qstokkink:
Good point, threading can be a bit nasty. I'll return this PR to WIP and investigate the most sane way to integrate `asyncio`.

On your second point: the broadcast address doesn't get added as a peer. Because I was a bit lazy (my apologies for this) I didn't supply this PR with an overview image, which should clarify this. 

msgflow

The broadcast sockets are only used to catch the initial broadcast message. Once peers can be added to the overlay, the received and sent data passes the actual overlay sockets. The key here is that `walk_to` does not add peers, allowing for this handoff.

There is, however, a small window in which the estimated lan address of Overlay 1 uses the broadcast port (as step 3 uses the broadcast address as its destination address). In case another overlay happens to use this address, the same `on_packet` forwarding will kick in and self-repair using the correct Overlay 1 socket.

qstokkink:
@egbertbouman I offloaded the socket to a `DatagramProtocol` now, which should resolve the threading issue you raised. Assuming you accept my [ratification of your second point](https://github.com/Tribler/py-ipv8/pull/975#issuecomment-812988588), this PR is ready for review again.

One note on the `DatagramProtocol`: I use `self._socket` instead of `self._transport` for sending to avoid asyncio's QoS magic, which would have the sending take several minutes instead of several milliseconds.

egbertbouman:
> The broadcast sockets are only used to catch the initial broadcast message. Once peers can be added to the overlay, the received and sent data passes the actual overlay sockets. The key here is that `walk_to` does not add peers, allowing for this handoff.

OK, sounds good. I figured that the address given to `on_packet` was that of the broadcast socket, and that `add_verified_peer` would update `Peer.addresses`. But you're right, it's the IPv8 socket 👍

egbertbouman (approved):

PR #988, opened 2021-05-06
Added RequestCache.passthrough()

### LINKED ISSUE #987: Add passthrough for RequestCache
Predominantly in more complex community logic there is a heavier use of caches. Testing these constructions either requires (1) parametization on the Community level (forcing large configuration objects, like the `TunnelSettings`) or (2) patching the timeouts of caches during testing.

### Candidate solution

We can enable a clean solution to the problem by providing an override value for new cache timeouts in the `RequestCache`. When this value is set, instead of adopting the `timeout_delay` from the `NumberCache` subclass, the override value is used when adding the cache. Using appropriate `__enter__` and `__exit__` blocks for such an override value would allow clean use of `with`. The example usage of this solution in a `TestBase` subclass would be:

```python
with self.overlay(0).request_cache.passthrough():
    # All added NumberCaches will be instantly timed out.
    self.overlay(0).some_code_that_adds_a_cache()
# All added NumberCaches will once again use their specified timeout.
self.overlay(0).some_code_that_adds_a_cache()
```

Additionally this could take a parameter to specify a custom timeout (instead of `0.0`):

```python
with self.overlay(0).request_cache.passthrough(timeout=0.1):
    # All added NumberCaches will be timed out after 0.1 seconds.
    self.overlay(0).some_code_that_adds_a_cache()
```

And to make it feature complete (thanks @egbertbouman), this could also filter on which cache class timeouts should be overwritten:

```python
with self.overlay(0).request_cache.passthrough(filter=MyCache):
    # Only added MyCaches will be instantly timed out.
    self.overlay(0).some_code_that_adds_a_cache()
```

qstokkink:
Fixes #987

This PR:

 - Adds the `RequestCache.passthrough()`, as described in #987.
 - Updates `TestRequestCache` to use a common `self.request_cache` to reduce code duplication.

Note that the `passthrough()` does not allow nesting.

### DIFF
--- a/ipv8/requestcache.py
+++ b/ipv8/requestcache.py
@@ -1,8 +1,9 @@
 import logging
 from asyncio import CancelledError, gather
-from contextlib import suppress
+from contextlib import contextmanager, suppress
 from random import random
 from threading import Lock
+from typing import Generator, Iterable, Optional, Type
 
 from .taskmanager import TaskManager
 
@@ -95,6 +96,20 @@ class RequestCache(TaskManager):
         self.lock = Lock()
         self._shutdown = False
 
+        self._timeout_override: Optional[float] = None
+        """
+        If not None, this specifies the timeout to use instead of the one defined in
+        the ``timeout_delay`` of a ``NumberCache``.
+        This is used internally for ``passthrough()``, don't modify this directly!
+        """
+
+        self._timeout_filters: Optional[Iterable[Type[NumberCache]]] = None
+        """
+        If not None, this specifies the ``NumberCache`` (sub)classes to apply the
+        timeout override for.
+        This is used internally for ``passthrough()``, don't modify this directly!
+        """
+
     def add(self, cache):
         """
         Add CACHE into this RequestCache instance.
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

qstokkink:
validate

ipv8/requestcache.py
```diff
@@ -154,6 +179,59 @@ def pop(self, prefix, number):
         self.cancel_pending_task(cache)
         return cache
 
+    @contextmanager
+    def passthrough(self,  # pylint: disable=W1113
```
qstokkink, line 183: I chose to ignore `W1113 (Keyword argument before variable positional arguments list)` here as the Pylint-recommended approach does not allow our required function signature (using `None` as the "no-filter-specified" value). The nastiness here is that the default value for empty variable arguments is an empty tuple, which would be a valid filter to blacklist all classes (whereas `None` whitelists all classes).

The same function signature could be achieved using `functools.singledispatch`, but this would also double the size of this code while making it arguably less readable.
Yet another approach would be to change the function signature to use sentinel objects, which also seems like overkill.

egbertbouman (approved):

PR #995, opened 2021-06-02
Added basic simulation capabilities to IPv8

devos50:
This PR:

 - Adds basic simulation functionality to IPv8, including documentation and a code snippet on how to use this functionality.

### DIFF
--- a/simulation/discrete_loop.py
+++ b/simulation/discrete_loop.py
@@ -1,0 +1,94 @@
+import asyncio
+import heapq
+
+
+class DiscreteLoop(asyncio.AbstractEventLoop):
+    """
+    A discrete asyncio loop that immediately executes incoming tasks without a real-time waiting period.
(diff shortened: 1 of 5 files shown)


devos50:
validate

simulation/discrete_loop.py
```diff
@@ -0,0 +1,91 @@
+import asyncio
+import heapq
+
+
+class DiscreteLoop(asyncio.AbstractEventLoop):
+    """
+    A discrete asyncio loop, using async/await.
```
qstokkink, line 7: How is that different from the normal event loop?

simulation/discrete_loop.py
```diff
@@ -0,0 +1,91 @@
...
+    This loop can, for example, be used during simulations with IPv8.
```
qstokkink, line 8: Only for example? What else.

Also, how do I use this class?

devos50:
validate

qstokkink (approved):

PR #996, opened 2021-06-03
The key when creating a Peer can also be bytes

devos50:
This PR:

 - Updates the type of the `key` attribute passed to the constructor of the `Peer` class.


qstokkink (approved):
Good catch, thanks 👍

PR #998, opened 2021-06-03
Added link latencies to the IPv8 simulator

devos50:
This PR:

 - Adds support for link latencies in the simulator. Specifically, we modified the `SimulationEndpoint` class to achieve this.

### DIFF
--- a/simulation/simulation_endpoint.py
+++ b/simulation/simulation_endpoint.py
@@ -1,3 +1,6 @@
+from asyncio import get_event_loop
+from collections import defaultdict
+
 from ipv8.test.mocking.endpoint import AutoMockEndpoint
 from ipv8.util import succeed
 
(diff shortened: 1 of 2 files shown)


devos50:
validate

simulation/simulation_endpoint.py
```diff
@@ -11,3 +13,16 @@ class SimulationEndpoint(AutoMockEndpoint):
...
+        Return the link latency to a particular destination address, in seconds.
+        """
+        return 0
+
+    async def actual_send(self, socket_address, packet):
+        await sleep(self.get_link_latency(socket_address))
+        super().send(socket_address, packet)
+
+    def send(self, socket_address, packet):
+        ensure_future(self.actual_send(socket_address, packet))
```
qstokkink, line 28: It seems like `actual_send` will not be overwritten in subclasses. If that is the case, please use this instead:

```suggestion
    def send(self, socket_address, packet):
        get_event_loop().call_later(self.get_link_latency(socket_address), super().send,
                                    socket_address, packet)
```

simulation/simulation_endpoint.py
```diff
@@ -11,3 +13,16 @@ class SimulationEndpoint(AutoMockEndpoint):
     async def open(self):
         self._open = True
         return succeed(None)
+
+    def get_link_latency(self, to_address):
+        """
+        Return the link latency to a particular destination address, in seconds.
+        """
+        return 0
```
qstokkink, line 21: This is dead code. Please consider adding something like `self.latencies = defaultdict(int)` and changing this method to return `self.latencies[to_address]`.

qstokkink (changes requested):

simulation/simulation_endpoint.py
devos50: That's a more clean solution indeed, thanks for the suggestion! 👍

devos50:
validate

qstokkink (approved):
Perfect 👍  thanks!

PR #999, opened 2021-06-04
Initialize EndpointListener._local_interfaces

### LINKED ISSUE #997: `_local_interfaces` not initialized in `EndpointListener`
The `_local_interfaces` variable in the `EndpointListener` class is only set if the `_get_lan_address` is called and is not defined in the `__init__` method. Like the `_netifaces_failed` variable, I believe it should be initialized properly in `__init__`.

qstokkink:
Fixes #997

This PR:

 - Fixes `EndpointListener._local_interfaces` not being initialized in the `__init__`.


qstokkink:
validate

devos50 (approved):

devos50:
We're almost at issue/PR 1000! 🎉

PR #1008, opened 2021-06-25
Fixed missing token for TaskManager PyPi project

qstokkink:
This PR:

 - Fixes the `ipv8_taskmanager` publisher using a token only valid for `py-ipv8`.


drew2a (approved):
👍

PR #1015, opened 2021-07-08
Removed old database.py workarounds

### LINKED ISSUE #982: Remove old database.py workarounds
The `database.py` file is **really old** and includes two workarounds we no longer need.

### Workaround for MacOS Sierra

https://github.com/Tribler/py-ipv8/blob/d95b22344b0adac84619c039366fba7eb4f2b043/ipv8/database.py#L16-L35

This was fixed in Python 3.7 - which is our current minimum version. This can be replaced with `import sqlite3 ` and `cursor.execute(statement)` for all instances of `execute_or_script`.

### Python 2 compatibility

https://github.com/Tribler/py-ipv8/blob/d95b22344b0adac84619c039366fba7eb4f2b043/ipv8/database.py#L38-L39

`database_blob()` has now been reduced to `bytes()`: it only adds needless complexity and should be replaced with `bytes()` everywhere.

qstokkink:
Fixes #982

This PR:

 - Removes `database_blob`, which is just an alias for `bytes`.
 - Removes `execute_or_script`, which was only needed for MacOS <= Python 3.6.
 - Removes `import pysqlite2.dbapi2 as sqlite3`, which was only needed for MacOS <= Python 3.6.

### DIFF
--- a/ipv8/REST/attestation_endpoint.py
+++ b/ipv8/REST/attestation_endpoint.py
@@ -108,7 +108,7 @@ class AttestationEndpoint(BaseEndpoint):
         :param keys_to_keep: list of keys to not remove for
         :type keys_to_keep: [str]
         :return: the list of attestation hashes which have been removed
-        :rtype: [database_blob]
+        :rtype: [bytes]
         """
         database = self.identity_overlay.identity_manager.database
         all_identities = database.get_known_identities()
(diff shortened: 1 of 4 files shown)


qstokkink:
validate

devos50 (approved):

PR #1016, opened 2021-07-08
Allow custom my_peer instead of key curve in MockIPv8

qstokkink:
This PR:

 - Adds support for supplying a custom `Peer` as the `my_peer` of `MockIPv8`.
 - Fixes `MockIPv8` taking variable args after keyword args.


qstokkink:
validate

devos50 (approved):

PR #1019, opened 2021-07-18
Fixed crashing unitialized endpoint on shutdown

ichorid:
If `IdentityEndpoint`'s  `initialize` method was not called prior to shutting down IPv8, it will crash on calling `on_shutdown` callback trying to reference `shutdown()` method of `self.communication_manager` which is `None`.

This little change fixes that, helping with IPv8 testing.

### DIFF
--- a/ipv8/REST/identity_endpoint.py
+++ b/ipv8/REST/identity_endpoint.py
@@ -603,4 +603,5 @@ class IdentityEndpoint(BaseEndpoint):
         return Response({"outputs": formatted})
 
     async def on_shutdown(self, _):
-        await self.communication_manager.shutdown()
+        if self.communication_manager:
+            await self.communication_manager.shutdown()


qstokkink:
ok to test

ipv8/REST/identity_endpoint.py
```diff
@@ -30,11 +30,11 @@ class IdentityEndpoint(BaseEndpoint):
     def __init__(self, middlewares=()):
         super(IdentityEndpoint, self).__init__(middlewares)
         self.communication_manager = None
-        self.app.on_shutdown.append(self.on_shutdown)
 
     def initialize(self, session):
         super(IdentityEndpoint, self).initialize(session)
         self.communication_manager = CommunicationManager(session)
+        self.app.on_shutdown.append(self.on_shutdown)
```
qstokkink, line 37: This line is modifying the declared callback routes after their creation. This is not allowed, explaining the crash. This line change should be reverted.

Instead, in `on_shutdown` it should be checked `if self.communication_manager == None:` before `await self.communication_manager.shutdown()`.

qstokkink (changes requested):

ipv8/REST/identity_endpoint.py
ichorid: Yeah. I'm running with the thing disabled. I'll correct the fix.

qstokkink (approved):
Thanks 👍

ichorid:
:rocket: - fast, as usual :+1:

PR #1020, opened 2021-07-19
Added Payload match support

### LINKED ISSUE #989: Add `Payload.__match_args__` for Python 3.10
Python 3.10 is going to see the introduction of the `match` keyword. If we support `match` for `Payload` instances, this would allow any Python 3.10 users to easily branch off their logic using `Payload` instances, greatly improving the readibility of their code.

It seems we can support `match` by simply defining `Payload.__match_args__` and set it to `Payload.names`. Therefore, this change seems to also be fully backwards compatible.

<sub>Disclaimer: Python 3.10 is still in beta and is supposed to be released later this year. There is a small chance the `match` internals will change.</sub>

qstokkink:
Fixes #989

This PR:

 - Adds `__match_args__` to `VariablePayload` to allow for Python 3.10 structural pattern matching.
 - Fixes `asynctest.helpers.exhaust_callbacks` no longer working in Python 3.10.

Notes:
 - The `exec(compile(...))` construction in `test_lazy_payload.py` exists because Python crashes when importing files with syntax errors. This means that everyone running the tests with a Python version lower than 3.10 would crash when importing this file. As an aside: Python 3.9 has soft keywords for this, but we have to deal with Python 3.7 and Python 3.8 as well.
 - This makes IPv8 Python 3.10 compatible.

### DIFF
--- a/ipv8/messaging/lazy_payload.py
+++ b/ipv8/messaging/lazy_payload.py
@@ -65,6 +65,7 @@ class VariablePayload(Payload):
             raise KeyError("%s missing %d arguments!" % (self.__class__.__name__, len(args) - index))
         if kwargs:
             raise KeyError("%s has leftover keyword arguments: %s!" % (self.__class__.__name__, str(kwargs)))
+        self.__class__.__match_args__ = tuple(self.names)
 
     @classmethod
     def from_unpack_list(cls, *args) -> VariablePayload:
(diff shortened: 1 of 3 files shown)


qstokkink:
validate

qstokkink:
validate

devos50 (approved):
I'm not fully familiar with this part of IPv8, but it looks fine to me 👍 

(the `exec` part is not very nice but I see how it is a necessity)

PR #1026, opened 2021-07-23
Added best practices documentation

### LINKED ISSUE #1022: Add documentation for architectural best practices
This documentation should include:

 - The disclaimer and pledge that you can make whatever you like using IPv8. We only offer suggestions.
 - Try to KISS (dump everything in the same file and overlay class), but switch to SOLID to minimize your overlay code if your code becomes too complex or big.
 - To fulfill the SRP (single-responsibility principle, the S in SOLID), an overlay should only concern itself with messaging between identities - this is the core responsibility of IPv8 as a library.
 - One way to allow for good architectural separation is to keep the state of an overlay in a separate class (following a chain-of-responsibility pattern). This allows you to test the state transitions in isolation and allows you to mock/stage the state-tracking object when testing the overlay messaging. For example, `IdentityManager` and `PseudonymManager` are separately tested in this overlay:

https://github.com/Tribler/py-ipv8/blob/6f5b29c7ecbceb90275df87ead5b9b6a3e7e4340/ipv8/attestation/identity/community.py#L40-L41

 - To run IPv8 as a service (using `ipv8_service.py`), you need to be able to launch your overlay from user settings (i.e., a configuration dictionary of strings and ints). A recurring successful pattern in IPv8 is "create from configuration if not explicitly supplied", for example:

https://github.com/Tribler/py-ipv8/blob/705326bda5622f7e4e59b71a44690c07af79896a/ipv8_service.py#L78-L83

 - To pass a (partial) configuration to your overlay it is often better to supply a settings object instead of passing every configuration parameter separately (avoiding Pylint R0913). For example (note the "create from configuration if not explicitly supplied" pattern again):

https://github.com/Tribler/py-ipv8/blob/7a94e31c1036b7d13491dc63a8e8d55ed34f0c7d/ipv8/messaging/anonymization/community.py#L98-L101

qstokkink:
Fixes #1022

This PR:

 - Adds documentation for Community best practices.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -67,6 +67,7 @@ Table of contents
    :maxdepth: 2
    :caption: References:
 
+   reference/community_best_practices.rst
    reference/configuration.rst
    reference/bootstrapping.rst
    reference/keys.rst
--- a/doc/reference/community_best_practices.rst
+++ b/doc/reference/community_best_practices.rst
@@ -1,0 +1,131 @@
+Community Best Practices
+========================
+
+When working with a small ``Community`` class you can get away with putting all of your code into the same file.
+However, as your codebase grows this simple approach becomes unmaintainable (most importantly it becomes hard to test).
+More formally, whereas you should always start with the KISS principle (**keep it simple, stupid**) we'll show how you can organize your ``Community`` as it grows larger (using the SOLID design principles).
+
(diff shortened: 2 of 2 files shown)


qstokkink:
validate

devos50 (approved):
Looks good! Clear story 👍

PR #1031, opened 2021-07-30
Use system temp directory as base for creating temp dirs

ichorid:
When tests fail mid-way without proper cleanup, TestBase can leave temporary directories in the source dir. This little change solves the problem by pointing tests' temporary directories base to the system temp directory, the same way `pytest` does.

### DIFF
--- a/ipv8/test/base.py
+++ b/ipv8/test/base.py
@@ -2,11 +2,13 @@ import logging
 import os
 import shutil
 import sys
+import tempfile
 import threading
 import time
 import uuid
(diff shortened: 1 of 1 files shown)


qstokkink:
I will close this for two reasons:

1. This is intended behavior. For more information: https://github.com/Tribler/tribler/issues/6158#issuecomment-855891958
2. Files will still be left behind in your solution, just in a different directory. See also https://github.com/Tribler/tribler/pull/6179#issuecomment-867101574

You are welcome to use your changes in your own project by overwriting `temporary_directory` in your `TestBase` subclass.

PR #1039, opened 2021-08-19
Added option to configure IPv8 with raw private key material

### LINKED ISSUE #1037: Allow key configuration with bin
Currently our main downstream project, Tribler, loads IPv8 with this configuration:

```python
ipv8_config_builder = (ConfigBuilder()
                       .set_port(port)
                       .set_address(address)
                       .clear_overlays()
                       .clear_keys()  # We load the keys ourselves
                       .set_working_directory(str(state_dir))
                       .set_walker_interval(self.config.ipv8.walk_interval))
```

Note that both the overlays and keys are wiped, after which Tribler proceeds to load overlays **completely normally and completely manually**. This causes a lot of code duplication. All that seems to be stopping Tribler from making its overlay initialization 1/10<sup>th</sup> of its size (by depending more on the `ConfigBuilder`) is its external management of a single key singleton. We can, almost trivially, support this use case by replacing this code:

https://github.com/Tribler/py-ipv8/blob/705326bda5622f7e4e59b71a44690c07af79896a/ipv8_service.py#L111-L115

with (something like)

```python
 else: 
     self.keys[key_block['alias']] = Peer(default_eccrypto.key_from_private_bin(key_block['bin'])
                                          if 'bin' in key_block else
                                          default_eccrypto.generate_key(key_block['generation'])) 
     if key_block['file']: 
         with open(key_block['file'], 'wb') as f: 
             f.write(self.keys[key_block['alias']].key.key_to_bin()) 
```

and then supplying a `ConfigBuilder.add_key_from_bin()` method.

This would help to reduce the Tribler codebase and potentially other downstream projects as well.

### LINKED ISSUE #1038: Remove ancient key loading from IPv8 class
The `IPv8` class includes the following as part of the key loading procedure:

https://github.com/Tribler/py-ipv8/blob/705326bda5622f7e4e59b71a44690c07af79896a/ipv8_service.py#L100-L110

These key file formats were last used probably around 3 years ago. These workarounds can be removed now.

qstokkink:
Fixes #1037
Fixes #1038

This PR:

 - Adds the option to configure keys with raw private key material (in b64 format).
 - Updates the configuration documentation to have a section for the key format.
 - Removes ancient (and ugly!) key loading practices.

### DIFF
--- a/ipv8/configuration.py
+++ b/ipv8/configuration.py
@@ -320,6 +320,29 @@ class ConfigBuilder(object):
         })
         return self
 
+    def add_key_from_bin(self, alias: str, key_bin_b64: str, file_path: typing.Optional[str] = None):
+        """
+        Add a key by alias and  its raw key material, possibly stored at a certain file path.
+
+        If a key already exists at the given file path, that will be loaded instead of the given key material.
+
+        :param alias: the alias used to reference this key
+        :param key_bin_b64: the base64 encoded private key material
+        :param file_path: the optional file path to save the key to
+        """
+        if 'keys' in self.config:
+            self.config['keys'] = [key for key in self.config['keys'] if key['alias'] != alias]
+        else:
+            self.config['keys'] = []
+        key_config = {
+            'alias': alias,
+            'bin': key_bin_b64
+        }
+        if file_path is not None:
+            key_config['file'] = file_path
+        self.config['keys'].append(key_config)
+        return self
+
     def add_overlay(self,
                     overlay_class: str,
                     key_alias: str,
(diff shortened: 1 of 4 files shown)


qstokkink:
validate

qstokkink:
Ready for review; remaining Flake8 and Pylint errors are due to the `ipv8_service.py` import method.

devos50:
validate

ipv8/configuration.py
```diff
@@ -320,6 +320,29 @@ def add_key(self, alias: str, generation: str, file_path: str) -> ConfigBuilder:
...
+        Add a key by alias and  its raw key material, possibly stored at a certain file path.
+
+        If a key already exists at the given file path, that will be loaded instead of the given key material.
+
+        :param alias: the alias used to reference this key
+        :param key_bin_b64: the base64 encoded private key material
+        :param file_path: the optional file path to save the key to
+        """
+        if 'keys' in self.config:
+            self.config['keys'] = [key for key in self.config['keys'] if key["alias"] != alias]
```
devos50, line 334: Minor inconsistency of using " or ' for accessing dict indices.

devos50 (approved):

devos50 (changes requested):
The `M2CryptoSK` can actually be removed I think 👍

qstokkink:
validate

devos50 (approved):

PR #1040, opened 2021-09-29
Updated Endpoint initialization to use 'interfaces' configuration

### LINKED ISSUE #1036: Allow configuration per interface
In the past IPv8 only had a single IPv4 UDP socket. For its configuration, the IPv8 configuration has a `address` and `port`. However, nowadays IPv8 can dispatch to multiple interfaces, which have to be manually configured and forwarded to IPv8. Instead, we should allow the IPv8 service to automatically initialize endpoints again.

To allow automatic initialization through the configuration we should probably simply match the `DispatcherEndpoint` configuration.

https://github.com/Tribler/py-ipv8/blob/e378ff1389d32625d23cf4286dfa1fd69d09041a/ipv8_service.py#L76-L81

Instead of an `address` and `port` key, we could take an `interfaces` key in the configuration. This would be a named mapping of intialization parameters for the specific interface. For example:

```python
config = {
    'interfaces': [  # <- list
        {
            'interface': 'UDPIPv4',
            'address': '0.0.0.0',
            'port': 8090
        },
        {
            'interface': 'UDPIPv6',
            'address': '::',
            'port': 8092
        }
    ]
}
```

Note that the first argument to `DispatcherEndpoint` specifies the interface preference. The interfaces configuration should match this and be an ordered list instead of a set.

Removing `address` and `port` will break just about _everything_ using IPv8. Therefore, it may be wise to keep `address` and `port` as a deprecated fallback for now.

qstokkink:
Fixes #1036

This PR:

 - Adds configurable `DispatcherEndpoint` initialization in `ipv8_service.py`.
 - Adds a warning if the old `"address"` and `"port"` configuration is used. Old configurations still work though.
 - Updates the `configuration.rst` documentation to showcase the use of `'interfaces'` instead of `"address"` and `"port"`.
 - Updates the `configuration.py` to use the new `'interfaces'` configuration.
 - Removes some `asserts` (and corresponding tests) from `ConfigBuilder.finalize()` that were too strict or no longer generic enough for `'interfaces'`. For example, it is allowed (but unconventional) to launch IPv8 without any network interfaces at all.

Preview of documentation change:

preview

### DIFF
--- a/doc/reference/configuration.rst
+++ b/doc/reference/configuration.rst
@@ -7,7 +7,11 @@ By invoking ``get_default_configuration()``, you can get a dictionary copy of th
 
 .. |snip1| raw:: html
 
-  "0.0.0.0"
+  <small>[{<br />
+  &emsp;<nobr>'interface':<nobr>"UDPIPv4",<br />
+  &emsp;<nobr>'ip':<nobr>"0.0.0.0",<br />
+  &emsp;<nobr>'port':<nobr>8090<br />
+  }]</small>
 
 
 .. |nl| raw:: html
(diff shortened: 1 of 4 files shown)


qstokkink:
validate

qstokkink:
validate

devos50 (approved):

PR #1043, opened 2021-09-30
Fixed deprecated config not having fallbacks

qstokkink:
This PR:

 - Fixes the deprecated 'address'/'port' fallback mode not auto-filling the old default values.


qstokkink:
validate

devos50 (approved):

PR #1049, opened 2021-10-15
Fixed dataclass typing

### LINKED ISSUE #1047: Add dataclass-compatible Payload typing
IPv8 now has [dataclass Payloads](https://github.com/Tribler/py-ipv8/pull/1045). The downside of this is that `dataclass`-wrapped objects do not have a `dataclass` type. There is no such thing as `Type[dataclass]`, though this is exactly what our dataclass Payloads offer.

To address the issue of a missing `dataclass` type, #1045 uses `Union[Type[Payload], Type[Any]]`. This is equivalent to `Type[Any]`. Instead, we should have a new type in `types` that is better at suggesting to programmers that the IPv8 interfaces expect a `Payload` or `dataclass`-wrapped `Payload` here.

One idea is to provide a new `AnyPayload` type (equivalent to `Any`). This tells our programmers we want `Payload` instances, even though the static type checking does nothing.

### LINKED ISSUE #1048: `ez_send` typing incompatible with dataclass payloads
_[thanks @drew2a for reporting]_

The `ez_send` method expects `Payload` instances:

https://github.com/Tribler/py-ipv8/blob/1a25b0cdad299dce6f1989adc828e7904122ace0/ipv8/lazy_community.py#L189

This typing is incompatible with the new `dataclass`-wrapped Payloads of IPv8. This should be `Union[Type[Payload], Type[Any]]`, or a better type (see #1047).

qstokkink:
Fixes #1047
Fixes #1048

This PR:

 - Adds `DataclassPayload` typing.
 - Adds `AnyPayload` typing
 - Fixes `ez_send()` not allowing for dataclass payloads.
 - Updates the `lazy_community.py` bindings to use `AnyPayload` typing.

### DIFF
--- a/ipv8/lazy_community.py
+++ b/ipv8/lazy_community.py
@@ -1,12 +1,12 @@
 from abc import ABC, abstractmethod
 from functools import wraps
-from typing import Any, Iterable, List, Tuple, Type, Union
+from typing import Iterable, List, Tuple, Type, Union
 
 from .keyvault.crypto import default_eccrypto
 from .messaging.payload_headers import BinMemberAuthenticationPayload, GlobalTimeDistributionPayload
 from .overlay import Overlay
 from .peer import Peer
-from .types import Address, NumberCache, Payload
+from .types import Address, AnyPayload, AnyPayloadType, NumberCache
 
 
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

PR #1058, opened 2021-11-12
Fixed test suite crash with 0.0.0.0 subnet mask

### LINKED ISSUE #1057: IPv8 & Tribler tests hangs when there is a network interface with mask `0.0.0.0`
Yesterday core tests for Tribler stop working on my Windows machine. Some tests pass, but most of them hang and then terminate on timeout.

As it turns out, problem is in [`AutoMockEndpoint._generate_unique_address`](https://github.com/Tribler/py-ipv8/blob/fcf2c7971a9b3d91c4dfb159f6e17f68c081f1d0/ipv8/test/mocking/endpoint.py#L118) method:

```python
    def _generate_unique_address(self):
        address = self._generate_address()

        while address in internet or self._is_lan(address):
            address = self._generate_address()

        return address
```

This while loop inside the method works as an infinite loop. The reason for this bug is that one of the network interfaces has netmask `0.0.0.0`:

```python
>>> import netifaces
>>> netifaces.interfaces()
['{E98500B3-F5E4-F4C1-6D2B-5D3145D2B1AC}', '{76452E84-E598-4A0E-B612-6CBD22E518D8}', '{12FC1DF3-FDF1-44FD-BDD1-6562A0165CB1}', '{DE130FDB-E5EB-4E5A-A08E-C51F158183A4}', '{ED752F61-E6C3-11EA-9AF0-806E6F6E6963}']
>>> netifaces.ifaddresses('{E98500B3-F5E4-F4C1-6D2B-5D3145D2B1AC}')
{-1000: [{'addr': ''}], 2: [{'addr': '10.0.0.2', 'netmask': '0.0.0.0', 'broadcast': '255.255.255.255'}]}
```

When the `AutoMockEndpoint._is_lan(address)` method executes inside a loop, it compares a new random IP address with a local interface IP address. It applies the netmask `0.0.0.0` to both addresses reducing them to two zeroes, then compares two zeroes and concludes that the new random IP address belongs to this local network. So,  `_generate_unique_address` get stuck in an infinite loop because every random IP address is considered belonging to a local network with mask `0.0.0.0`.

When a network interface has mask `0.0.0.0`, it means that [the network interface isn't bound to any network at all; not even localhost requests will be acknowledged. If anything, it's a way of the operating system to present the interface as available but inactive and un-routed](https://superuser.com/questions/305850/what-does-mask-0-0-0-0-mean-when-its-assigned-to-a-network-interface).

I think the `EndpointListener._get_interface_addresses` method should skip interfaces with mask `0.0.0.0`; it should solve the problem.

kozlovsky:
Fixes #1057

This PR:

 - Adds `netifaces` as a pylint external package.
 - Fixes yielding `Endpoint` `Interface` objects with a netmask of `0.0.0.0`.

### DIFF
--- a/ipv8/messaging/interfaces/endpoint.py
+++ b/ipv8/messaging/interfaces/endpoint.py
@@ -214,17 +214,13 @@ class EndpointListener(metaclass=abc.ABCMeta):
 
                 else:
                     for option in addresses.get(netifaces.AF_INET, []):
-                        try:
-                            # On Windows netifaces currently returns IP addresses as unicode,
-                            # and on *nix it returns str. So, we convert any unicode objects to str.
-                            # Python 3 port update: In Python 3 everything is unicode and we should do nothing.
-                            # On Python 2 the above can happen: instead of checking for unicode, we check for `not str`.
-                            unicode_to_str = lambda s: s.encode('utf-8') if s and not isinstance(s, str) else s
-                            yield Interface(interface,
-                                            unicode_to_str(option.get("addr")),
-                                            unicode_to_str(option.get("netmask")),
-                                            unicode_to_str(option.get("broadcast")))
+                        addr, netmask, broadcast = [option.get(field) for field in ("addr", "netmask", "broadcast")]
 
+                        if netmask == "0.0.0.0":
+                            continue  # The network interface isn't bound to any network, so we skip it
+
+                        try:
+                            yield Interface(interface, addr, netmask, broadcast)
                         except TypeError:
                             # some interfaces have no netmask configured, causing a TypeError when
                             # trying to unpack _l_netmask
--- a/ipv8/test/messaging/interfaces/test_endpoint.py
+++ b/ipv8/test/messaging/interfaces/test_endpoint.py
@@ -1,0 +1,30 @@
+from unittest.mock import patch
+
+from netifaces import AF_INET
+
+from ...base import TestBase
+from ...mocking.endpoint import AutoMockEndpoint
+from ....messaging.interfaces.endpoint import EndpointListener
+
+FAKE_ADDRESSES = {
+    'interface1': {AF_INET: [{'addr': '10.0.0.2', 'netmask': '0.0.0.0', 'broadcast': '255.255.255.255'}]},
+    'interface2': {AF_INET: [{'addr': '192.168.15.1', 'netmask': '255.255.255.0', 'broadcast': '192.168.15.255'}]},
+    'interface3': {AF_INET: [{'addr': '127.0.0.1', 'netmask': '255.0.0.0', 'broadcast': '127.255.255.255'}]}
+}
+
+
+class TestableEndpointListener(EndpointListener):
+    def on_packet(self, packet):
+        pass
+
+
+class TestEndpointListener(TestBase):
+
+    @patch('netifaces.interfaces', lambda: FAKE_ADDRESSES)
+    @patch('netifaces.ifaddresses', lambda iface: FAKE_ADDRESSES[iface])
+    def test_get_interface_addresses(self):
+        """Test that the _get_interface_addresses method returns all interfaces except the one with netmask 0.0.0.0"""
+        listener = TestableEndpointListener(AutoMockEndpoint())
+        interfaces = list(listener._get_interface_addresses())
+        self.assertNotIn("0.0.0.0", [interface.netmask for interface in interfaces])
+        self.assertEqual(len(FAKE_ADDRESSES) - 1, len(interfaces))  # all except the one with netmask 0.0.0.0
--- a/.pylintrc
+++ b/.pylintrc
@@ -33,7 +33,7 @@ unsafe-load-any-extension=no
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code
-extension-pkg-whitelist=libtorrent
+extension-pkg-whitelist=libtorrent,netifaces
 
 # Allow optimization of some AST trees. This will activate a peephole AST
 # optimizer, which will apply various small optimizations. For instance, it can


qstokkink (changes requested):
I have two issues with this PR:
1. It is setting `netmask` after `self.netmask = netmask`.
2. The Interface object is being returned even though the interface is essentially defunct. For this reason, I'd prefer not yielding it at all in line 223 (old)/228 (new).

kozlovsky:
> The Interface object is being returned even though the interface is essentially defunct.
> For this reason, I'd prefer not yielding it at all in line 223 (old)/228 (new).

Skipping this interface altogether was my initial thought as well. But what if some random IP address accidentally matches the IP address of this interface? Should we allow for `_generate_unique_address` to return this IP address, or can it break something?

> It is setting netmask after self.netmask = netmask

It was deliberate. If we change it before the assignment to `self.netmask`, we could not see the real netmask of the interface. We can mistakenly consider it later as a valid interface with a single IP address.

By assigning it later, we still have the real netmask `0.0.0.0` and correct comparison in the `__contains__` method.

We can remove this logic if it is not a problem to accidentally choose this interface's IP address in `_generate_unique_address. But I'm not sure about that.

qstokkink:
> Should we allow for _generate_unique_address to return this IP address, or can it break something?

It should be fine. But, based on your comment, I will extend my request for changes to also add a unit test for this situation.

Regarding the `netmask`: `_l_netmask` *should* only be the unpacked form of `netmask`. Your patch violates this relationship. I'd rather see the interface not offered at all than the interface being internally inconsistent.

kozlovsky:
> Regarding the netmask: _l_netmask should only be the unpacked form of netmask.

Ok, understood

kozlovsky:
It is indeed possible to skip Interfaces with netmask 0.0.0.0 in the current IPv8 code, as AutoMockEndpoint never attempts to use its random IP addresses. Based on this, I changed the PR code to skip interface creation, as you suggested.

I added the `netifaces` property to EndpointListener to be able to override it in tests easily. If you don't like it, I can revert to patching the global netifaces module instead.

Before this PR, the `my_estimated_wan` attribute of the endpoint listener could take the IP address of an unbound network interface with mask `0.0.0.0`. I added a test that this network interface is skipped when determining estimated LAN and WAN addresses.

kozlovsky:
Oops, deleted remote branch accidentally. Will reopen it in a different PR

ipv8/messaging/interfaces/endpoint.py
```diff
@@ -11,6 +11,8 @@
 except ImportError:
     netifaces = None
 
+# pylint: disable=c-extension-no-member
```
qstokkink, line 14: Please enable linting instead of disabling it:

```patch
diff --git a/.pylintrc b/.pylintrc
index 383925d..02d8f02 100644
--- a/.pylintrc
+++ b/.pylintrc
@@ -33,7 +33,7 @@ unsafe-load-any-extension=no
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code
-extension-pkg-whitelist=libtorrent
+extension-pkg-whitelist=libtorrent,netifaces
 
 # Allow optimization of some AST trees. This will activate a peephole AST
 # optimizer, which will apply various small optimizations. For instance, it can
```

ipv8/test/messaging/interfaces/udp/test_endpoint.py
```diff
@@ -41,6 +77,10 @@ async def tearDown(self):
             self.endpoint2.close()
         await super(TestUDPEndpoint, self).tearDown()
 
+    def test_estimated_wan(self):
+        # The first address 10.0.0.2 is skipped as it has netmask 0.0.0.0
+        self.assertEqual(self.endpoint2_listener.my_estimated_wan[0], '192.168.15.1')
```
qstokkink, line 82: The following needs to be addressed for this unit test:

 - The test is testing `EndpointListener` functionality and should therefore be in the`TestEndpointListener` test case, in the `ipv8/test/messaging/endpoint.py` file.
 - You changed `_get_interface_addresses()` and that method is not being called directly, making this test very fragile in case of future changes to any of the intermediate calls between `_get_interface_addresses()` and `my_estimated_wan`.
 - The arranging for this test is passive, providing this particular netmask case for all other units.
 - There is no inverted unit test to accompany this test (what if the netmask is not `0.0.0.0`?).
 - The unit test itself should have a single-line docstring explaining what the test does, preferably without reading the code of the test itself.

ipv8/messaging/interfaces/endpoint.py
```diff
@@ -128,11 +130,15 @@ def __init__(self, endpoint, main_thread=True):
...
-        self._netifaces_failed = netifaces is None
+        self._netifaces_failed = self.netifaces is None
         self._local_interfaces = []
         self._my_estimated_lan = None
         self.my_estimated_wan = self.my_estimated_lan
 
+    @property
+    def netifaces(self):  # For easier mocking in tests
+        return netifaces
+
```
qstokkink, line 141: I'd consider creating mocking code for an entire external dependency the very last resort. It is unnecessary here. Also, since the `EndpointListener` is an ABC you have to make a subclass anyway. Please refactor the `netifaces.interfaces()` call into its own method and use that in `_get_interface_addresses()`.

 For example, in `EndpointListener`:

```python
@classmethod
def get_interfaces(cls):
    return netifaces.interfaces()

@classmethod
def get_interface_addresses(cls, interface):
    return netifaces.ifaddresses(interface)

@classmethod
def _get_interface_addresses(cls):
    ...
    for interface in cls.get_interfaces():
        try:
            addresses = cls.get_interface_addresses(interface)
    ...
```

In the test class you can implement the ABC using an artificial return value for `get_interfaces()`:

```python
class ClassUnderTest(EndpointListener):
    @classmethod
    def get_interfaces(cls):
        return FAKE_ADDRESSES.keys()

    @classmethod
    def get_interface_addresses(cls, interface):
        return FAKE_ADDRESSES[interface]
```

qstokkink (changes requested):
Thanks for your contribution. Please resolve the issues I raised in this review. Secondly, please make sure your PR formatting follows our contributing guidelines.

ipv8/messaging/interfaces/endpoint.py
kozlovsky: Done

ipv8/test/messaging/interfaces/udp/test_endpoint.py
kozlovsky: I think I addressed all of these points:
* I put the test in a new file `ipv8/test/messaging/interfaces/test_endpoint.py` (note the `interfaces` subfolder, I think it is where the test should be placed, as `EndpointListener` is also inside the `interfaces` subfolder).
* Now the test tests the `_get_interface_addresses()` method directly.
* The test is local and does not change the behavior of other tests.
* The test checks both the absence of interface with netmask 0.0.0.0 and the presence of other interfaces.
* The docstring is added.

ipv8/messaging/interfaces/endpoint.py
kozlovsky: Instead of adding these helper methods, I used the `patch` decorator of the `unittest.mock` library, this way I do not overcomplicate the class just for the testing.

ipv8/messaging/interfaces/endpoint.py
qstokkink: Thanks 👍

ipv8/test/messaging/interfaces/udp/test_endpoint.py
qstokkink: Nice, thanks 👍

ipv8/messaging/interfaces/endpoint.py
qstokkink: I think that is an acceptable solution for the target of this PR. However, I consider having to use `patch` to be a code smell of bad architecture (this code is 8 years old after all) and I will open an issue after this PR has been merged.

ipv8/test/messaging/interfaces/test_endpoint.py
```diff
@@ -0,0 +1,30 @@
+from unittest.mock import patch
+
+from ipv8.test.mocking.endpoint import AutoMockEndpoint
```
qstokkink, line 3: This needs to be a relative import.

qstokkink (changes requested):
Nice changes 👍 Only the absolute import absolutely needs to be changed.

Please also follow the PR message template (see also the links I posted in https://github.com/Tribler/py-ipv8/pull/1060#pullrequestreview-943919794). Lastly, a bit of a nit pick, the PR title doesn't capture the changes you made concisely (e.g., "Fixed test suite crash with 0.0.0.0 subnet mask" captures the PR's core without referring to issue numbers).

ipv8/messaging/interfaces/endpoint.py
kozlovsky: I agree that there are many ways to approach the testability of `_get_interface_addresses`; each has its trade-offs. In my opinion, the solution with class methods is more problematic for the following reasons:
* It continues the general trend used in IPv8 to rely on class inheritance. It is commonly considered better practice to prefer [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) to make code decoupled.
* Class methods are often not flexible enough. If someone needs to write several test cases with different sets of fake addresses, it may be harder to do it with class methods.
* New class methods became part of `EndpointListener` API, complicating it. The naming can become an issue. Do you really want two methods, `get_interface_addresses` and `_get_interface_addresses`, with very similar names but different return types and parameters?
* Why these new methods are public (that is, named without a leading underscore) while they are only called from a protected method `_get_interface_addresses`? Should they be protected methods as well?

Solution with `patch` avoids these issues and looks a bit clearer for me. But it is a matter of personal preferences.

I agree that it is better to refactor the `_get_interface_addresses` method to make it more testable. For example, move the `Interface` class to the top-level to make it testable as well. But it is indeed better to do it in a separate PR.

ipv8/test/messaging/interfaces/test_endpoint.py
kozlovsky: Fixed!

qstokkink:
Validate

qstokkink (approved):
Thanks!

I modified your PR title and message to fit the IPv8 template.

ipv8/messaging/interfaces/endpoint.py
qstokkink: Just because composition is preferred, does not mean inheritance is bad. One is not a better practice than the other. A good architecture applies both to meet its goals. There is more than one design pattern/consideration that must be applied. "Everything must be composable" is a cargo-cult-programming mindset (see also [the list of architecture anti-patterns](https://en.wikibooks.org/wiki/Introduction_to_Software_Engineering/Architecture/Anti-Patterns)).

Class methods are more than flexible enough to do anything short from hyper-composition microservices, which is not how IPv8 is used (and I refer you to [the last Q&A point of IPv8's contributing guidelines](https://github.com/Tribler/py-ipv8/blob/master/CONTRIBUTING.md#common-qa) if you think it is).

> Solution with `patch` avoids these issues and looks a bit clearer for me. But it is a matter of personal preferences.

I see how the resulting unit test code is smaller for the current implementation. However, it does not solve the underlying issue of the code not being easy to test. I don't agree that bad design of the `Endpoint` legacy code is a matter of personal preference. To a certain extent I agree that we should not touch legacy code: "if it works, don't fix it". However, I just linked to a list of anti-patterns that include the "lava flow" anti-pattern: if we start adding more tests, the lack of testability will only make tests worse and worse (even `patch` won't help you then). Then again, will we actually add more tests any time soon? Probably not.

Reflecting your comments---and this thread as a whole---I'll opt not to make an issue. If anyone wants to chase 100% test coverage in the (far?) future, they can solve the testability issue at that point in time. Let's leave the legacy code rest in peace.

PR #1060, opened 2021-11-12
Fixed shutdown of RequestCache in tests

### LINKED ISSUE #1089: Missed shutdown of RequestCache in tests
* Run tests from test_requestcache.py
* You can see multiple warnings:

```
Task was destroyed but it is pending!
task: <Task pending name='RequestCache:_check_tasks' coro=<interval_runner() running at C:\dev\py-ipv8\ipv8\taskmanager.py:16> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x000002587FCF8880>()] created at C:\Python39\lib\asyncio\base_events.py:424> cb=[TaskManager.register_task.<locals>.done_cb() at C:\dev\py-ipv8\ipv8\taskmanager.py:128] created at C:\dev\py-ipv8\ipv8\taskmanager.py:110>
```

The reason for the warnings is missed shutdown of RequestCache instance in multiple tests. Proper shutdown should be added.

kozlovsky:
Fixes #1089

This PR:
* Fixes the shutdown of the `RequestCache` instance in multiple tests of `test_requestcache.py`

### DIFF
--- a/ipv8/test/test_requestcache.py
+++ b/ipv8/test/test_requestcache.py
@@ -59,14 +59,13 @@ class TestRequestCache(TestBase):
         """
         Test if RequestCache does not allow new Caches after shutdown().
         """
-        num_tasks = len(all_tasks())
-        request_cache = RequestCache()  # This adds a task, don't use ``self.request_cache`` here!
-        request_cache.add(MockCache(request_cache))
-        self.assertEqual(len(all_tasks()), num_tasks + 2)
-        await request_cache.shutdown()
-        self.assertEqual(len(all_tasks()), num_tasks)
-        request_cache.add(MockCache(request_cache))
-        self.assertEqual(len(all_tasks()), num_tasks)
+        num_tasks = len(all_tasks())  # [Background tasks (depends on test runner) + RequestCache]
+        self.request_cache.add(MockCache(self.request_cache))
+        self.assertEqual(len(all_tasks()), num_tasks + 1)  # [Background + RequestCache + Cache]
+        await self.request_cache.shutdown()
+        self.assertEqual(len(all_tasks()), num_tasks - 1)  # [Background]
+        self.request_cache.add(MockCache(self.request_cache))  # No tasks should have been added
+        self.assertEqual(len(all_tasks()), num_tasks - 1)  # [Background]
 
     async def test_timeout(self):
         """
@@ -75,6 +74,7 @@ class TestRequestCache(TestBase):
         cache = MockCache(self.request_cache)
         self.request_cache.add(cache)
         await cache.timed_out
+        await self.request_cache.shutdown()
 
     async def test_add_duplicate(self):
         """
@@ -151,6 +151,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_timeout(self):
         """
         Test if passthrough respects the timeout value.
@@ -162,6 +164,8 @@ class TestRequestCache(TestBase):
             await sleep(0.0)
 
         self.assertFalse(cache.timed_out)
+
+        await self.request_cache.shutdown()
 
     async def test_passthrough_filter_one_match(self):
         """
@@ -175,6 +179,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_one_mismatch(self):
         """
         Test if passthrough filters correctly with one filter, that doesn't match
@@ -186,6 +192,8 @@ class TestRequestCache(TestBase):
             await sleep(0.0)
 
         self.assertFalse(cache.timed_out)
+
+        await self.request_cache.shutdown()
 
     async def test_passthrough_filter_many_match(self):
         """
@@ -199,6 +207,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_some_match(self):
         """
         Test if passthrough filters correctly with many filters, for which some match
@@ -211,6 +221,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_no_match(self):
         """
         Test if passthrough filters correctly with many filters, for which none match
@@ -222,3 +234,5 @@ class TestRequestCache(TestBase):
             await sleep(0.0)
 
         self.assertFalse(cache.timed_out)
+
+        await self.request_cache.shutdown()


qstokkink (changes requested):
Thanks for reporting these leaking tasks. However, your fix is hiding potential errors. These tests are supposed to manually shut down the `RequestCache` to make leaking tasks visible. If you shut down the `RequestCache` twice the tests could erroneously pass.

I'd prefer this:

<details> <summary> Click me for patch. </summary>

```python
diff --git a/ipv8/test/test_requestcache.py b/ipv8/test/test_requestcache.py
index e8cebcb..8e24306 100644
--- a/ipv8/test/test_requestcache.py
+++ b/ipv8/test/test_requestcache.py
@@ -59,14 +59,13 @@ class TestRequestCache(TestBase):
         """
         Test if RequestCache does not allow new Caches after shutdown().
         """
-        num_tasks = len(all_tasks())
-        request_cache = RequestCache()  # This adds a task, don't use ``self.request_cache`` here!
-        request_cache.add(MockCache(request_cache))
-        self.assertEqual(len(all_tasks()), num_tasks + 2)
-        await request_cache.shutdown()
-        self.assertEqual(len(all_tasks()), num_tasks)
-        request_cache.add(MockCache(request_cache))
-        self.assertEqual(len(all_tasks()), num_tasks)
+        num_tasks = len(all_tasks())  # [Background tasks (depends on test runner) + RequestCache]
+        self.request_cache.add(MockCache(self.request_cache))
+        self.assertEqual(len(all_tasks()), num_tasks + 1)  # [Background + RequestCache + Cache]
+        await self.request_cache.shutdown()
+        self.assertEqual(len(all_tasks()), num_tasks - 1)  # [Background]
+        self.request_cache.add(MockCache(self.request_cache))  # No tasks should have been added
+        self.assertEqual(len(all_tasks()), num_tasks - 1)  # [Background]
 
     async def test_timeout(self):
         """
@@ -75,6 +74,7 @@ class TestRequestCache(TestBase):
         cache = MockCache(self.request_cache)
         self.request_cache.add(cache)
         await cache.timed_out
+        await self.request_cache.shutdown()
 
     async def test_add_duplicate(self):
         """
@@ -151,6 +151,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_timeout(self):
         """
         Test if passthrough respects the timeout value.
@@ -163,6 +165,8 @@ class TestRequestCache(TestBase):
 
         self.assertFalse(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_one_match(self):
         """
         Test if passthrough filters correctly with one filter, that matches
@@ -175,6 +179,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_one_mismatch(self):
         """
         Test if passthrough filters correctly with one filter, that doesn't match
@@ -187,6 +193,8 @@ class TestRequestCache(TestBase):
 
         self.assertFalse(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_many_match(self):
         """
         Test if passthrough filters correctly with many filters, that all match
@@ -199,6 +207,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_some_match(self):
         """
         Test if passthrough filters correctly with many filters, for which some match
@@ -211,6 +221,8 @@ class TestRequestCache(TestBase):
 
         self.assertTrue(cache.timed_out)
 
+        await self.request_cache.shutdown()
+
     async def test_passthrough_filter_no_match(self):
         """
         Test if passthrough filters correctly with many filters, for which none match
@@ -222,3 +234,5 @@ class TestRequestCache(TestBase):
             await sleep(0.0)
 
         self.assertFalse(cache.timed_out)
+
+        await self.request_cache.shutdown()
```

</details>

Secondarily, please ensure your PR formatting meets the IPv8 contributing guidelines.

kozlovsky:
> I'd prefer this

Done, I also added a check that the `add` method of the RequestCache emits a log warning when called after the shutdown of RequestCache.

> Secondarily, please ensure your PR formatting meets the IPv8 contributing guidelines.

I edited the PR title to meet the guidelines. Probably it is too late to open a corresponding issue for this PR, as we already discussed all topics, I think :)

ipv8/test/test_requestcache.py
```diff
@@ -59,14 +60,16 @@ async def test_shutdown(self):
...
+        num_tasks = len(all_tasks())  # [Background tasks (depends on test runner) + RequestCache]
+        self.request_cache.add(MockCache(self.request_cache))
+        self.assertEqual(len(all_tasks()), num_tasks + 1)  # [Background + RequestCache + Cache]
+        await self.request_cache.shutdown()
+        self.assertEqual(len(all_tasks()), num_tasks - 1)  # [Background]
+        with self.assertLogs(level=logging.WARNING) as log:
+            self.request_cache.add(MockCache(self.request_cache))  # No tasks should have been added
+        self.assertEqual(len(all_tasks()), num_tasks - 1)  # [Background]
+        self.assertEqual(len(log.records), 1)
+        self.assertEqual(log.records[0].msg, 'Dropping %s due to shutdown!')
```
qstokkink, line 72: Your change is asserting "logging output" on a unit that is asserting the "number of asyncio tasks". Therefore, your change should be in a separate unit test.

My secondary concern with your change is that you're checking the logging message itself. Whereas I see "logging a warning"---though this is not done in any other IPv8 unit test yet---as an improvement over the current tests, the exact log message is an implementation detail that should not be asserted. [In my opinion this StackOverflow thread nicely covers why and the differing opinions on this topic.](https://softwareengineering.stackexchange.com/q/384164)

```suggestion
        self.request_cache.add(MockCache(self.request_cache))  # No tasks should have been added
        self.assertEqual(len(all_tasks()), num_tasks - 1)  # [Background]
        
    async def test_add_post_shutdown_warning(self):
        """
        Test if RequestCache logs a warning when adding a cache after shutdown.
        """
        await self.request_cache.shutdown()
        with self.assertLogs(level=logging.WARNING) as log:
            self.request_cache.add(MockCache(self.request_cache))
        self.assertGreaterEqual(len(log.records), 1)
```

qstokkink (changes requested):
Thanks for the changes. However, I still have two open requests for changes:

 - As you introduced a new change since my last review, I have one another request for changes.
 - Your PR title meets the guidelines, but your PR message does not. You can find the template [here](https://github.com/Tribler/py-ipv8/blob/master/.github/pull_request_template.md) and check out #1088 for an example of how to use it.

ipv8/test/test_requestcache.py
kozlovsky: > Your change is asserting "logging output" on a unit that is asserting the "number of asyncio tasks". Therefore, your change should be in a separate unit test.

I'm not sure I understand correctly what you mean by "unit" here, but the unit test mentioned was not for testing the "number of asyncio tasks". It is named "test_shutdown" and tests the behavior of the "request_cache.shutdown()" call. From the docstring: "Test if RequestCache does not allow new Caches after shutdown()". Nothing in it assumes that the only check belonging to this test is for the "number of asyncio tasks". As issuing a log warning is part of the behavior of "request_cache.shutdown()", it looks like the proper place to test it.

> the exact log message is an implementation detail that should not be asserted

This was the reason I was testing the message template and not the **exact** log message, as the exact log message includes unstable repr of a cache object.

> this StackOverflow thread nicely covers why

The most popular answer in the threads tells that "you should absolutely test your logging" and that "you should test that the action of logging was done", "not the log files concretely". Testing the log record and not the resulted formatted output of a physical log stream is exactly what is advised here. Also, it is brittle just to test that the log action was done without checking the message, as it may be some unrelated log action, like an unrelated deprecation warning.

Anyway, I think it is not as important to test log messages here, so I removed the check. Now, the code exactly corresponds to your last proposal.

qstokkink: My thoughts on this are as follows:

The log message is definitely part of the current implementation. It is definitely helpful when debugging. However, a programmer should _never_ depend on the template of the log message when writing code. Therefore, it should not be in a unit test.

I think that there is a case to be made that receiving a warning (without testing its mesage or message template) is part of the promise to our users ("behavior" as you call it). I agree that it is hard (impossible?) to test whether a log message orginates from any particular code and I agree with your resolution.

qstokkink:
validate

qstokkink (approved):
Thanks 👍

PR #1068, opened 2021-12-12
Added documentation on third-party serialization (and nested payloads)

### LINKED ISSUE #978: Extend documentation on serialization
The first question everyone seems to have is how to use JSON with IPv8. We should extend the documentation to cover:

 - How to send JSON.
 - Why you would* and why you would not want to use JSON.
 - An example on when JSON goes south.

An easy example would be sending over a byte that contains flag bits and a public key (binary data, fixed length). This takes IPv8 3 lines of code. Attempting the same thing in JSON shows that (1) you have no control over your binary format and (2) it takes about 50 lines of code to cover the same edge cases when unpacking.

<sub>*This is not a hit piece on JSON. When a simple key-value storage is required, it is perfectly fine to use.</sub>

qstokkink:
Fixes #978

This PR:

 - Adds documentation for the three non-destructive ways to hook in third-party serialization.
 - Adds short examples to documentation for nested payloads.
 
Here's the page printed to PDF (doesn't use RTD theme):
[Message serialization — IPv8 2.8.0 documentation.pdf](https://github.com/Tribler/py-ipv8/files/7698920/Message.serialization.IPv8.2.8.0.documentation.pdf)
New stuff starts from the `Using external serialization options` section.

### DIFF
--- a/doc/reference/serialization.rst
+++ b/doc/reference/serialization.rst
@@ -236,3 +236,20 @@ Note that we need to do our own signature checks now.
 
 .. literalinclude:: serialization_6.py
    :lines: 11-37
+
+
+Nested Payloads
+---------------
+
+It is possible to put a ``Payload`` inside another ``Payload``.
+We call these nested payloads.
+You can specify them by using the ``"payload"`` datatype and setting the ``Payload`` class in the format list.
+For a ``VariablePayload`` this looks like the following example.
+
+.. literalinclude:: serialization_7.py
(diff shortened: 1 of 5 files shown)


qstokkink:
validate

devos50:
Since I've been personally summoned, I'll review it after lunch 👍

devos50 (approved):
Looks good 👍

PR #1071, opened 2021-12-20
Fixed bootstrap received addresses callback not checking for cancel

qstokkink:
Fix for downstream report https://github.com/Tribler/tribler/issues/6648

This PR:

 - Updates `Community.received_bootstrapped_addresses` to check if the passed `Future` is cancelled.

### DIFF
--- a/ipv8/test/test_community.py
+++ b/ipv8/test/test_community.py
@@ -1,6 +1,7 @@
 from .base import TestBase
 from .mocking.endpoint import AutoMockEndpoint, MockEndpointListener
 from .mocking.ipv8 import MockIPv8
+from ..bootstrapping.dispersy.bootstrapper import DispersyBootstrapper
 from ..community import Community
 from ..peer import Peer
 from ..peerdiscovery.network import Network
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

ipv8/test/test_community.py
```diff
@@ -97,3 +98,44 @@ async def test_init_strange_id(self):
...
+    Tests for the Community to Bootstrapper interface.
+
+    Note: don't put tests for the Bootstrapper implementations here.
+    """
+
+    async def test_empty_bootstrap(self):
+        """
+        Check if unloading a Community after waiting for bootstrapping results exits cleanly.
+        """
+        community = NewCommunity(Peer(b'LibNaCLPK:' + b'0' * 32), AutoMockEndpoint(), Network())
```
devos50, line 114: **Suggestion:** Are there more places in the code where you require a "fake"/empty key? If so, you might consider extracting this in a constant.

devos50 (approved):
I think I also saw this error show in the console when closing Tribler in GUI test mode so hopefully that's also fixed now.

ipv8/test/test_community.py
qstokkink: Thanks for the suggestion. For now, there are just (including this PR) 4 instances and only in this file. I do see equivalent logic in Tribler appearing [here](https://github.com/Tribler/tribler/blob/e6f9dae5d8b10372c7623b0a11b90eae2dc2c7ff/src/tribler-core/tribler_core/components/payout/tests/test_payout_manager.py#L14-L16).

If either 0-key peers become too common in IPv8 or there is a clear need for this in Tribler I'll extract a 0-key `Peer` into `test/mocking/peer.py`. To keep this PR focused, I'll defer that change (if needed) to a different PR.

PR #1075, opened 2021-12-22
Updated overlay tutorial to use dataclass payloads

### LINKED ISSUE #1055: Change overlay tutorial to use dataclass payloads
Currently our [overlay tutorial](https://py-ipv8.readthedocs.io/en/latest/basics/overlay_tutorial.html) shows this to first-time overlay programmers (in [overlay_tutorial_5.py](https://github.com/Tribler/py-ipv8/blob/master/doc/basics/overlay_tutorial_5.py)):

```python
# <current imports, etc.>

@vp_compile
class MyMessage(VariablePayload):
    msg_id = 1  # The byte identifying this message, must be unique per community.
    format_list = ['I']  # When reading data, we unpack an unsigned integer from it.
    names = ["clock"]  # We will name this unsigned integer "clock"
```

We could replace this with (something like) the following:

```python
from dataclasses import dataclass
# <current imports, etc.>
from pyipv8.ipv8.messaging.payload_dataclass import overwrite_dataclass

dataclass = overwrite_dataclass(dataclass)

@dataclass(msg_id=1)  # The (byte) value 1 identifies this message and must be unique per community
class MyMessage:
    clock: int  # We add an integer (technically a "long long") field "clock" to this message
```

qstokkink:
Fixes #1055

This PR:

 - Updates `doc/basics/overlay_tutorial_5.py` to use dataclass payloads instead of `VariablePayload`.

### DIFF
--- a/doc/basics/overlay_tutorial_5.py
+++ b/doc/basics/overlay_tutorial_5.py
@@ -1,18 +1,19 @@
 import os
 from asyncio import ensure_future, get_event_loop
+from dataclasses import dataclass
 
 from pyipv8.ipv8.community import Community
 from pyipv8.ipv8.configuration import ConfigBuilder, Strategy, WalkerDefinition, default_bootstrap_defs
 from pyipv8.ipv8.lazy_community import lazy_wrapper
-from pyipv8.ipv8.messaging.lazy_payload import VariablePayload, vp_compile
+from pyipv8.ipv8.messaging.payload_dataclass import overwrite_dataclass
 from pyipv8.ipv8_service import IPv8
 
+dataclass = overwrite_dataclass(dataclass)  # Enhance normal dataclasses for IPv8 (see the serialization documentation)
 
-@vp_compile
-class MyMessage(VariablePayload):
-    msg_id = 1  # The byte identifying this message, must be unique per community.
-    format_list = ['I']  # When reading data, we unpack an unsigned integer from it.
-    names = ["clock"]  # We will name this unsigned integer "clock"
+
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

doc/basics/overlay_tutorial_5.py
```diff
@@ -1,27 +1,28 @@
...
+from dataclasses import dataclass
 
 from pyipv8.ipv8.community import Community
 from pyipv8.ipv8.configuration import ConfigBuilder, Strategy, WalkerDefinition, default_bootstrap_defs
 from pyipv8.ipv8.lazy_community import lazy_wrapper
-from pyipv8.ipv8.messaging.lazy_payload import VariablePayload, vp_compile
+from pyipv8.ipv8.messaging.payload_dataclass import overwrite_dataclass
 from pyipv8.ipv8_service import IPv8
 
+dataclass = overwrite_dataclass(dataclass)
```
devos50, line 11: **Question:** I had to lookup why this overwrite was necessary. Since this file is part of a tutorial, is it clear for the reader why this logic is here? If not, I would recommend to add a comment.

devos50 (approved):

InvictusRMC (dismissed):

doc/basics/overlay_tutorial_5.py
qstokkink: Thanks for the suggestion. I'll add a comment to this line.

qstokkink:
validate

qstokkink:
@devos50 I implemented your suggested change. Could you re-review this line?

devos50 (approved):

PR #1080, opened 2022-01-15
Updated exit data rules

### LINKED ISSUE #1079: Datachecker is too strict on extension
**Traceback and/or steps to reproduce**

Related to https://github.com/Tribler/tribler/issues/6720. The terminal of exit nodes is spammed with dropped packets.

---

Packets are dropped by exit nodes because they have the Bittorrent protocol extension `3`, which is not defined in [BEP29](https://www.bittorrent.org/beps/bep_0029.html). This is implemented in our code as follows.

https://github.com/Tribler/py-ipv8/blob/85f7eeeefecf2d6a2d0370494112b7a2c2ddd9f2/ipv8/messaging/anonymization/tunnel.py#L65-L66

However, this actually seems to be a known extension which we can allow (namely `utp_close_reason`): https://github.com/arvidn/libtorrent/blob/95afba30e9acfd8411076a0b05f7ca69bf33b9bb/include/libtorrent/aux_/utp_stream.hpp#L141

To allow this traffic to pass through, we simply need to change the above comment and conditional to allow extensions up to a value of `3`.

qstokkink:
Fixes #1079

This PR:

 - Updates the `DataChecker` to allow uTP packets with extension `3`.
 - Updates the log level of `TunnelExitSocket.is_allowed` to warning, matching the level of `TunnelExitSocket.datagram_received`.

### DIFF
--- a/ipv8/messaging/anonymization/tunnel.py
+++ b/ipv8/messaging/anonymization/tunnel.py
@@ -211,7 +211,7 @@ class TunnelExitSocket(Tunnel, DatagramProtocol, TaskManager):
         if not (is_bt and PEER_FLAG_EXIT_BT in self.overlay.settings.peer_flags) \
            and not (is_ipv8 and PEER_FLAG_EXIT_IPV8 in self.overlay.settings.peer_flags) \
            and not (is_ipv8 and self.overlay._prefix == data[:22]):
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

devos50 (approved):

PR #1086, opened 2022-02-11
Updated support for different coverage versions

### LINKED ISSUE #1085: `create_test_coverage_report.py` incompatible with `coverage>=6.0`
Since `coverage` version `5.0` the `Analysis` class `__init__` signature has changed:

```python
def __init__(self, data, file_reporter):  # <= 4.5.4
def __init__(self, data, file_reporter, file_mapper):  # < 6.0
def __init__(self, data, precision, file_reporter, file_mapper):  # >= 6.0
```

This causes two of our three front-page builders to fail:
screenshot

**Traceback and/or steps to reproduce**

```python
$ python3 create_test_coverage_report.py
  File "create_test_coverage_report.py", line 78, in <module>
    analysis = Analysis(cov.get_data(), file_reporter, abs_file)
TypeError: __init__() missing 1 required positional argument: 'file_mapper'
```

_Admin notice: this issue has been assigned high priority as it directly affects our ability to guarantee quality control. However, this does not cause crashes for users._

qstokkink:
Fixes #1085

This PR:

 - Updates `create_test_coverage_report.py` to support `coverage` versions lower than `5`, between `5` and `6`, and higher than `6`.

Tested with `coverage==5.5` and `coverage==6.3.1`.

### DIFF
--- a/create_test_coverage_report.py
+++ b/create_test_coverage_report.py
@@ -3,6 +3,7 @@ import os
 import pathlib
 import shutil
 import sys
+from distutils.version import LooseVersion
 from io import StringIO
 from unittest import TextTestRunner, defaultTestLoader
 from unittest.suite import TestSuite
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

devos50 (approved):

PR #1095, opened 2022-05-06
Fix _ez_unpack return type

### LINKED ISSUE #1094: Underspecified result type of `EZPackOverlay._ez_unpack_auth` and `EZPackOverlay._ez_unpack_noauth`
Currently, the type of `_ez_unpack_auth` unpacked payload is not understood correctly by PyCharm:

You can see that PyCharm can only see the base Payload methods and not the methods of IntroductionRequestPayload.

In the following PR, I suggest a fix that provides the correct result type:

Also, while fixing that issue, I noticed that currently, DataclassPayload defined as

```python
DataclassPayload = typing.TypeVar('DataclassPayload')
...
AnyPayload = typing.Union[Payload, DataclassPayload]
AnyPayloadType = typing.Union[typing.Type[Payload], typing.Type[DataclassPayload]]
```

In my opinion, it is not correct to use `TypeVar` here. Type variables are for generic functions when the function result type should be derived from the function arguments'. Here, `DataclassPayload` is not used for generics and does not specify that the class is related to data classes. So, as a part of a fix, I'm replacing it with a protocol that checks that the class is actually a data class.

kozlovsky:
Fixes #1094

This PR:

 - Fixes result type of `EZPackOverlay._ez_unpack_auth` and `EZPackOverlay._ez_unpack_noauth`;
 - Fixes the definition of `DataclassPayload` type.

### DIFF
--- a/ipv8/types.py
+++ b/ipv8/types.py
@@ -1,7 +1,17 @@
+import sys
 import typing
 
 Address = typing.Tuple[str, int]
-DataclassPayload = typing.TypeVar('DataclassPayload')
+
+
+if sys.version_info >= (3, 8):
+    class DataclassPayload(typing.Protocol):
+        # Checking for this attribute is currently the most reliable way to ascertain that something is a dataclass.
+        # See https://stackoverflow.com/questions/54668000/type-hint-for-an-instance-of-a-non-specific-dataclass
+        __dataclass_fields__: dict
+else:
+    DataclassPayload = type
+
 
 # pylint: disable=unused-import
 
--- a/ipv8/lazy_community.py
+++ b/ipv8/lazy_community.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from functools import wraps
-from typing import Iterable, List, Tuple, Type, Union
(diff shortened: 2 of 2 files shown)


ipv8/types.py
```diff
@@ -1,7 +1,17 @@
...
-DataclassPayload = typing.TypeVar('DataclassPayload')
+
+
+if sys.version_info >= (3, 8):
+    class DataclassPayload(typing.Protocol):
+        # Checking for this attribute is currently the most reliable way to ascertain that something is a dataclass.
+        # See https://stackoverflow.com/questions/54668000/type-hint-for-an-instance-of-a-non-specific-dataclass
+        __dataclass_fields__: dict
+else:
+    DataclassPayload = type
```
kozlovsky, line 13: As it turns out, protocols were added to `typing` in Python 3.8, so it is necessary to add this workaround for Python 3.7

qstokkink:
This PR has been open for 13 weeks without activity and it lacks sufficient motivation for its proposed changes. I will close this PR until the (equally inactive) discussion in #1094 has been concluded.

PR #1101, opened 2022-07-27
Fixed `create_task` in `DiscreteLoop`

devos50:
This PR:

 - Fixes an issue with the `create_task` method in the discrete event loop. Details: The wrapper inner function in `create_task` in the `DiscreteLoop` is not returning the result of the task. This causes higher-level asyncio routines that depend on this function, such as `gather` and `ensure_future`, to not correctly fire with the right result.


devos50:
validate

qstokkink (approved):

PR #1118, opened 2023-01-24
Fixed circuit retry log message

### LINKED ISSUE #1117: Incorrect log message when creating circuits
`TunnelCommunity.send_initial_create` checks if there's a preexisting retry cache for a circuit and removes it if necessary:

https://github.com/Tribler/py-ipv8/blob/a5aa49d3ae8313b4de774c440de45a966e54f04a/ipv8/messaging/anonymization/community.py#L306-L310

However, the log message should appear when a previous cache was overwritten. Now, the message is incorrectly logged when the opposite happens: when there was no preexisting cache.

These lines should be as follows:

```python
 try: 
     self.request_cache.pop("retry", circuit.circuit_id) 
     self.logger.info("Overwriting existing retry attempt for initial creation of circuit %d", 
                      circuit.circuit_id) 
 except KeyError: 
     pass
```

qstokkink:
Fixes #1117

This PR:

 - Fixes the log message for overwriting an existing circuit retry appearing only when no circuit was overwritten.

For consistency, the inline comment was copied from line 331.

### DIFF
--- a/ipv8/messaging/anonymization/community.py
+++ b/ipv8/messaging/anonymization/community.py
@@ -305,9 +305,10 @@ class TunnelCommunity(Community):
 
         try:
             self.request_cache.pop("retry", circuit.circuit_id)
-        except KeyError:
             self.logger.info("Overwriting existing retry attempt for initial creation of circuit %d",
                              circuit.circuit_id)
+        except KeyError:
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

drew2a (approved):

PR #1121, opened 2023-02-06
WIP: Added Pydantic configuration option

### LINKED ISSUE #1046: Supply Pydantic configuration model
To enable any and all configuration of IPv8, an `IPv8` instance should always be configured through a `dict`. Pydantic is one of the most popular configuration management tools available and it can also create a `dict`. Therefore, Pydantic be used with IPv8. However, programmers have to define their own Pydantic model to fit the IPv8 configuration `dict` to use Pydantic.

To avoid excessive recreation of Pydantic models, IPv8 can supply a pre-made Pydantic model. Of course, to keep IPv8 lightweight and all non-essential dependencies optional, this model should only be supplied if `pydantic` is available on the system running IPv8.

An example that highlights the envisioned usage is as follows.

```python
from ipv8.configuration import ConfigBuilder  # Should be fine if `import pydantic` fails
from ipv8.configuration import IPv8Model   # Errors out if `import pydantic` fails

m = IPv8Model(logger={'level': "ERROR"})

ipv8_instance = IPv8(m.dict())  # Keep the interface clean: IPv8(m) should never be supported
```

qstokkink:
Fixes #1046

This PR:

 - Adds a `pydantic` BaseModel for the IPv8 configuration (including a schema formatter).
 - Adds support for casting `ConfigBuilder` to `dict`.
 - Updates the `IPv8` class to attempt a cast to `dict` if the supplied configuration is not a `dict`.

<details>
<summary>But, does it work for Tribler?</summary>

Sure does!

```
{
    "TriblerConfig": {
        "general": "#/definitions/GeneralSettings",
        "tunnel_community": "#/definitions/TunnelCommunitySettings",
        "bandwidth_accounting": "#/definitions/BandwidthAccountingSettings",
        "bootstrap": "#/definitions/BootstrapSettings",
        "ipv8": "#/definitions/Ipv8Settings",
        "discovery_community": "#/definitions/DiscoveryCommunitySettings",
        "dht": "#/definitions/DHTSettings",
        "trustchain": "#/definitions/TrustchainSettings",
        "watch_folder": "#/definitions/WatchFolderSettings",
        "chant": "#/definitions/ChantSettings",
        "torrent_checking": "#/definitions/TorrentCheckerSettings",
        "libtorrent": "#/definitions/LibtorrentSettings",
        "download_defaults": "#/definitions/DownloadDefaultsSettings",
        "api": "#/definitions/APISettings",
        "resource_monitor": "#/definitions/ResourceMonitorSettings",
        "popularity_community": "#/definitions/PopularityCommunitySettings",
        "remote_query_community": "#/definitions/RemoteQueryCommunitySettings",
        "upgrader_enabled": "boolean",
        "gui_test_mode": "boolean"
    },
    "DiscoveryCommunitySettings": {
        "enabled": "boolean"
    },
    "DownloadDefaultsSettings": {
        "anonymity_enabled": "boolean",
        "number_hops": "integer",
        "safeseeding_enabled": "boolean",
        "saveas": "string",
        "seeding_mode": "#/definitions/SeedingMode",
        "seeding_ratio": "number",
        "seeding_time": "number",
        "channel_download": "boolean",
        "add_download_to_channel": "boolean"
    },
    "SeedingMode": {
        "forever": "forever",
        "never": "never",
        "ratio": "ratio",
        "time": "time"
    },
    "LibtorrentSettings": {
        "enabled": "boolean",
        "port": "integer",
        "proxy_type": "integer",
        "proxy_server": "string",
        "proxy_auth": "string",
        "max_connections_download": "integer",
        "max_download_rate": "integer",
        "max_upload_rate": "integer",
        "utp": "boolean",
        "dht": "boolean",
        "dht_readiness_timeout": "integer",
        "upnp": "boolean",
        "natpmp": "boolean",
        "lsd": "boolean"
    },
    "TrustchainSettings": {
        "ec_keypair_filename": "string",
        "ec_keypair_pubfilename": "string",
        "testnet_keypair_filename": "string",
        "secondary_key_filename": "string",
        "testnet": "boolean"
    },
    "APISettings": {
        "http_enabled": "boolean",
        "http_port": "integer",
        "https_enabled": "boolean",
        "https_port": "integer",
        "https_certfile": "string",
        "key": "string",
        "retry_port": "boolean"
    },
    "TunnelCommunitySettings": {
        "enabled": "boolean",
        "exitnode_enabled": "boolean",
        "random_slots": "integer",
        "competing_slots": "integer",
        "testnet": "boolean",
        "min_circuits": "integer",
        "max_circuits": "integer"
    },
    "ResourceMonitorSettings": {
        "enabled": "boolean",
        "cpu_priority": "integer",
        "poll_interval": "integer",
        "history_size": "integer"
    },
    "DHTSettings": {
        "enabled": "boolean"
    },
    "ChantSettings": {
        "enabled": "boolean",
        "manager_enabled": "boolean",
        "channel_edit": "boolean",
        "channels_dir": "string",
        "testnet": "boolean",
        "queried_peers_limit": "integer",
        "max_mapped_query_peers": "integer"
    },
    "RemoteQueryCommunitySettings": {
        "minimal_blob_size": "integer",
        "maximum_payload_size": "integer",
        "max_entries": "integer",
        "max_query_peers": "integer",
        "max_response_size": "integer",
        "max_channel_query_back": "integer",
        "push_updates_back_enabled": "boolean"
    },
    "Ipv8Settings": {
        "enabled": "boolean",
        "port": "integer",
        "address": "string",
        "bootstrap_override": "string",
        "statistics": "boolean",
        "walk_interval": "number",
        "walk_scaling_enabled": "boolean",
        "walk_scaling_upper_limit": "number"
    },
    "WatchFolderSettings": {
        "enabled": "boolean",
        "directory": "string"
    },
    "BootstrapSettings": {
        "enabled": "boolean",
        "max_download_rate": "integer",
        "infohash": "string"
    },
    "GeneralSettings": {
        "version": "string",
        "log_dir": "string",
        "version_checker_enabled": "boolean",
        "testnet": "boolean"
    },
    "TorrentCheckerSettings": {
        "enabled": "boolean"
    },
    "BandwidthAccountingSettings": {
        "testnet": "boolean",
        "outgoing_query_interval": "integer",
        "max_tx_returned_in_query": "integer"
    },
    "PopularityCommunitySettings": {
        "enabled": "boolean",
        "cache_dir": "string"
    }
}
```

</details>

### DIFF
--- a/ipv8/configuration_pydantic.py
+++ b/ipv8/configuration_pydantic.py
@@ -1,0 +1,402 @@
+"""
+Pydantic configuration BaseModels for IPv8.
+
+If you want to see the available configuration options, print the schema as follows:
+
+::
+
+        from json import dumps
+        print(dumps(format_schema_recursive(IPv8Configuration), indent=4))
+
+|
+You can use the ``IPv8Configuration by simply passing it to an ``IPv8`` constructor:
+
+::
+
+        ipv8_instance = IPv8(IPv8Configuration())
+
+"""
+
+import base64
+import inspect
+import sys
+from enum import Enum
+from typing import Any, Dict, List, Optional, Set, Type, Union
+
+from pydantic import BaseModel, Field, validator
+from pydantic.validators import dict_validator
+
+from .configuration import DISPERSY_BOOTSTRAPPER, default
+from .keyvault.crypto import default_eccrypto
+
+
+class AbstractModel:
+    """
+    Mixin to disallow a model to be instantiated directly.
+
+    This construct is useful for DRY configurations that instantiate objects in some inheritance structure.
+    For example, a ``DuckType`` is not a bird but a ``Pigeon`` and a ``Crow`` are (though they both quack)
+    and this can be reflected when making birds (``make_bird``) by configuration:
+
+    |
+
+    ::
+
+        @dataclass
+        class DuckType(ABC):  # Abstract: a DuckType is not a bird, it just quacks
+            quacks_per_second: int
+
+        @dataclass
+        class Pigeon(DuckType):
+            in_flock: bool  # A group of pigeons is called a flock, not a murder
+
+        @dataclass
+        class Crow(DuckType):
+            in_murder: bool  # A group of crows is called a murder, not a flock
+
+        class DuckTypeConfig(BaseModel, AbstractModel):
+            quacks_per_second: int
+
+        class PigeonConfig(DuckTypeConfig):
+            in_flock: bool
+
+        class CrowConfig(DuckTypeConfig):
+            in_murder: bool
+
+        def make_bird(bird_config: DuckTypeConfig) -> Optional[DuckType]:
+            '''
+            Our composable architecture needs something that quacks per second.
+            The user registers implementations:
+
+            composable_architecture.register(Crow)
+            composable_architecture.register(Pigeon)
+            etc.
+
+            We don't need defensive programming in the composable_architecture
+            if the user can't make a DuckTypeConfig in the first place!
+            '''
+            return composable_architecture.get_implementer(bird_config)
+
+    :returns: The mixin for pydantic BaseModel subclasses.
+    """
+
+    def __new__(cls, *args, **kwargs):
+        """
+        If mixed-in to ``check`` the MRO is [child 1 .. n, check, .. non-AbstractModel .., AbstractModel, object]
+        If there are no children this becomes [check, .. non-AbstractModel .., AbstractModel, object]
+        """
+        check = cls.mro()[1:2] or [object]  # Still works if the mro is empty
+        if not issubclass(check[0], AbstractModel) and check is not AbstractModel:
+            # Try to give the caller some hints.
+            parent_module = sys.modules[cls.__module__]
+            possibilities = [cname for cname, cinstance in inspect.getmembers(parent_module)
+                             if inspect.isclass(cinstance) and cname != "AbstractModel"
+                             and issubclass(cinstance, cls) and cinstance is not cls]
+            raise NotImplementedError(f"{cls} is an abstract model and must be implemented by a subclass! "
+                                      + (f"Perhaps you want one of these: {possibilities}?" if possibilities else ""))
+        return super().__new__(cls)
+
+
+class BaseConfigurationClass(BaseModel):
+    """
+    Augment the ``BaseModel`` class with 1) dict cast unwrapping and 2) inherited validators.
+    """
+
+    def __iter__(self, *args, **kwargs):
+        """
+        Make ``dict(model)`` equal to ``model.dict()``.
+
+        Normally this function only only gives the child model references instead of unwrapping them.
+        """
+        return iter(self.dict().items())
+
+    @classmethod
+    def get_validators(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, value):
+        if isinstance(value, cls):
+            return value
+        else:
+            return cls(**dict_validator(value))
+
+    def dict(self, *args, **kwargs):
+        kwargs['by_alias'] = True
+        return super().dict(*args, **kwargs)
+
+    class Config:
+        allow_population_by_field_name = True
+
+
+class BootstrapperInit(BaseConfigurationClass, AbstractModel):
+    pass
+
+
+class UDPBroadcastBootstrapperInit(BootstrapperInit):
+    bootstrap_timeout: float = 30.0
+
+
+class DispersyBootstrapperInit(BootstrapperInit):
+    ip_addresses: Optional[List[tuple]] = None
+    dns_addresses: Optional[List[tuple]] = None
+    bootstrap_timeout: float = 30.0
+
+    @validator('ip_addresses', always=True)
+    @classmethod
+    def _default_ip_addresses(cls, v):
+        return DISPERSY_BOOTSTRAPPER['init']['ip_addresses'] if v is None else v
+
+    @validator('dns_addresses', always=True)
+    @classmethod
+    def _default_dns_addresses(cls, v):
+        return DISPERSY_BOOTSTRAPPER['init']['dns_addresses'] if v is None else v
+
+
+class Bootstrapper(BaseConfigurationClass, AbstractModel):
+    cls: str = Field(alias='class')
+    init: BootstrapperInit = Field(..., arbitrary_types_allowed=True)
+
+
+class DispersyBootstrapper(Bootstrapper):
+
+    def __init__(self, **kwargs):
+        kwargs["class"] = "DispersyBootstrapper"
+        super().__init__(**kwargs)
+
+
+class UDPBroadcastBootstrapper(Bootstrapper):
+
+    def __init__(self, **kwargs):
+        kwargs["class"] = "UDPBroadcastBootstrapper"
+        super().__init__(**kwargs)
+
+
+class Interface(BaseConfigurationClass):
+    interface: str = default["interfaces"][0]["interface"]
+    ip: str = default["interfaces"][0]["ip"]
+    port: int = default["interfaces"][0]["port"]
+
+
+class Logger(BaseConfigurationClass):
+    level: str = default["logger"]["level"]
+
+    @validator("level", always=True)
+    @classmethod
+    def validate_level(cls, v):
+        allowed = ['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG', 'NOTSET']
+        assert v in allowed, f"Illegal log level {v} specified! Should be one of {allowed}!"
+        return v
+
+
+class Key(BaseConfigurationClass, AbstractModel):
+    alias: str = default["keys"][0]["alias"]
+
+
+class PreloadedKey(Key):
+    bin: str
+
+
+def generate_ephemeral_key():
+    return base64.b64encode(default_eccrypto.generate_key("curve25519").key_to_bin()).decode()
+
+
+class EphemeralKey(Key):
+    bin: str = Field(frozen=True, default_factory=generate_ephemeral_key)
+
+
+class FileKey(Key):
+    generation: str = default["keys"][0]["generation"]
+    file: str = default["keys"][0]["file"]
+
+
+class Walker(BaseConfigurationClass, AbstractModel):
+    strategy: str = Field(...)
+    peers: int = 20
+    init: dict = Field(default_factory=dict)
+
+    @validator("peers", always=True)
+    @classmethod
+    def validate_peer_count(cls, v):
+        assert v >= 0 or v == -1, f"A walker's peer count must be >= 0 or set to -1 (infinite), got: {v}!"
+        return v
+
+
+class RandomWalk(Walker):
+    def __init__(self, **kwargs):
+        kwargs["strategy"] = "RandomWalk"
+        if "init" not in kwargs:
+            kwargs["init"] = {"timeout": 3.0}
+        super().__init__(**kwargs)
+
+
+class EdgeWalk(Walker):
+    def __init__(self, **kwargs):
+        kwargs["strategy"] = "EdgeWalk"
+        super().__init__(**kwargs)
+
+
+class RandomChurn(Walker):
+    def __init__(self, **kwargs):
+        kwargs["strategy"] = "RandomChurn"
+        if "peers" not in kwargs:
+            kwargs["peers"] = -1
+        if "init" not in kwargs:
+            kwargs["init"] = {'sample_size': 8, 'ping_interval': 10.0, 'inactive_time': 27.5, 'drop_time': 57.5}
+        super().__init__(**kwargs)
+
+
+class PeriodicSimilarity(Walker):
+    def __init__(self, **kwargs):
+        kwargs["strategy"] = "PeriodicSimilarity"
+        if "peers" not in kwargs:
+            kwargs["peers"] = -1
+        super().__init__(**kwargs)
+
+
+class PingChurn(Walker):
+    def __init__(self, **kwargs):
+        kwargs["strategy"] = "PingChurn"
+        if "peers" not in kwargs:
+            kwargs["peers"] = -1
+        super().__init__(**kwargs)
+
+
+class Overlay(BaseConfigurationClass):
+    cls: str = Field(alias='class')
+    key: str = default["keys"][0]["alias"]
+    walkers: List[Walker] = [RandomWalk()]
+    bootstrappers: List[Bootstrapper] = [DispersyBootstrapper(init=DispersyBootstrapperInit())]
+    initialize: dict = Field(default_factory=dict)
+    on_start: List[tuple] = []
+
+
+class IPv8Configuration(BaseConfigurationClass):
+    """
+    Main pydantic IPv8 configuration model, can be fed directly into the ``IPv8`` class constructor.
+    """
+    interfaces: List[Interface] = [Interface()]
+    key_aliases: List[Key] = Field([FileKey()], alias="keys")
+    logger: Logger = Logger()
+    working_directory: str = default["working_directory"]
+    walker_interval: float = default["walker_interval"]
+    overlays: List[Overlay] = [
+        Overlay(cls="DiscoveryCommunity", walkers=[RandomWalk(), RandomChurn(), PeriodicSimilarity()]),
+        Overlay(cls="HiddenTunnelCommunity", initialize=default["overlays"][1]["initialize"],
+                on_start=default["overlays"][1]["on_start"]),
+        Overlay(cls="DHTDiscoveryCommunity", walkers=[RandomWalk(), PingChurn()])
+    ]
+
+    @validator("interfaces", always=True)
+    @classmethod
+    def validate_interfaces(cls, v):
+        it_names = [spec.interface for spec in v]
+        assert len(v) == len(set(it_names)), f"Duplicate interface names specified: {', '.join(it_names)}!"
+        return v
+
+
+def format_schema_recursive(*base_models: Type[BaseModel]) -> dict:
+    """
+    Create a schema (``dict``) describing a BaseModel class.
+
+    |
+    For example, the following pretty-prints the schema of ``IPv8Configuration``:
+
+    ::
+
+        from json import dumps
+        print(dumps(format_schema_recursive(IPv8Configuration), indent=4))
+
+    |
+    Note that the schema shows the expected interface, which is not necessarily equal to its implementations!
+    For example, only ``A`` and ``Main`` are shown when calling ``format_schema_recursive(Main)`` using this code:
+
+    ::
+
+        class A(BaseModel):
+            property: number
+
+        class B(A):
+            property2: number
+
+        class Main(BaseModel):
+            a_implementation: A
+
+    In the above example, you could force generation for ``B`` by calling ``format_schema_recursive(Main, B)``.
+
+    :param base_models: the models to generate the schema for
+    :returns: the mapping of types for the given collection of base models
+    """
+    def format_schema_single(model_schema: dict, refs: set) -> Union[dict, list, str]:
+        """
+        Try to convert the types given by pydantic into actual Python objects. Turns objects into dicts, arrays into
+        lists, custom definitions into their BaseModel class names, and keeps whatever other primitive is used.
+
+        Note: this function is recursive!
+
+        :param model_schema: the result of ``<BaseModel>.schema()``
+        :param refs: a reference to a set in which to add encountered references
+        :returns: either a dict describing the schema or a str describing a single primitive type
+        """
+        properties: Union[str, List[Any], Dict[Any, Any]]
+        if "type" in model_schema:
+            is_object = model_schema["type"] in ["object", "array"]
+            name = model_schema["title"] if is_object else model_schema["type"]
+            if "properties" in model_schema:
+                # It's a dict!
+                properties = {k: format_schema_single(v, refs) for k, v in model_schema["properties"].items()}
+            elif "items" in model_schema:
+                # It's a list!
+                if '$ref' in model_schema['items']:
+                    properties = [model_schema['items']['$ref']]
+                    refs.add(model_schema['items']['$ref'])
+                else:
+                    # It's an empty list!
+                    properties = []
+            elif is_object:
+                # It's an empty dict!
+                properties = {}
+            else:
+                # It's a "none of the above or below" (number, string, etc.)!
+                return f"{name.lower()}"
+        else:
+            # It's a custom definition!
+            if '$ref' in model_schema:
+                properties = model_schema["$ref"]
+                refs.add(model_schema["$ref"])
+            elif len(model_schema["allOf"]) == 1:
+                properties = model_schema["allOf"][0]["$ref"]
+                refs.add(model_schema["allOf"][0]["$ref"])
+            else:
+                properties = []
+                for m in model_schema["allOf"]:
+                    properties.append(m["$ref"])
+                    refs.add(m["$ref"])
+        return properties
+
+    finished_refs: Set[str] = set()
+    known_refs = {f"#/definitions/{base_model.__name__}" for base_model in base_models}
+    out = {}
+    while len(finished_refs) != len(known_refs):
+        # 1. Fetch the Python class "model" belonging to the (str) definition name
+        next_ref = list(known_refs - finished_refs)[0]
+        assert next_ref.startswith("#/definitions/")
+        clsname = next_ref[14:]
+        model = None
+        if clsname in globals():
+            # Easy: the class has already been imported.
+            model = globals()[clsname]
+        else:
+            # Hard: we need to find the class definition in the loaded modules.
+            for module in sys.modules:
+                if clsname in dir(sys.modules[module]):
+                    model = getattr(sys.modules[module], clsname)
+                    break
+            # Extra hard: it is still possible to make exotic definitions that escape this search.
+            #             We'll just let the caller manually work around this (e.g., by adding the class to globals).
+        assert model is not None, f"Failed to locate class belonging to {next_ref}!"
+        # 2. Format the model class into a dict
+        out[model.__name__] = ({i.name: i.value for i in model} if issubclass(model, Enum)
+                               else format_schema_single(model.schema(), known_refs))
+        finished_refs.add(next_ref)
+    return out
--- a/ipv8/test/test_configuration_pydantic.py
+++ b/ipv8/test/test_configuration_pydantic.py
@@ -1,0 +1,253 @@
+import base64
+import enum
+import importlib
+import unittest
+
+from .base import TestBase
+from ..configuration import default
+
+try:
+    importlib.import_module("pydantic")
+
+    SKIP_TEST_CASE = False
+except ImportError:
+    SKIP_TEST_CASE = True
+else:
+    from ..configuration_pydantic import (BaseConfigurationClass, Bootstrapper, BootstrapperInit, DispersyBootstrapper,
+                                          DispersyBootstrapperInit, EdgeWalk, EphemeralKey, FileKey, IPv8Configuration,
+                                          Interface, Key, Overlay, PeriodicSimilarity, PingChurn, PreloadedKey,
+                                          RandomChurn, RandomWalk, UDPBroadcastBootstrapper,
+                                          UDPBroadcastBootstrapperInit, Walker, format_schema_recursive)
+
+
+class CustomColor(enum.Enum):
+    """
+    Note ``Color`` is an illegal name and will be overwritten by pydantic!
+    """
+    RED = 1
+    GREEN = 2
+    BLUE = 3
+
+
+@unittest.skipIf(SKIP_TEST_CASE, "Skipping optional pydantic tests because the library is not installed.")
+class TestConfigurationPydantic(TestBase):
+    """
+    Tests to check IPv8 pydantic integration.
+
+    Note that this TestCase tests IPv8<->pydantic integration, not if pydantic works as promised.
+    """
+    FAKE_KEY_BIN = base64.b64encode(b"0" * 74).decode()
+
+    def test_abstract_key(self):
+        """
+        Check if ``Key`` models are not allowed to be instantiated.
+        """
+        self.assertRaises(NotImplementedError, Key)
+        self.assertRaises(NotImplementedError, Key, alias="keyname")
+
+    def test_implemented_keys(self):
+        """
+        Check if the child classes of ``Key`` are allowed to be instantiated.
+        """
+        keys = [
+            EphemeralKey(),
+            PreloadedKey(bin=TestConfigurationPydantic.FAKE_KEY_BIN),
+            FileKey()
+        ]
+
+        for key in keys:
+            self.assertIsInstance(key, Key)
+
+    def test_inherit_field_keys(self):
+        """
+        Check if the child classes of ``Key`` inherit the ABC's fields.
+        """
+        named_keys = [
+            EphemeralKey(alias="keyname"),
+            PreloadedKey(alias="keyname", bin=TestConfigurationPydantic.FAKE_KEY_BIN),
+            FileKey(alias="keyname")
+        ]
+
+        for key in named_keys:
+            self.assertEqual("keyname", key.alias)
+
+    def test_abstract_walker(self):
+        """
+        Check if ``Walker`` models are not allowed to be instantiated.
+        """
+        self.assertRaises(NotImplementedError, Walker)
+        self.assertRaises(NotImplementedError, Walker, strategy="strategyname")
+
+    def test_implemented_walkers(self):
+        """
+        Check if the child classes of ``Walker`` are allowed to be instantiated.
+        """
+        walkers = [
+            RandomWalk(),
+            EdgeWalk(),
+            RandomChurn(),
+            PeriodicSimilarity(),
+            PingChurn()
+        ]
+
+        for walker in walkers:
+            self.assertIsInstance(walker, Walker)
+
+    def test_inherit_field_walkers(self):
+        """
+        Check if the child classes of ``Walker`` inherit the ABC's fields.
+        """
+        walkers_peers = [
+            RandomWalk(peers=42),
+            EdgeWalk(peers=42),
+            RandomChurn(peers=42),
+            PeriodicSimilarity(peers=42),
+            PingChurn(peers=42)
+        ]
+
+        for walker in walkers_peers:
+            self.assertEqual(42, walker.peers)
+            self.assertEqual(walker.__class__.__name__, walker.strategy)
+
+    def test_abstract_bootstrapper(self):
+        """
+        Check if ``Bootstrapper`` models are not allowed to be instantiated.
+        """
+        self.assertRaises(NotImplementedError, Bootstrapper)
+        self.assertRaises(NotImplementedError, Bootstrapper, kwargs={"class": "boostrappername", "init": {}})
+
+    def test_implemented_bootstrappers(self):
+        """
+        Check if the child classes of ``Bootstrapper`` are allowed to be instantiated.
+        """
+        bootstrappers = [
+            DispersyBootstrapper(init=DispersyBootstrapperInit()),
+            UDPBroadcastBootstrapper(init=UDPBroadcastBootstrapperInit())
+        ]
+
+        for bootstrapper in bootstrappers:
+            self.assertIsInstance(bootstrapper, Bootstrapper)
+
+    def test_inherit_field_bootstrappers(self):
+        """
+        Check if the child classes of ``Bootstrapper`` inherit the ABC's fields.
+        """
+        bootstrappers = [
+            DispersyBootstrapper(init=DispersyBootstrapperInit()),
+            UDPBroadcastBootstrapper(init=UDPBroadcastBootstrapperInit())
+        ]
+
+        for bootstrapper in bootstrappers:
+            self.assertEqual(bootstrapper.__class__.__name__, bootstrapper.cls)
+
+    def test_abstract_bootstrapper_init(self):
+        """
+        Check if ``BootstrapperInit`` models are not allowed to be instantiated.
+        """
+        self.assertRaises(NotImplementedError, BootstrapperInit)
+
+    def test_default_dict_direct(self):
+        """
+        Check if directly calling the ``dict()`` method on the base model returns the default dict.
+        """
+        self.assertDictEqual(default, IPv8Configuration().dict())
+
+    def test_default_dict_cast(self):
+        """
+        Check if casting the base model to ``dict`` returns the default dict.
+        """
+        self.assertDictEqual(default, dict(IPv8Configuration()))
+
+    def test_format_schema_single(self):
+        """
+        Check if the schema for a single BaseModel is correctly generated.
+        """
+        expected = {
+            "Interface": {
+                "interface": "string",
+                "ip": "string",
+                "port": "integer"
+            }
+        }
+        self.assertDictEqual(expected, format_schema_recursive(Interface))
+
+    def test_format_schema_double(self):
+        """
+        Check if the schema for two BaseModels is correctly generated.
+        """
+        expected = {
+            "Interface": {
+                "interface": "string",
+                "ip": "string",
+                "port": "integer"
+            },
+            "Key": {
+                "alias": "string"
+            }
+        }
+        self.assertDictEqual(expected, format_schema_recursive(Interface, Key))
+
+    def test_format_schema_abstract(self):
+        """
+        Check if the schema for a single BaseModel that inherits from an ABC is correctly generated.
+        """
+        expected = {
+            "FileKey": {
+                "alias": "string",  # Inherited from Key, not defined in FileKey itself
+                "generation": "string",
+                "file": "string"
+            }
+        }
+        self.assertDictEqual(expected, format_schema_recursive(FileKey))
+
+    def test_format_schema_nested(self):
+        """
+        Check if the schema for a single BaseModel that nests another BaseModel is correctly generated.
+        """
+        expected = {
+            "UDPBroadcastBootstrapper": {
+                "class": "string",
+                "init": "#/definitions/BootstrapperInit"
+            },
+            "BootstrapperInit": {}
+        }
+        self.assertDictEqual(expected, format_schema_recursive(UDPBroadcastBootstrapper))
+
+    def test_format_schema_nested_list(self):
+        """
+        Check if the schema for a single BaseModel that nests a list of other BaseModels is correctly generated.
+        """
+        expected = {
+            "BootstrapperInit": {},
+            "Bootstrapper": {"class": "string", "init": "#/definitions/BootstrapperInit"},  # test_format_schema_nested
+            "Walker": {"init": {}, "peers": "integer", "strategy": "string"},
+            "Overlay": {
+                "bootstrappers": ["#/definitions/Bootstrapper"],  # List of nested BaseModels
+                "class": "string",
+                "initialize": {},
+                "key": "string",
+                "on_start": [],
+                "walkers": ["#/definitions/Walker"]  # List of nested BaseModels
+            }
+        }
+        self.assertDictEqual(expected, format_schema_recursive(Overlay))
+
+    def test_format_schema_enum(self):
+        """
+        Check if schemas for Enums (also allowed inside of pydantic BaseModels) are correctly generated.
+
+        Note: we don't want to perform pydantic imports in this module, so we perform some `globals()` magic instead.
+        """
+        class MockEnumModel(BaseConfigurationClass):
+            color: CustomColor
+        globals()["MockEnumModel"] = MockEnumModel
+
+        expected = {
+            "MockEnumModel": {"color": "#/definitions/CustomColor"},
+            "CustomColor": {
+                "RED": 1,
+                "GREEN": 2,
+                "BLUE": 3
+            }
+        }
+        self.assertDictEqual(expected, format_schema_recursive(MockEnumModel))
--- a/ipv8_service.py
+++ b/ipv8_service.py
@@ -7,6 +7,7 @@ from contextlib import suppress
 from os.path import isfile
 from threading import RLock
 from traceback import format_exception
+from typing import Dict, Type, Union
 
 if hasattr(sys.modules['__main__'], "IPv8"):
     sys.modules[__name__] = sys.modules['__main__']
@@ -28,6 +29,7 @@ else:
         from ipv8.peerdiscovery.discovery import EdgeWalk, RandomWalk
         from ipv8.peerdiscovery.network import Network
         from ipv8.dht.discovery import DHTDiscoveryCommunity
+        from ipv8.types import ConfigBuilder, DictCastable, Endpoint, IPv8Configuration, Overlay
         from ipv8.util import maybe_coroutine
     else:
         from .ipv8.messaging.interfaces.statistics_endpoint import StatisticsEndpoint
@@ -46,6 +48,7 @@ else:
         from .ipv8.peerdiscovery.discovery import EdgeWalk, RandomWalk
         from .ipv8.peerdiscovery.network import Network
         from .ipv8.dht.discovery import DHTDiscoveryCommunity
+        from .ipv8.types import ConfigBuilder, DictCastable, Endpoint, IPv8Configuration, Overlay
         from .ipv8.util import maybe_coroutine
 
     _COMMUNITIES = {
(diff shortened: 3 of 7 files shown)


qstokkink:
validate

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,357 @@
...
+        class B(A):
+            property2: number
+
+        class Main(BaseModel):
+            a_implementation: A
+
+    In the above example, you could force generation for ``B`` by calling ``format_schema_recursive(Main, B)``.
+
+    :param base_models: the models to generate the schema for
+    :returns: the mapping of types for the given collectiton of base models
```
drew2a, line 283: NIT: typo "collectiton"

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,357 @@
...
+
+    @validator("interfaces", always=True)
+    @classmethod
+    def validate_interfaces(cls, v):
+        it_names = [spec.interface for spec in v]
+        assert len(v) == len(set(it_names)), f"Duplicate interface names specified: {', '.join(it_names)}!"
+        return v
+
+
+def format_schema_recursive(*base_models: Type[BaseModel]) -> dict:
```
drew2a, line 253: Thank you for this method. It is quite helpful for the generation of documentation.

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,357 @@
...
+from pydantic import BaseModel, Field, validator
+from pydantic.validators import dict_validator
+
+from .configuration import DISPERSY_BOOTSTRAPPER, default
+from .keyvault.crypto import default_eccrypto
+
+
+class AbstractModel:
+    """
+    Mixin to disallow a model to be instantiated directly.
```
drew2a, line 35: NIT: the purpose of this mixin  is not clear. 

The missed piece of documentation is the answer to the question "why it is necessary to disallow a model to be instantiated directly?"

qstokkink:
validate

ipv8/configuration_pydantic.py
qstokkink: Thanks for the feedback 👍 I added some documentation (see below) that exemplifies why you would want this to maintain a beautiful* DRY and composable architecture that tightly ties configuration models to classes.

*Disclaimer: in my opinion.

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,357 @@
...
+class Overlay(BaseConfigurationClass):
+    cls: str = Field(alias='class')
+    key: str = default["keys"][0]["alias"]
+    walkers: List[Walker] = [RandomWalk()]
+    bootstrappers: List[Bootstrapper] = [DispersyBootstrapper(init=DispersyBootstrapperInit())]
+    initialize: dict = Field(default_factory=dict)
+    on_start: List[tuple] = []
+
+
+class IPv8Configuration(BaseConfigurationClass):
```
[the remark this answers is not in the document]
qstokkink: Thanks! The problem is a bit worse though: this is already the third configuration option (with `dict` and `ConfigBuilder` being the other two). I agree that one option should be leading to avoid duplication and for this IPv8 uses the `dict` as the leading config. The reasoning for this is that _(a)_ `dict` requires no external dependencies, allowing all other options to be optional (and less hard dependencies is better for just about every metric), _(b)_ most configuration libraries can convert to a `dict` relatively easily, _(c)_ pulling from the common source `dict` does not require one optional configuration library to depend on importing another (now non-optional) configuration library that is otherwise unused.

The IPv8 ethos is to support everything for everyone as much as possible. IPv8 does not care about who (or what) calls it and whether they think XML is superior to pydantic or JSON. It should all _just work_, no matter what your preference may be.

ipv8_service.py
```diff
@@ -71,37 +71,37 @@ class IPv8:
 
         def __init__(self, configuration, endpoint_override=None, enable_statistics=False, extra_communities=None):
```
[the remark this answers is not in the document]
qstokkink: I adopted your suggestion. I also added the other configuration options to the type.

qstokkink:
@drew2a thanks for reviewing! I addressed all of your comments and I described how I resolved them in their replies. Please have another look at them to see if you agree with the presented resolutions.

ipv8/configuration_pydantic.py
drew2a: Thank you for this explanation. Now your intention is clear.

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
drew2a: > I agree that one option should be leading to avoid duplication and for this IPv8 uses the dict as the leading config.

Not sure this is true. `pydantic` model could not be created from "leading" dict. That means it does not avoid duplication but produces it.

As far as I understand the code, the description of `ConfigBuilder`'s fields also could not be created from the `dict`, which doesn't make the `dict` a single source of the configuration's description.

If for changing/adding a field to the configuration, you should add it in three places, then it shows there is a problem in the internal configs implementation. 
This implementation adds one more possibility for future developers of ipv8 to make a mistake while changing configs.

This sentence is unclear to me:
> The IPv8 ethos is to support everything for everyone as much as possible. IPv8 does not care about who (or what) calls it and whether they think XML is superior to pydantic or JSON. It should all just work, no matter what your preference may be.

As a library, `IPv8` should provide a clear API for users, without keeping in mind any ethos. That means it should be clear for users which options `IPv8` has for pushing a configuration to it.

However, this reasoning is irrelevant to the problem described above, as the problem is with the internal config organization rather than the user API.

ipv8/configuration_pydantic.py
drew2a: Despite the good intention, I don't see a necessity of using `AbstractModel` here.

The `IPv8` configuration is not designed to be "extendable" by users. This means there is no need to artificially prohibit instantiations of some classes. It just doesn't add any value to the code. But it adds complexity.

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
qstokkink: > As far as I understand the code, the description of ConfigBuilder's fields also could not be created from the dict, which doesn't make the dict a single source of the configuration's description.

This is incorrect. In fact, the `ConfigBuilder` IS just that same dict, there is no other state:

https://github.com/Tribler/py-ipv8/blob/ce53a318c6e36846f19dd92d39b9014c90516255/ipv8/configuration.py#L193-L195

> This implementation adds one more possibility for future developers of ipv8 to make a mistake while changing configs.

This is true and it is the reason I never added support for pydantic before. However, pydantic appears to be popular.

> As a library, IPv8 should provide a clear API for users, without keeping in mind any ethos. That means it should be clear for users which options IPv8 has for pushing a configuration to it.

Ethos is extremely important. Different domains have very different APIs. For example, High-Performance Computing libraries have vastly different API looks/feels/controls/outlooks than networking APIs or HTTP APIs: you would "never" find a REST API when using HPC or OS libaries. The clarity of use comes with the expertise and domain of its users (networking/distributed systems engineers in this case).

I agree with the latter half of your statement, regarding the code showing what options for configuration exist, which is why I explicitly specified the permissible types that can serve as a configuration.

qstokkink:
@drew2a I don't believe the two unresolved issues merit futher discussion and I've highlighted why in their respective threads. I've left the discussions unresolved so the discussions don't get hidden.

ipv8/test/test_configuration_pydantic.py
```diff
@@ -0,0 +1,250 @@
...
+try:
+    from ..configuration_pydantic import (BaseConfigurationClass, Bootstrapper, BootstrapperInit, DispersyBootstrapper,
+                                          DispersyBootstrapperInit, EdgeWalk, EphemeralKey, FileKey, IPv8Configuration,
+                                          Interface, Key, Overlay, PeriodicSimilarity, PingChurn, PreloadedKey,
+                                          RandomChurn, RandomWalk, UDPBroadcastBootstrapper,
+                                          UDPBroadcastBootstrapperInit, Walker, format_schema_recursive)
+
+    SKIP_TEST_CASE = False
+except ImportError:
+    SKIP_TEST_CASE = True
```
drew2a, line 17: This try-except block looks incorrect to me.

The intention of this block (as far as I understand) is skipping the tests in the case `pydantic` library is not installed.

In this case, the test suite prints the following message:

```
SKIPPED: Skipping optional pydantic tests because the library is not installed.
```

But a user will see the same error in case any of the import attempts fail. This behavior will hide the real problem.

The correct behavior here is to try to import something from the `pydantic` library and in the case of an error, sets  `SKIP_TEST_CASE = True`
```python

try:
    from pydantic import BaseModel

    SKIP_TEST_CASE = False
except ImportError:
    SKIP_TEST_CASE = True

from ..configuration_pydantic import (BaseConfigurationClass, Bootstrapper, BootstrapperInit, DispersyBootstrapper,
                                      DispersyBootstrapperInit, EdgeWalk, EphemeralKey, FileKey, IPv8Configuration,
                                      Interface, Key, Overlay, PeriodicSimilarity, PingChurn, PreloadedKey,
                                      RandomChurn, RandomWalk, UDPBroadcastBootstrapper, UDPBroadcastBootstrapperInit,
                                      Walker, format_schema_recursive)

```

ipv8/configuration_pydantic.py
drew2a: > That is an interesting statement to make. It definitely is designed to be just that: I would know because I designed it.

I mean that the typical (and maybe the single one) use case of the provided `pydanting` config is the following:
1. A user creates the `IPv8Configuration` class provided by `IPv8`
2. The user customizes the values of this class
3. The user passes the config to the `IPv8` constructor 

Eg:
```python
from ipv8.configuration_pydantic import IPv8Configuration
from ipv8_service import IPv8

config = IPv8Configuration()
config.walker_interval = 60
ipv8 = IPv8(config)
```

It is hard to imagine that a user will try to redefine (extend) some classes inside `IPv8Configuration` having inherited them from `AbstractModel`.
Even if a user wants to redefine classes, most likely they will inherit them from the child class eg `DispersyBootstrapperInit` rather than `BootstrapperInit`.


> Perhaps downstream projects may not want to reuse it. However, the code is used in this file so it definitely objectively provides value here.

Yes, the code is used in this file, agree. But `AbstractModel` is not needed here. You can safely remove it. The value of the code would be the same.

This is the `configuration_pydantic.py` without using `AbstractModel` (the same value, less code, less complexity):
https://gist.github.com/drew2a/3168571b2051cd49cd49032498c64c00


Downstream code in both cases (with using `AbstractModel`  and without) will be the same.

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
drew2a: > This is incorrect. In fact, the ConfigBuilder IS just that same dict, there is no other state.

Well. Try to rename the field `walker_interval` and you will see that you have to do it at least in three places. In the `dict`, in the `ConfigBuilder` and in the `pydantic` configuration. This example shows that the config definition is duplicated across the codebase.

As a counterargument, you will probably say that the `walker_interval` name would never be changed because the config must be backward compatible. It is true, but it is not the counterargument to the statement that "config definition is duplicated across the codebase".

As a less simple, but more realistic example: imagine a developer that wants to add a new field to the config. They should change at least three different places in IPv8 codebase:  the `dict`, in the `ConfigBuilder` and in the `pydantic`.

> This implementation adds one more possibility for future developers of ipv8 to make a mistake while changing configs.
This is true and it is the reason I never added support for pydantic before. However, pydantic appears to be popular.

I completely agree with adding support for pydantic. I'm only questioning the way it is implemented right now.

As an approach that could decrease code duplication, I suggest making `pydantic` the "main" place when the config is defined. It is not hard and it is backward compatible.

drew2a (changes requested):
If we are unable to come to a consensus, we could consider seeking a second opinion (@kozlovsky, @xoriole) to help inform our decision 🌱

ipv8/test/test_configuration_pydantic.py
drew2a: You are right.
This is the working example (tested with `pydantic`, without `pydantic` and with an import error):

```python
try:
    from pydantic import BaseModel

    SKIP_TEST_CASE = False
except ImportError:
    SKIP_TEST_CASE = True
else:
    from ..configuration_pydantic import (BaseConfigurationClass, Bootstrapper, BootstrapperInit, DispersyBootstrapper,
                                          DispersyBootstrapperInit, EdgeWalk, EphemeralKey, FileKey, IPv8Configuration,
                                          Interface, Key, Overlay, PeriodicSimilarity, PingChurn, PreloadedKey,
                                          RandomChurn, RandomWalk, UDPBroadcastBootstrapper,
                                          UDPBroadcastBootstrapperInit,
                                          Walker, format_schema_recursive)
```

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
qstokkink: > As a counterargument, you will probably say that the walker_interval name would never be changed because the config must be backward compatible.

No I wouldn't. I'd argue that `walker_interval` is the setting name used throughout the entire project. The names of (most of) the settings directly relate to the parameters of classes in the project. Reducing duplication in hard-coded _values_ is good. However, reducing duplication in the _keys_ of the configuration is impossible. You cannot have two configurations set a different key for the same configurable class. I can provide examples if you like.

> It is not hard and it is backward compatible.

It makes `pydantic` non-optional. I can provide crash logs if you give me an implementation.

ipv8/configuration_pydantic.py
qstokkink: > You can safely remove it. 

No, you can't. Unless you assume that users only construct configurations that are already valid, like you appear to argue for. However, then why use pydantic validation in the first place?

qstokkink:
@drew2a I have one simple issue and that is: if I implement your suggestions, the code crashes.

I don't mind debates on styles/ideology/etc. with second opinions but a crash is a crash.

ipv8/test/test_configuration_pydantic.py
qstokkink: Done 👍

ipv8/configuration_pydantic.py
drew2a: Could you provide a realistic example of how users will use the `IPv8` library with the invalid configuration?

> However, then why use pydantic validation in the first place?

If I understand you correctly, `pydantic` validation is must-have for validating actual config values: https://docs.pydantic.dev/usage/validators/

It has nothing in common with `AbstactModel` because `AbstactModel` is aimed to "validate" the class hierarchy.

ipv8/test/test_configuration_pydantic.py
qstokkink: That causes a Flake8 error: https://jenkins-ci.tribler.org/job/ipv8/job/test_ipv8_PR_flake8_check/1493//artifact/output/ipv8.test.test_configuration_pydantic.report.html

drew2a: Thanks!

drew2a: > F401: 'pydantic.BaseModel' imported but unused

Which is true. I'll check how to avoid or ignore this issue.

qstokkink: I used importlib instead. It should do the same, without triggering Flake8.

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
qstokkink: "Dependency hell" is something that goes back decades. People have spent years of development time on refactoring to newer and hipper libraries (e.g., https://github.com/Tribler/tribler/issues/68). Better alternatives get discovered, support gets dropped, and external libraries turn out to not work quite as well on all operating systems as years go by. Keeping external libraries optional, if they can reasonably be optional, saves a lot of work in the long run.

Even if pydantic is a safe bet now, I don't think it's unreasonable to expect pydantic to be dethroned in the future, by [Gin](https://github.com/google/gin-config) for example. Future generations will curse our names if we go all-in when we don't have to. Therefore, it seems prudent to keep pydantic optional.

ipv8/configuration_pydantic.py
qstokkink: > Could you provide a realistic example of how users will use the IPv8 library with the invalid configuration?

Sure: suppose a user runs `format_schema_single()` to discover the pydantic API. The format will show that a `Walker` configuration is necessary. The first thing a user would then reasonable try is to instantiate such a `Walker`:

```python
>> Walker()
  File "C:\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\py-ipv8\ipv8\configuration_pydantic.py", line 405, in <module>
    Walker()
  File "C:\py-ipv8\ipv8\configuration_pydantic.py", line 95, in __new__
    raise NotImplementedError(f"{cls} is an abstract model and must be implemented by a subclass! "
NotImplementedError: <class '__main__.Walker'> is an abstract model and must be implemented by a subclass! Perhaps you want one of these: ['EdgeWalk', 'PeriodicSimilarity', 'PingChurn', 'RandomChurn', 'RandomWalk']?
```

The user is immediately greeted with all known valid subclasses automatigically. Very user-friendly.

I agree with your point. However, I consider both of the elements that you mention to be part of creating the configuration.

qstokkink:
validate

ipv8/configuration_pydantic.py
drew2a: > Sure: suppose a user runs format_schema_single() to discover the pydantic API. The format will show that a Walker configuration is necessary.
The user is immediately greeted with all known valid subclasses automagically. Very user-friendly.

Ok. I guess now I started to understand your reasoning 👍 . 

Basically, you use `AbstactModel` for helping the user to create a valid non-default config, if we assume he extensively uses `format_schema_recursive`/`format_schema_single`.

Agre that your example is pretty valid.

---

However, I doubt that this is a realistic scenario. For example, when I will construct a config I never use `format_schema_recursive` (because I'll go for documentation first). 

When I'll see that there is a `Walker` class, I never use `format_schema_single`, because I ask PyCharm for subclasses instead:
image

It is much easier and user-friendly.

`format_schema_recursive` is a quite good function, and I like it pretty much. But I see that the main user of this function is a library author. He uses this function to generate documentation. This function is not for extensive use by regular library users.

qstokkink: I don't agree with your assumption of the behavior of average users, which relates to the following two statements you made:

> (because I'll go for documentation first)

Where would you get documentation? User models can come from anywhere in the runtime: there is no central documentation (nor can there be any - even if you want docstrings, you would have to know about the classes beforehand).

> When I'll see that there is a Walker class, I never use format_schema_single, because I ask PyCharm for subclasses instead

How do you know that `Walker` needs to be subclassed? `pydantic` offers no hints that a user should do this and `Walker` can be instantiated without any warnings.

drew2a: > That would be absolutely correct, the defaults are indeed documented. The big caveat is this: these are only the default walkers.

Agree. 
I meant not Walker's documentation, but IPv8 configuration documentation. It should be there and it should describe the documentation format `IPv8` expect from a user.

>Absolutely! IPv8 users are (this is a slight repetition of an earlier point I made) distributed systems and network engineers. Some real examples of their use cases are these thesis projects: Tribler/tribler#2623, Tribler/tribler#2754. Specifically for custom walking, I would bring up Tribler's TunnelCommunity and the GigaChannelCommunity. More example use cases are pretty much every single Gumby experiment.

I check the links and I don't see any code that uses "models that come from anywhere". I'm pretty sure for all these projects there are no necessities and no profit from `AbstactModel`'s introduction.

> Why not? As a user that doesn't know anything, I would expect a configuration that takes a Walker instance to accept a walker instance that walks.

Fair point. But for a user  that doesn't know anything I would expect the behavior described above:

```python
from ipv8_service import IPv8

config = IPv8Configuration()
config.walker_interval = 60
ipv8 = IPv8(config)
```

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
drew2a: Your concerns are understandable and valid. 
But I doubt that the `pydantic` dependency is a correct example of the "dependency hell" phenomenon.

> Future generations will curse our names if we go all-in when we don't have to.

Agree, but I don't think this is correct in regard to `pydantic`. Even if Gin dethrones it in the future, this will not necessarily lead to the need of rewriting the codebase. 
However, even if you have to replace  `pydantic` with something else, it is ok as long as `pydantic` configuration remains clean and simple. This is one more point to remove unnecessary and complex `AbstactModel`. 

Future generations will curse our names if they should rewrite the unreasonably complicated configuration.

ipv8/configuration_pydantic.py
qstokkink: Alright, let me first enumerate what we agree on then:

 1. Everything that can be documented should be documented (and it already is).
 2. There are no necessity to use `AbstactModel`.

Let me know if I got that wrong. Assuming we do agree on at least that, then let's drill down to what we still disagree on:

 - "There is no profit from using `AbstactModel`": I disagree with this. At the very least, the class hierarchy declares that a class should not be instantiated directly. That is some value, at least to me. Now, I do not wish to argue that _you_ in particular may not have any use for it because you use PyCharm and read documentation and I would also like to clarify that this code _does not force you to use it_. However, other people, like myself, do benefit from this construction and if it doesn't hurt you, why disallow other people from using it?

The example you give will work but `walker_interval` is not a model that inherits from `AbstractModel`. Therefore, I don't know how that relates to the discussion. So, I'm still not sure if I agree or disagree - I don't understand your argument.

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
qstokkink: I do agree that code should not be unreasonably complex.

My test for dependency hell is the following question: "Can the dependency reasonably be made optional". The rest of my flow chart is as follows:

 - **No:** restructure the code and use it everywhere.
 - **Yes:** make it optional.

I think our main disagreement here is whether it is reasonable to make pydantic optional. Considering the fact that this is controlled by a single import statement (`import configuration_pydantic`) I don't see any spectacular or unreasonable complexity. I will agree that it makes a singular unit test case need a single if-statement and that is more complex than having no if-statement.

Let's keep the `AbstractModel` discussion in the other thread, or we will derail this one.

ipv8/configuration_pydantic.py
drew2a: > Alright, let me first enumerate what we agree on then

Yes, you are right!

To be honest I would like to see this fragment also in the documentation (but maybe it is not worth it regarding maintenance cost (somebody should keep it actual):

```
{
    "IPv8Configuration": {
        "interfaces": [
            "#/definitions/Interface"
        ],
        "keys": [
            "#/definitions/Key"
        ],
        "logger": [
            "#/definitions/LogLevel"
        ],
        "working_directory": "string",
        "walker_interval": "number",
        "overlays": [
            "#/definitions/Overlay"
        ]
    },
    "Key": {
        "alias": "string"
    },
    "Overlay": {
        "class": "string",
        "key": "string",
        "walkers": [
            "#/definitions/Walker"
        ],
        "bootstrappers": [
            "#/definitions/Bootstrapper"
        ],
        "initialize": {},
        "on_start": []
    },
    "Bootstrapper": {
        "class": "string",
        "init": [
            "#/definitions/BootstrapperInit"
        ]
    },
    "BootstrapperInit": {},
    "LogLevel": {
        "level": "string"
    },
    "Interface": {
        "interface": "string",
        "ip": "string",
        "port": "integer"
    },
    "Walker": {
        "strategy": "string",
        "peers": "integer",
        "init": {}
    }
}
```

>  However, other people, like myself, do benefit from this construction and if it doesn't hurt you, why disallow other people from using it?

Because all code that has been written should be maintained. That is why you should avoid writing unnecessary code. Most likely somebody in the future has to fix/rewrite it. 

The code you add is tricky:
```python
class AbstractModel:
    def __new__(cls, *args, **kwargs):
        """
        If mixed-in to ``check`` the MRO is [child 1 .. n, check, .. non-AbstractModel .., AbstractModel, object]
        If there are no children this becomes [check, .. non-AbstractModel .., AbstractModel, object]
        """
        check = cls.mro()[1:2] or [object]  # Still works if the mro is empty
        if not issubclass(check[0], AbstractModel) and check is not AbstractModel:
            # Try to give the caller some hints.
            parent_module = sys.modules[cls.__module__]
            possibilities = [cname for cname, cinstance in inspect.getmembers(parent_module)
                             if inspect.isclass(cinstance) and cname != "AbstractModel"
                             and issubclass(cinstance, cls) and cinstance is not cls]
            raise NotImplementedError(f"{cls} is an abstract model and must be implemented by a subclass! "
                                      + (f"Perhaps you want one of these: {possibilities}?" if possibilities else ""))
        return super().__new__(cls)
```

It is hard to maintain code like this. 
I'm not against tricky code. I'm just against unnecessary tricky code.

> The example you give will work but walker_interval is not a model that inherits from AbstractModel. Therefore, I don't know how that relates to the discussion.

I provide this as an example of the most probable (from my POV) `IPv8` config use case. In particular, the use case is the following:
1. Create a default config
2. Change it
3. Pass it to the `IPv8` class constructor

It is not an argument. It is an attempt of switching from imaginary no-code examples to more concrete examples. Therefore it will be easier to discuss them.

ipv8/configuration_pydantic.py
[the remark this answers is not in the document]
drew2a: > I think our main disagreement here is whether it is reasonable to make pydantic optional

Yes, I guess so :)

ipv8/configuration_pydantic.py
qstokkink: > Because all code that has been written should be maintained. That is why you should avoid writing unnecessary code. 

That I agree with. However, you apparently do not share the same problem that causes the necessity. The code you cite is not that tricky at all, [it is very close to the minimal example of disallowed subclassing](https://stackoverflow.com/a/16056691). Unless we get Python 4, this small snippet will probably not need maintenance quickly. You have my lifetime guarantee that I will maintain this code, if you are confident that it is too tricky to maintain.

> It is not an argument. It is an attempt of switching from imaginary no-code examples to more concrete examples. Therefore it will be easier to discuss them.

Alright. However, you're not using `AbstractModel` in this new example. Could we discuss a different example that uses `AbstractModel` instead?

qstokkink: > Everyone from the team is able to maintain this class. The problem is that the codebase of Tribler and IPv8 is huge.

Ok. Good to hear that maintainability is not an issue for this class: let's switch to the next problem of code base size then. Code base size is only an issue if there is high coupling between class. This class is fully optional, even within this module/file, so there is no coupling and, therefore, no issue.

> That's the point, I find it hard to imagine an example that would demonstrate the need for the AbstractModel.

I see that from your example: you're not using any customization. Let me give a complete and runnable example/problem then. Your challenge is to add a configuration that launches `MyWalker` with `foo=13` in the following code snippet:

```python
from asyncio import ensure_future, get_event_loop
from ipv8.community import Community
from ipv8.configuration_pydantic import IPv8Configuration, Overlay, Walker
from ipv8.peerdiscovery.discovery import DiscoveryStrategy
from ipv8_service import _WALKERS, IPv8

class MyOverlay(Community):
    community_id = b'\xFF' * 20

    def do_foo(self, foo: int):
        print(f"[{foo}] FOO!")

class MyWalker(DiscoveryStrategy):
    def __init__(self, overlay, foo: int):
        super().__init__(overlay)
        self.foo = foo

    def take_step(self):
        self.foo += 1
        self.overlay.do_foo(self.foo)

async def main():
    config = IPv8Configuration()
    config.overlays = [
        Overlay(
            cls="MyOverlay",
            walkers=[
            ## ??? Set only MyWalker with a value of foo = 13 ???
        ])
    ]
    ipv8 = IPv8(config, extra_communities={"MyOverlay": MyOverlay})
    await ipv8.start()

if __name__ == "__main__":
    ensure_future(main())
    get_event_loop().run_forever()

```

My solution is as follows (honor system: no peeking until you made your own solution):

<details> <summary> Click here for my solution </summary>

```python
class MyWalkerConfig(Walker):
    strategy = "MyWalker"
_WALKERS[MyWalkerConfig.strategy] = MyWalker

async def main():
    config = IPv8Configuration()
    config.overlays = [
        Overlay(
            cls="MyOverlay",
            walkers=[
                MyWalkerConfig(init={"foo": 13})
        ])
    ]
    ipv8 = IPv8(config, extra_communities={"MyOverlay": MyOverlay})
    await ipv8.start()
```

</details>

drew2a: > Ok. Good to hear that maintainability is not an issue for this class: let's switch to the next problem of code base size then.

Unfortunately, you get me wrong. The problem is not the size of a codebase, but its complexity. This PR adds complexity and we should ensure that it is necessary.

> Let me give a complete and runnable example/problem then
Thank you for the example.

The current PR forces me to write something like this:
```python
from asyncio import ensure_future, get_event_loop

from ipv8.community import Community
from ipv8.configuration_pydantic import IPv8Configuration, Overlay, Walker
from ipv8.peerdiscovery.discovery import DiscoveryStrategy
from ipv8_service import IPv8


class MyWalker(DiscoveryStrategy):
    def __init__(self, overlay, foo: int):
        super().__init__(overlay)
        self.foo = foo

    def take_step(self):
        self.foo += 1
        self.overlay.do_foo(self.foo)


class MyOverlay(Community):
    community_id = b'\xFF' * 20

    def get_available_strategies(self):
        return {'MyWalker': MyWalker}

    def do_foo(self, foo: int):
        print(f"[{foo}] FOO!")


class MyWalkerConfig(Walker):
    strategy = 'MyWalker'
    init = {'foo': 13}


async def main():
    config = IPv8Configuration()
    config.overlays = [
        Overlay(
            cls="MyOverlay",
            walkers=[
                MyWalkerConfig()
            ])
    ]
    ipv8 = IPv8(config, extra_communities={"MyOverlay": MyOverlay})
    await ipv8.start()


if __name__ == "__main__":
    ensure_future(main())
    get_event_loop().run_forever()
```

But I would prefer the following version (which works fine if there is no `AbstractModel`:

```python
from asyncio import ensure_future, get_event_loop

from ipv8.community import Community
from ipv8.configuration_pydantic import IPv8Configuration, Overlay, Walker
from ipv8.peerdiscovery.discovery import DiscoveryStrategy
from ipv8_service import IPv8


class MyWalker(DiscoveryStrategy):
    def __init__(self, overlay, foo: int):
        super().__init__(overlay)
        self.foo = foo

    def take_step(self):
        self.foo += 1
        self.overlay.do_foo(self.foo)


class MyOverlay(Community):
    community_id = b'\xFF' * 20

    def get_available_strategies(self):
        return {'MyWalker': MyWalker}

    def do_foo(self, foo: int):
        print(f"[{foo}] FOO!")

async def main():
    config = IPv8Configuration()
    config.overlays = [
        Overlay(
            cls="MyOverlay",
            walkers=[
                Walker(strategy='MyWalker', init={'foo': 13})
            ])
    ]
    ipv8 = IPv8(config, extra_communities={"MyOverlay": MyOverlay})
    await ipv8.start()


if __name__ == "__main__":
    ensure_future(main())
    get_event_loop().run_forever()

```

So, your example shows that `AbstractModel` didn't help me. Moreover, this class forced me to write unnecessary code.

drew2a: I checked your solution. Some parts of it require knowledge about IPv8 internals, but this is fine.

Anyway even this solution will look more elegant without `AbstractModel`:

```python
_WALKERS['MyWalker'] = MyWalker

async def main():
    config = IPv8Configuration()
    config.overlays = [
        Overlay(
            cls="MyOverlay",
            walkers=[
                Walker(init={"foo": 13}, strategy='MyWalker')
        ])
    ]
    ipv8 = IPv8(config, extra_communities={"MyOverlay": MyOverlay})
    await ipv8.start()
```

qstokkink: Could I ask you to define a single problem that I can address? You've changed the problem FIVE TIMES now. These statements are all taken from your own text:

> NIT: the purpose of this mixin is not clear.

> I don't see a necessity of using AbstractModel here.

> Basically, you use AbstactModel for helping the user to create a valid non-default config, if we assume he extensively uses format_schema_recursive/format_schema_single. [...] However, I doubt that this is a realistic scenario. 

> Because all code that has been written should be maintained. That is why you should avoid writing unnecessary code. 

> Everyone from the team is able to maintain this class. The problem is that the codebase of Tribler and IPv8 is huge.

> Unfortunately, you get me wrong. The problem is not the size of a codebase, but its complexity.

We've moved from the problem being (1) unknown class purpose, (2) a lack of necessity, (3) a lack of realistic use case, (4) code needing maintenance, (5) the code base becoming too big, and now (6) the complexity of the code base. Your last three problem definitions even directly contradict each other.

I'll end this reply with my rebuttal for all of the above except the original NIT raised, which was already resolved in the code.

---

Thanks for filling out the example. You perfectly provided the two cases I was looking for:

```python
class MyWalkerConfig(Walker):
    strategy = 'MyWalker'
    init = {'foo': 13}

walkers = [
    MyWalkerConfig()
]
```

vs.

```python
walkers = [
    Walker(strategy='MyWalker', init={'foo': 13})
]
```

The second is shorter and it will work, yes. However, the first option (1) describes the `MyWalker` class without the `MyWalker` class code, (2) exists without being included in the list of walkers (meaning it can be part of the "API" of third party community designers), (3) will be recommended by `AbstractModel` to people that don't know what walkers are made available by third party overlay engineers, (4) is declarative instead of anonymous (the same reason you would want `NamedTuple` instead of `tuple`), (5) lets you distinguish the intention of configurations that use the same strategy, and (6) does not turn the list of walkers into a spaghetti block of arguments.

You can get all of those advantages with the complexity of 11 lines of code, of which 5 are used to format an error message: a single if-statement in the `__new__` method of `AbstractModel`.

Coming back to your problems, starting with complexity: do these 11 lines of code add complexity? Yes, all code adds complexity - but this is a single if-statement performing a textbook operation. Does this make the code base bigger? Yes, all code does - but this only adds 11 lines of code (ignoring docstrings). Is this code hard to maintain? Perhaps, for people that are not familiar with Python class initialization - but I will gladly shepherd this code. Is the scenario unrealistic? No - at least one person will benefit from this. Is this code necessary? Perhaps not for everyone - but it will at least make their resulting code more approachable for others.

In conclusion, even after working through all of your arguments and problems, I only see added value from `AbstractModel`.

drew2a: > You've changed the problem FIVE TIMES now.

It is the same problem. The sentences that you extracted from the comments are my attempts to explain the problem.

The problem is: "In this PR you have added unnecessary complexity by introducing the `AbstractModel` class".

---

Let me comment on the "advantages" of mandatory intermediate config classes:

> (1) describes the MyWalker class without the MyWalker class code, 

In some cases, the code could benefit from splitting between abstractions (examples: MVC, DBAL, etc). In some cases not. As an `IPv8`  user I would prefer to decide by myself, do I want to introduce another layer of abstraction or not. 
With the current implementation it is no possible. `AbstractModel` forces me to write intermediate config classes.

> (2) exists without being included in the list of walkers (meaning it can be part of the "API" of third party community designers),

The example without `AbstractModel` class also has this ability. 

> (3) will be recommended by AbstractModel to people that don't know what walkers are made available by third party overlay engineers, 

No one has such a need yet. 

However, from my POV using the `AbstractModel`+`format_schema_recursive` is not the best way for "third party overlay engineers" to inform their users about their walkers.
A more straightforward and intuitive way is to provide good documentation with the list of added walkers. 
Or, if users want this list in runtime, "third-party overlay engineers" could provide a variable with this list.

> (4) is declarative instead of anonymous (the same reason you would want NamedTuple instead of tuple), 

With no-`AbstractModel` approach, developers can write code both declarative and anonymous.  So it is not a benefit of the `AbstractModel` approach.

Moreover, It is not always true that developers prefer `NamedTuple` instead of `tuple`.

> (5) lets you distinguish the intention of configurations that use the same strategy, and 

The same as above. It is not a benefit of the `AbstractModel` approach. Without `AbstractModel` developers could also write code in this manner.

> (6) does not turn the list of walkers into a spaghetti block of arguments.

I would say that for a configuration sometimes it is even better to have code with "the spaghetti block of arguments" because it is easier to read and understand. 

But, again, It is up to the end user how they want to organize the code that uses `IPv8`.

drew2a: Thanks, @qstokkink for the latest comments! I guess I figure out another problem with `AbstractModel`. I'll post it in another thread.

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,402 @@
...
+from typing import Any, Dict, List, Optional, Set, Type, Union
+
+from pydantic import BaseModel, Field, validator
+from pydantic.validators import dict_validator
+
+from .configuration import DISPERSY_BOOTSTRAPPER, default
+from .keyvault.crypto import default_eccrypto
+
+
+class AbstractModel:
```
drew2a, line 33: The problem: "`AbstractModel` is too restrictive to the end users (developers)."

In the thread below we came up with two possible examples of config usage

1: 
```python
class MyWalkerConfig(Walker):
    strategy = 'MyWalker'
    init = {'foo': 13}

walkers = [
    MyWalkerConfig()
]
```

2:
```python
walkers = [
    Walker(strategy='MyWalker', init={'foo': 13})
]
```

As an end user, I would like to have both options of using the `IPv8` configuration. 
Therefore I can decide which approach suits my application better. 
`AbstractModel` forces me to use only (1) approach.

ipv8/configuration_pydantic.py
qstokkink: _[Sorry about the late reply, some other stuff came up.]_

From your previous response, and my own response to that (below), I gather that you're mostly opposed to declarative configuration. Most of my answers consist of "at that point pydantic adds no value over dict, or dataclass, or `TypedDict`". At this point, I'm no longer seeing the added value of supporting `pydantic`. In the spirit of avoiding unnecessary complexity, should we perhaps drop this PR and leave configuration out-of-scope for the IPv8 respository and leave it to users?

That aside, my response to the individual points is as follows:

---

> > (1) describes the MyWalker class without the MyWalker class code,
>
> With the current implementation it is no possible. `AbstractModel` forces me to write intermediate config classes.

Yes, by design.

> > (2) exists without being included in the list of walkers (meaning it can be part of the "API" of third party community designers),
>
> The example without AbstractModel class also has this ability.

Sure, so does a dictionary (and a dataclass, and `TypedDict`, etc.). The difference is how explicit it is.

> > (3) will be recommended by AbstractModel to people that don't know what walkers are made available by third party overlay engineers,
>
> No one has such a need yet.

I had a need. I am someone.

> > (4) is declarative instead of anonymous (the same reason you would want NamedTuple instead of tuple),
>
> With no-AbstractModel approach, developers can write code both declarative and anonymous. So it is not a benefit of the AbstractModel approach.

(also in response to 5) At that point you don't even need pydantic anymore. The currect `dict` can do the same. If you want more, you can use dataclasses or `TypedDict`. The entirety of pydantic would be a wasted dependency.

> > (6) does not turn the list of walkers into a spaghetti block of arguments.
>
> I would say that for a configuration sometimes it is even better to have code with "the spaghetti block of arguments" because it is easier to read and understand.

Perhaps with very few entries. However, it is not uncommon to configure IPv8 with a dozen overlays, each with up to ~4 walkers. Then, it turns into a mess.

```python
walkers = [
    Walker(strategy='MyWalker', init={'foo': 13, 'bar': 1234, 'baz': "hello world!"}),
    Walker(strategy='MyWalker', init={'foo': 42, 'bar': 69, 'baz': "hello world!"}),
    Walker(strategy='CustomWalker2', init={'foo': "hey", 'bar': "there"})
]
```

vs.

```python
walkers = [
    FastMyWalker(),
    SlowMyWalker(),
    CustomWalker2()
]
```

> I guess I figure out another problem with AbstractModel. I'll post it in another thread.

That is fine. I would appreciate it if you don't redefine your problem whenever I address it though. That is called gaslighting and I consider it to be a pretty rude thing to do to someone.

drew2a: Different perspectives can give a better illustration of the exact problem. It could be seen as a redefinition or changing the problem. But it is not, at least from my side.

The PR adds value to the `IPv8` library. For me, as a user of this library, this is a plus. It is more convenient and safer to use pydantic models rather than dicts. The internals of `IPv8` also could benefit from using pydantic as well. That's why I see this PR as an improvement.

The only question I tried to discuss in this thread is whether `AbstractModel` needs to be introduced.

qstokkink:
_administration notice:_ This PR contains too many (at least 5) fundamental ideological differences that do not seem to be reconcilable. Unless any new reconciliations are found (i.e., not arguments that sow further division), I'll close this PR on March 28, 2023.

kozlovsky:
Please give me a few hours to finish the review!

qstokkink:
@kozlovsky I'll keep it open for the remainder of the working day. However, you're essentially reviewing a dead pull request.

ipv8/configuration_pydantic.py
kozlovsky: I agree with the importance of the config creator being able to specify that only subclasses are suitable in some places, and in that sense, `AbstractModel` serves its purpose. However, I wonder if it might be too restrictive to specify this restriction at the class level and control it directly when the class instance is created.

IPv8 is a fundamental-level library, and many other high-level libraries and applications with different configs can be built upon it. As such, a config class that cannot be directly instantiated in one library or application might be perfectly suitable for use in another library. While it is true that the IPv8 config does not allow directly specifying a `Walker` instance as an element of the `Overlay.walkers` field, it is possible that some other library built upon IPv8 allows a base `Walker` instance in some part of its obscure and esoteric config. We might not want to dictate for this other library never to directly use a `Walker` instance.

For that reason, it seems more flexible to me to specify the restriction not at the class level but at the field level: it would be the `Overlay.walkers` field that should not allow a `Walker` instance in its validation rules, and the `Walker` class itself probably should not make the assumption of whether it is suitable as an element or not.

I propose an approach to implement field-level restriction, demonstrated in the following example:

```python
from typing import Any, List, Type

import pytest
from pydantic import BaseModel, ValidationError, validator


def check_subclass_only(base_class: Type) -> Any:
    def validator_func(cls, value):
        if type(value) is base_class:
            raise ValueError(f"Only instances of subclasses of {base_class.__name__} are allowed")
        return value
    return validator_func


class Pet(BaseModel):
    name: str

    class Config:
        arbitrary_types_allowed = True


class Dog(Pet):
    breed: str


class Cat(Pet):
    color: str


class Owner(BaseModel):
    name: str
    pets: List[Pet]

    @validator("pets", each_item=True)
    def check_pet_subclass(cls, pet):
        return check_subclass_only(Pet)(cls, pet)


def test_owner_pets_subclass_validation():
    dog = Dog(name="Buddy", breed="Golden Retriever")
    cat = Cat(name="Whiskers", color="black")
    pet = Pet(name="Invalid Pet")

    with pytest.raises(ValidationError):
        owner = Owner(name="John Doe", pets=[dog, cat, pet])
```

In this example, `Owner.pets` field does not allow instances of a base `Pet` class as its elements. The same approach can be applied for `Overlay.walkers`.

This way, it becomes possible to maintain all current restrictions while allowing developers of IPv8-based libraries more freedom in deciding whether to have similar restrictions or not in their extended configs.

`AbstractModel` not only restricts direct instantiation of a class but also helpfully lists all subclasses specified in the same parent module for the user. To keep this functionality with the validator approach, we could extend the `check_subclass_only` function:


```python
def check_subclass_only(base_class: Type) -> Any:
    def validator_func(cls, value):
        if type(value) is base_class:
            msg = f"Only instances of subclasses of {base_class.__name__} are allowed"
            parent_module = sys.modules[cls.__module__]
            possibilities = [cname for cname, cinstance in inspect.getmembers(parent_module)
                             if isinstance(cinstance, type) and issubclass(cinstance, base_class)
                             and cinstance is not base_class]
            if possibilities:
                msg += f"Perhaps you want one of these: {possibilities}?"
            raise ValueError(msg)
        return value
    return validator_func
```

This way, it is possible to not use "mro magic", so the code probably looks slighlty easier to read and understand.

So, I think I prefer the field-level restrictions, but the `AbstractModel` approach is also possible.

ipv8/test/test_configuration_pydantic.py
```diff
@@ -0,0 +1,253 @@
...
+                "key": "string",
+                "on_start": [],
+                "walkers": ["#/definitions/Walker"]  # List of nested BaseModels
+            }
+        }
+        self.assertDictEqual(expected, format_schema_recursive(Overlay))
+
+    def test_format_schema_enum(self):
+        """
+        Check if schemas for Enums (also allowed inside of pydantic BaseModels) are correctly generated.
```
kozlovsky, line 237: "of" looks a bit redundant here
```suggestion
        Check if schemas for Enums (also allowed inside pydantic BaseModels) are correctly generated.
```

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,402 @@
...
+class BaseConfigurationClass(BaseModel):
+    """
+    Augment the ``BaseModel`` class with 1) dict cast unwrapping and 2) inherited validators.
+    """
+
+    def __iter__(self, *args, **kwargs):
+        """
+        Make ``dict(model)`` equal to ``model.dict()``.
+
+        Normally this function only only gives the child model references instead of unwrapping them.
```
kozlovsky, line 109: Typo: duplicated word
```suggestion
        Normally this function only gives the child model references instead of unwrapping them.
```

kozlovsky (approved):
In my opinion, this PR adds valuable functionality, and to me, it seems like a reasonable trade-off. Pydantic is a complex library, and it might be better for IPv8 not to depend on it, given that the lifespan of IPv8 could be longer.

For validation, I slightly prefer the field-based approach, as it appears more flexible and extendable. However, the currently implemented approach is also valid and works successfully.

I believe that in the distant future, IPv8 could benefit from a dataclass-based main config, but this is debatable and may require significant refactoring, potentially leading to backward incompatibilities. As such, the current approach seems like a good middle path.

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    Create a schema (``dict``) describing a BaseModel class.
+
+    |
+    For example, the following pretty-prints the schema of ``IPv8Configuration``:
+
+    ::
+
+        from json import dumps
+        print(dumps(format_schema_recursive(IPv8Configuration), indent=4))
```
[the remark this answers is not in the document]
qstokkink: With the goal of offering a runtime reference/reminder, I have created this method (instead of using the default) for two reasons:

1. Reducing the output size.
2. Increasing abstraction.

The output size difference is substantial. The default `schema_json` outputs 797 lines, while my method provides the necessary information in 49 lines. On my screen resolution and with my editor, I can see 52 lines at once. Therefore, this is the difference between reading through 16 pages of text or having all of the info at once.

The defaults and general abstraction level offered by `schema_json` are overly complex for the goal of capturing the essential schema. For example, lets take the `Overlay` and its `initialize` and `on_start` fields. In `schema_json` they are given as follows (nested 4 levels deep!):

```json
                "initialize": {
                    "title": "Initialize",
                    "type": "object"
                },
                "on_start": {
                    "title": "On Start",
                    "default": [],
                    "type": "array",
                    "items": {
                        "type": "array",
                        "items": {}
                    }
                }
```

My output is (disclaimer: in my opinion) equally expressive with regard to the schema but with less nesting (2 levels less) and lesser cognitive complexity:

```json
        "initialize": {},
        "on_start": []
```

(note: the above should actually be `"on_start": [[]]`, this is a bug)

ipv8/configuration_pydantic.py
```diff
@@ -0,0 +1,402 @@
...
+    def __iter__(self, *args, **kwargs):
+        """
+        Make ``dict(model)`` equal to ``model.dict()``.
+
+        Normally this function only only gives the child model references instead of unwrapping them.
+        """
+        return iter(self.dict().items())
+
+    @classmethod
+    def get_validators(cls):
```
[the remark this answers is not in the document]
qstokkink: Apparently [this is deprecated syntax](https://prompt-build--pydantic-docs.netlify.app/#custom-data-types). The new way is indeed with underscores.

qstokkink:
@kozlovsky thanks for the insights. I think your commentary is helpful if anyone wants to reuse this PR in the future. I added to the discussion where you requested it (or where I felt more needed to be mentioned).

That said, I'll still close this PR because its controversy remains. In fact, I consider your suggestion of class mro/subclass validation (through a function and/or field validators) to be in even starker contrast to the concerns of @drew2a than my own approach.

qstokkink:
Alright, this PR will now join the majestic list of [PRs that never got merged](https://github.com/Tribler/py-ipv8/pulls?q=is%3Apr+is%3Aclosed+is%3Aunmerged) as entry number 52.

As part of the reason for closing this PR is that there are fundamental differences in opinion that can lead to infinite discussion, I'll also lock the conversation.

PR #1132, opened 2023-05-31
Fix discovery cache creation

### LINKED ISSUE #1131: `Network.get_peers_for_service()` can cause peers to vanish
For unknown reason, a call to `Network.get_peers_for_service()` can (in some cases) cause a following call to `Network.get_peers_for_service()` to produce wrong results. This call _should_ only be a getter and not modify the state of the `Network` instance. However, with the right series of successive calls, its return values change.

This unit test (added to `TestNetwork`) reproduces the issue:

```python
    def test_cache_reconstruct(self):
        """
        Check if services caches are reconstructed properly.
        """
        service1 = bytes(range(20))
        service2 = bytes(range(20, 40))
        service3 = bytes(range(40, 60))
        self.network.reverse_service_cache_size = 1
        self.network.add_verified_peer(self.peers[0])
        self.network.add_verified_peer(self.peers[1])

        self.network.discover_services(self.peers[0], [service1, ])
        # We know of {<Peer 0>: ["service 1"]}

        peers_for_s3_none = [p for p in self.network.get_peers_for_service(service3)]
        # > Result for "service 3" should be an empty list.
        self.assertListEqual([], peers_for_s3_none)

        self.network.get_peers_for_service(service2)
        self.network.discover_services(self.peers[0], [service3, service2])
        # We know of {<Peer 0>: ["service 1", "service 2", "service 3"]}

        self.network.get_peers_for_service(service2)  # <--- !!!This breaks the test!!! --->>>
        self.network.discover_services(self.peers[1], [service3])
        # We know of {<Peer 0>: ["service 1", "service 2", "service 3"], <Peer 1>: ["service 3"]}

        peers_for_s3_two = [p for p in self.network.get_peers_for_service(service3)]
        # > Result for "service 3" should be [<Peer 0>, <Peer 1>].
        self.assertIn(self.peers[1], peers_for_s3_two)
        self.assertIn(self.peers[0], peers_for_s3_two)
```

By commenting out the line marked with `# <--- !!!This breaks the test!!! --->>>`, the unit test passes. To reiterate, this is only a getter and _should not change_ the outcome of the following `get_peers_for_service()` call.

EDIT: This unit test can be smaller:

```python
    def test_cache_reconstruct(self):
        """
        Check if services caches are reconstructed properly.
        """
        service1 = bytes(range(20))
        service2 = bytes(range(20, 40))
        self.network.reverse_service_cache_size = 1
        self.network.add_verified_peer(self.peers[0])
        self.network.add_verified_peer(self.peers[1])

        self.network.discover_services(self.peers[0], [service2, service1])
        self.network.get_peers_for_service(service1)  # <--- !!!This breaks the test!!! --->>>
        self.network.discover_services(self.peers[1], [service2])

        peers_for_s2_two = [p for p in self.network.get_peers_for_service(service2)]
        self.assertIn(self.peers[1], peers_for_s2_two)
        self.assertIn(self.peers[0], peers_for_s2_two)
```

qstokkink:
Fixes #1131

This PR:

 - Fixes `Network.discover_services()` creating new caches instead of just updating existing ones.
 - Removes calls to `Network.reverse_service_lookup.pop()` using peer addresses.

### DIFF
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -274,7 +274,6 @@ class Network(object):
         """
         with self.graph_lock:
             self._all_addresses.pop(address, None)
-            self.reverse_service_lookup.pop(address, None)
             # Note that the services_per_peer will never be 0, we abuse the lazy `or` to pop the peers from
             # the services_per_peer mapping if they are no longer included. This is fast.
             self.verified_peers = {peer for peer in self.verified_peers
@@ -290,7 +289,6 @@ class Network(object):
         with self.graph_lock:
             for address in peer.addresses.values():
                 self._all_addresses.pop(address, None)
-                self.reverse_service_lookup.pop(address, None)
             if peer in self.verified_peers:
                 self.verified_peers.remove(peer)
             self.verified_by_public_key_bin.pop(peer.public_key.key_to_bin(), None)
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -110,12 +110,13 @@ class Network(object):
             else:
                 self.services_per_peer[key_material] |= set(services)
             for service in services:
-                service_cache = self.reverse_service_lookup.get(service, [])
-                # Ensure that the peer instance in the cache is the same one as in verified_peers.
-                if peer in service_cache:
-                    service_cache.remove(peer)
-                service_cache.append(self.verified_by_public_key_bin.get(key_material, peer))
-                self.reverse_service_lookup[service] = service_cache
+                service_cache = self.reverse_service_lookup.get(service, None)
+                if service_cache is not None:
+                    # Ensure that the peer instance in the cache is the same one as in verified_peers.
+                    if peer in service_cache:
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

ipv8/peerdiscovery/network.py
```diff
@@ -273,7 +274,6 @@ def remove_by_address(self, address: Address) -> None:
         """
         with self.graph_lock:
             self._all_addresses.pop(address, None)
-            self.reverse_service_lookup.pop(address, None)
```
kozlovsky, line 276: Can you explain why this line can be removed? (as well as the following similar line)

qstokkink: Sure: `reverse_service_lookup` is a `Cache of service_id -> [Peer]` (see line 57) and `address` is of type `Address` (the same is true for the next line `Peer.address`). This will never fail but it will also never do anything useful because addresses are not part of the `reverse_service_lookup`. Therefore, it has been removed.

kozlovsky (approved):

PR #1142, opened 2023-06-26
WIP: Fixed vp_compile not passing default values

### LINKED ISSUE #1141: `vp_compile` does not preserve default values in init
### Traceback

The following is the intended way to specify defaults in `VariablePayload` subclasses:

```python
class T(VariablePayload):
    format_list = ['I', 'H']
    names = ["a", "b"]

    def __init__(self, *args, b=3, **kwargs):
        super().__init__(*args, b, **kwargs)

T(1)  # OK
```

For the base `VariablePayload` this works. However, this fails when using `vp_compile`:

```python
@vp_compile
class T(VariablePayload):
    format_list = ['I', 'H']
    names = ["a", "b"]

    def __init__(self, *args, b=3, **kwargs):
        super().__init__(*args, b, **kwargs)

T(1)  # TypeError: __init__() missing 1 required positional argument: 'b'
```

### Cause

This happens because `.messaging.lazy_payload._compile_init()` only takes the names defined in the `Payload` and does not bother to compile in the defined defaults, producing:

```python
class T(Payload):
    format_list = ['I', 'H']
    names = ["a", "b"]
    __match_args__ = ("a", "b")

    def __init__(self, a, b):
        self.a = a
        self.b = b

    def from_unpack_list(...): ...

    def to_pack_list(...): ...
```

What would be expected is the following:

```python
class T(Payload):
    format_list = ['I', 'H']
    names = ["a", "b"]
    __match_args__ = ("a", "b")

    def __init__(self, a, b=3):
        self.a = a
        self.b = b

    def from_unpack_list(...): ...

    def to_pack_list(...): ...
```

### Fix

The fix for this issue is to pass the default values from the orginal `inspect.signature()` of the parent `__init__()`.

qstokkink:
Fixes #1141

This PR:

 - Fixes `vp_compile` to pass on default values, like its uncompiled counterpart.


qstokkink:
validate

qstokkink:
validate

InvictusRMC (approved):

InvictusRMC:
Tested locally and everything seems to function. Nice job!

PR #1146, opened 2023-07-06
WIP: Added test for JSON serialization of config

### LINKED ISSUE #1124: Add unit test for config json serialization
One of the key parts of IPv8's base* configuration is its ability to be serialized to JSON. That means that all keys are UTF-8 strings and all values are [JSON-compatible types](https://docs.python.org/3/library/json.html#json.JSONEncoder).

We do not have tests for JSON serialization in `TestConfiguration`. Unit test(s) for this behavior should be added.

_*Individual overlays may choose to insert raw Python objects into the configuration dict, if they extend it._

qstokkink:
Fixes #1124

This PR:

 - Adds a test to check if the default configuration is JSON serializable (and retains its schema).

### DIFF
--- a/ipv8/test/test_configuration.py
+++ b/ipv8/test/test_configuration.py
@@ -1,10 +1,54 @@
 import base64
+import json
+
+import marshmallow
 
 from .base import TestBase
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

PR #1148, opened 2023-07-07
Added Network PeerObservers

### LINKED ISSUE #1145: Add verified peer removal callback
I'm observing a lot of instances of use cases for peer removal callbacks in Tribler and other experimental downstream projects.

Right now, to observe peer removal, users of IPv8 have to overwrite their `Network` instance:

```python
class MyNetwork(Network):

    def remove_peer(self, peer: Peer) -> None:
        print(peer.public_key, "was online for", time.time() - peer.creation_time, "seconds")
        super().remove_peer(peer)
```

Because the `Network` instance should (usually) be shared between `Community` instances, it is not always possible to perform this overwrite. Therefore, we should create a subscriber pattern that allows communities to subscribe to peer removal events.

qstokkink:
Fixes #1145

This PR:

 - Adds an observer pattern to `Network` (using the `PeerObserver` interface) that is notified upon peer addition or removal.

Note:
 - The notification happens inside of the graph lock to disallow observers causing more notifications while being notified.

### DIFF
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -1,5 +1,7 @@
+import abc
 import logging
 from collections import OrderedDict, namedtuple
+from operator import methodcaller
 from threading import RLock
 from typing import Dict, Iterable, List, Optional, Set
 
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

xoriole (approved):

PR #1150, opened 2023-07-10
Automated Version Update

### LINKED ISSUE #1116: `github_increment_version.py` depends on old PyGithub version
Our release authoring script (`github_increment_version.py`) depends on `PyGithub` version `1.54.0`, which is not the latest release (`1.57` - at the time of writing).

Currently, our authoring script works as expected with the old version of `PyGithub`. However, we may run into trouble later by depending on a deprecated version. We should look into updating `github_increment_version.py` to be compatible with the latest version of `PyGithub`.

qstokkink:
Fixes #1116

Suggested release message
---
Tag version: 2.11
Release title: IPv8 v2.11.1899 release
Body:
Includes the first 1899 commits (+29 since v2.10) for IPv8, containing:

 - Added Network PeerObservers
 - Added test for JSON serialization of config
 - Fixed MyPy issues after upgrade
 - Fixed PeriodicSimilarity sampling from set
 - Fixed circuit retry log message
 - Fixed discovery cache creation
 - Fixed truncation in verbose test results
 - Fixed vp_compile not passing default values
 - Removed `asynctest` requirement for Python > 3.7
 - Updated `IPv8` class to guard against `get_peers()`
 - Updated `run_all_tests.py` to handle segfaults
 - Updated LAN interface discovery
 - Updated Windows test runner script

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -26,9 +26,9 @@ copyright = '2017-2023, Tribler'  # Do not change manually! Handled by github_in
 author = u'Tribler'
 
 # The short X.Y version
-version = '2.10'  # Do not change manually! Handled by github_increment_version.py
+version = '2.11'  # Do not change manually! Handled by github_increment_version.py
 # The full version, including alpha/beta/rc tags
-release = '2.10.0'  # Do not change manually! Handled by github_increment_version.py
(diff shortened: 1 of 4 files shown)


qstokkink:
validate

xoriole (approved):

PR #1151, opened 2023-07-13
Updated `MockRestIPv8` to not use real socket

### LINKED ISSUE #1149: Don't use real socket for REST tests
Our REST tests are roughly a factor 100 slower than our other unit tests. ~~The main reason for these slow tests is that~~ _[EDIT: apparently sockets were not to blame for these slow tests]_ our tests open and use a real socket for these tests:

https://github.com/Tribler/py-ipv8/blob/0444baa7c09b74b4495f1cadbf97de548755c661/ipv8/test/REST/rest_base.py#L118

To speed up our test suite, we should avoid opening a real socket.

---

One solution to this issue is to implement the `aiohttp.web.BaseSite` interface and use the mocked internet approach of `AutoMockEndpoint`, instead of using `aiohttp.web.TCPSite` when testing:

https://github.com/Tribler/py-ipv8/blob/ea41739927a975442c7bcdeb28078e676087c614/ipv8/REST/rest_manager.py#L105

We would also have to modify the `RESTManager` to allow the site to be altered.

qstokkink:
Fixes #1149

This PR:

 - Updates the `MockRestIPv8` to use `Transport` instances based on IPv8's `MockEndpoint` (circumventing real sockets).

Notes:

 - Looking at the data, this PR makes the tests **slower** instead of faster: from `608 tests in 5.55 seconds (39.29 seconds total in tests)` to `7.03 seconds (44.46 seconds total in tests)`.

### DIFF
--- a/ipv8/REST/rest_manager.py
+++ b/ipv8/REST/rest_manager.py
@@ -101,6 +101,9 @@ class RESTManager:
 
         runner = web.AppRunner(self.root_endpoint.app, access_log=None)
         await runner.setup()
+        await self.start_site(runner, host, port, ssl_context)
+
+    async def start_site(self, runner, host, port, ssl_context):
         # If localhost is used as hostname, it will randomly either use 127.0.0.1 or ::1
         self.site = web.TCPSite(runner, host, port, ssl_context=ssl_context)
         await self.site.start()
--- a/ipv8/test/REST/rest_base.py
+++ b/ipv8/test/REST/rest_base.py
@@ -1,18 +1,23 @@
 import functools
+from asyncio import Transport
 from threading import RLock
 
-from aiohttp import ClientSession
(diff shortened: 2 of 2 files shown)


qstokkink:
validate

qstokkink:
This PR works and it is ready for review but I'm still not convinced we actually want this.

qstokkink:
I reflected on this for a few days. I believe it is in fact cleaner to not open real sockets during testing. Secondly, [the second of slowdown for the parallel Linux tests](https://github.com/Tribler/py-ipv8/issues/1149#issuecomment-1633878335) I think will be inconsequential since the other tests take more than a second longer anyway. Therefore, the total test time of a PR is not longer. All in all, I think this PR should be merged after all (after approval of course).

xoriole (approved):

Issue #1157, opened 2023-07-26
Don't read/write to files while unit testing


qstokkink:
Our `TestBase` case has facilities to create temporary files. However, ideally, these should never be used. Even so, this functionality is used in the following two locations:

```python
File "ipv8/test/REST/test_identity_endpoint.py", line 34, in create_node
    temp_dir = self.temporary_directory()
```

https://github.com/Tribler/py-ipv8/blob/84df0143ad19710bcc084a2358b191af89e8118b/ipv8/test/REST/test_identity_endpoint.py#L29-L50

<details> <summary>Fixed in #1158</summary>

```python
File "ipv8/test/attestation/wallet/test_attestation_community.py", line 307, in test_load_key
    temp_folder = self.temporary_directory()
```

https://github.com/Tribler/py-ipv8/blob/0572084b5a797b088c1242a9bfad61aef3634216/ipv8/test/attestation/wallet/test_attestation_community.py#L302-L323

</details>

We should try to refactor these two cases to not use real file I/O.

PR #1158, opened 2023-07-26
Updated `test_load_key` to not use a real db file

qstokkink:
Related to #1157

This PR:

 - Updates `test_load_key` to use a mocked database response instead of reloading the database from an actual file.


qstokkink:
validate

PR #1159, opened 2023-07-26
Updated `TestIdentityEndpoint` to not use real files

### LINKED ISSUE #1157: Don't read/write to files while unit testing

qstokkink:
Fixes #1157

This PR:

 - Updates `CommunicationManager` to use a `PseudonymFolderManager` to perform all file I/O.
 - Updates `TestIdentityEndpoint` to use a dictionary instead of a real folder structure.

### DIFF
--- a/ipv8/attestation/communication_manager.py
+++ b/ipv8/attestation/communication_manager.py
@@ -9,7 +9,7 @@ from .identity.manager import IdentityManager
 from .wallet.community import AttestationCommunity
 from ..keyvault.crypto import ECCrypto
 from ..messaging.anonymization.hidden_services import HiddenTunnelCommunity
-from ..types import IPv8, Peer
+from ..types import IPv8, Peer, PrivateKey
 from ..util import succeed
 
 AttributePointer = typing.Tuple[Peer, str]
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

ipv8/attestation/communication_manager.py
```diff
@@ -189,7 +189,62 @@ def verify(self, peer: Peer, attribute_hash: bytes, reference_values: typing.Lis
...
+        """
+        Manage/store pseudonyms in a given directory.
+
+        :param directory: the directory to store pseudonyms
+        :returns: None
+        """
+        self.pseudonym_folder = directory
+        self.crypto = ECCrypto()
+
+    def get_private_key(self, name: str) -> PrivateKey:
```
xoriole, line 207: With `get_` in the name it gives the sense of query method but it is also creating the key if it does not exist. Therefore, I suggest to rename as `get_or_create_private_key`.

qstokkink:
validate

xoriole (approved):

PR #1164, opened 2023-08-09
Updated codebase with pyupgrade

### LINKED ISSUE #1160: Replace deprecated typing versions of standard collections
The `typing` versions of standard collections are deprecated since Python 3.9 (see https://docs.python.org/3/library/typing.html#deprecation-timeline-of-major-features) and these should be replaced with their actual class type:

- `tuple` instead of `typing.Tuple`
- `list` instead of `typing.List`
- `dict` instead of `typing.Dict`
- `set` instead of `typing.Set`
- `frozenset` instead of `typing.FrozenSet`
- `type` instead of `typing.Type`
- `contextlib.AbstractContextManager` instead of `typing.ContextManager`
- `contextlib.AbstractAsyncContextManager` instead of `typing.AsyncContextManager`
- `re.Pattern` instead of `typing.Pattern` and `typing.re.Pattern`
- `re.Match` instead of `typing.Match` and `typing.re.Match`

This substitution will require `from __future__ import annotations` for older versions of Python.

I'll give this low priority, since no removal of these deprecated types is planned in any Python version yet.

qstokkink:
Fixes #1160

This PR:

 - Updates our code style to `pyupgrade --py37-plus`
 - Removes specifying list of nested `Serializable` instances in `dataclass` payloads without specifying the `list` type.

We're losing some functionality due to future compatibility:

 - Specifying list of nested `Serializable` instances in `dataclass` payloads without specifying the `list` type is now removed. For example:

```python
@dataclass
class Embedded:
    a: int

@dataclass
class Thing:
   a: [Embedded]  # Removed!
   b: list[Embedded]  # O.K.
   c: typing.List[Embedded]  # O.K. (but deprecated by Python)
```

 - Using inner classes as types is also removed. For example:

```python
@dataclass
class Embedded:
    a: int

@dataclass
class Thing:
   @dataclass
   class Inner:
       a: int
   a: Inner  # Removed: TypeError!
   b: Embedded  # O.K.
```

Both `test_payload_dataclass.py` and `communication_manager.py` use (future) deprecated typing syntax for backward compatibility.


qstokkink:
validate

qstokkink:
validate

qstokkink:
validate

qstokkink:
validate

xoriole (approved):

PR #1184, opened 2023-09-14
Fixed ruff violations for the irmaexact folder

qstokkink:
This PR:

 - Fixes the `ruff` issues in the `ipv8/attestation/wallet/irmaexact` folder.

This folder had the vast majority of violations so I'm separating it from the other `ipv8/attestation` fixes. However, some files will need to be revisited after the API's they implement are typed.


qstokkink:
validate

egbertbouman (approved):

PR #1187, opened 2023-09-20
Updated mid to always be the hash of the public key

qstokkink:
This PR:

 - Updates the `Peer.mid` to be the `SHA-1` of the public key, not the key (in most cases this would be the same though).


qstokkink:
validate

egbertbouman (approved):

Issue #1188, opened 2023-09-21
LAN discovery blocks asyncio thread


egbertbouman:
Since IPv8 2.11 `on_introduction_request`/ `on_new_introduction_request`intermittently takes 0.85s on my Windows 10 machine, blocking the `asyncio` thread. It turns out that the new LAN address discovery is causing the problems:

<p align="center"><img src="https://github.com/Tribler/py-ipv8/assets/3785124/49c0429f-5dc9-49e6-bfe8-87c12d489392" width=50% height=50%></p>

qstokkink:
This code is replacing `netifaces`, which [used to block for `1.577` seconds on some machines.](https://github.com/Tribler/tribler/issues/7427) With `0.85` seconds now, at least it's getting better.. I guess?

Jokes aside, it is especially important to wait for these results before sending out the first intro messages to the bootstrappers. Many of our experiments depend on local peers quickly realising that they are on each other's LAN (waiting for the next results after 10 seconds is unacceptable). All following calls can be dumped on some thread though. This is not trivial to implement.

_Administration_: I'll assign this issue "medium" priority for three reasons. Firstly, Tribler still ships with `netifaces` and this code is, therefore, not run in production (yet). Secondly, the blocking code is only executed every 10 seconds. Lastly, this does not crash the IPv8 application. However, this is definitely not "low" priority as blocking the reactor thread and, therefore, the socket for even a second can lead to major packet loss. This should be fixed before the next IPv8 release.

egbertbouman:
With `netifaces` (IPv8 2.10) I'm seeing no delays on the `asyncio` thread. So at least on my machine, the newer code is significantly worse.

Note that the issue you're referring to talks about a delay of 1.577s per 580 calls, which I would consider to be pretty fast. I'm currently seeing delays of 0.85s for a single call.

qstokkink:
I guess the problem with all of the calls in `messaging.interfaces.lan_addresses` is that each and every one of them _may_ block on _some_ machines.

Ah, yes you're right, the `1.577` was cumulative, my bad. The actual problem was described seems to be **worse** with `netifaces` described as "intermittently exhibits slow performance, sometimes taking several seconds to execute a single call, blocking the asyncio loop" (see same linked post). However, again, on _one particular machine_.

egbertbouman:
[...]

PR #1189, opened 2023-09-23
Updated LAN discovery to run on a different thread

egbertbouman:
This PR fixes https://github.com/Tribler/py-ipv8/issues/1188 by letting all communities periodically run a LAN discovery task using the threadpool. The `AddressProvider.addresses_ts` tries to avoid concurrent/frequent runs.

Note that blocking is still possible for the first call to `discover_addresses` (e.g., when `EndpointListener.__init__` calls `my_estimated_lan`). Although this may only be a problem when adding communities at runtime (which is what we do in Tribler). Alternatively, we could remove the first 2 lines from `get_addresses_buffered` and be unaware of our LAN address until one of the discovery tasks completes.

I removed `verbose` kwarg in interfaces.py because with this PR we can't use `lru_cache(maxsize=2)` anymore. Allowing the cache to be larger than 1 means we don't have a single timestamp for each `AddressProvider` anymore. For debugging purposes I added a global `VERBOSE` variable. I'm not seeing a need to change the verbosity at runtime, so hopefully this will do.

### DIFF
--- a/ipv8/messaging/interfaces/lan_addresses/addressprovider.py
+++ b/ipv8/messaging/interfaces/lan_addresses/addressprovider.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import traceback
 import typing
 from abc import ABC, abstractmethod
-from functools import lru_cache
 from time import time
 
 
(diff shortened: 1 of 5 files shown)


egbertbouman:
validate

qstokkink (approved):
Nice 👌 This turned out way cleaner than I had feared initially.

Issue #1175, opened 2023-09-05
Post-ruff minor issues


qstokkink:
Now that all the core IPv8 logic has had its `ruff` issues fixed, including its types, I'm left with a list of minor issues that should be solved at some point:

 - [x] Both `TestBase` and `DiscoveryStrategy` should be turned into `Generic[SomeOverlayType]`. *WHY: Having to assure mypy that they were loaded with a particular type of overlay, at runtime, is a pain.* **Fixed in** #1181 and #1192
 - [ ] Create a `CommunityConfig` (or a dictionary) to store `Community` configurations. *WHY: We now have 6 arguments to the `Community` constructor and, not only is this triggering `PLR0913` everywhere, this leads to mass repetition of the default arguments.*
 - [ ] Related to the previous point: the docs should be updated accordingly.
 - [x] `PD011` is throwing false-positive `numpy` errors and should be disabled. *WHY: We don't use `numpy`.* **Fixed in** #1190
 - [x] The `RequestCache` should type its return values. *WHY: Currently we need to cast all of the results of `pop()` from the `RequestCache`.* **Fixed in** #1191
 - [x] We could shorten some types to be more descriptive. *WHY: We now have very long types which are (probably) correct but - for example - one `bytes` may describe a 20 byte infohash while the other is a 64 byte public key.* **Fixed in** #1197

qstokkink:
I'll handle the `TestBase` generic-ification in the `test` folder PR.

qstokkink:
A `TypedDict` may be a good way to implement a `CommunityConfig` (or `CommunitySettings`) object.

PR #1191, opened 2023-09-25
Added typed request cache add/get/pop

qstokkink:
Related to #1175

This PR:

 - Adds overloaded versions of `RequestCache`'s `.add()`, `.get()`, and `pop()` that return the given class type.
 - Updates the simple cases where this new functionality can be used.
 - Removes all of the superfluous `typing.cast()` calls where the code was updated.

Pro:
 - We don't have to cast everywhere.

Con:
 - We have to mix-in `CacheWithName` for all caches that have a name. Thanks @xoriole for suggesting the `NumberCacheWithName` and `RandomNumberCacheWithName` shortcuts.

Ideally, I would've like to have `CacheTypeVar` bound to `NumberCache` and to implement the `CacheWithName` protocol. This is a construction that the `PyCharm` type checker recognizes but `mypy` unfortunately doesn't.

### DIFF
--- a/ipv8/messaging/anonymization/caches.py
+++ b/ipv8/messaging/anonymization/caches.py
@@ -7,7 +7,7 @@ from asyncio import Future
 from functools import reduce
 from typing import TYPE_CHECKING, Callable
 
-from ...requestcache import NumberCache, RandomNumberCache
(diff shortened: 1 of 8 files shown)


qstokkink:
validate

qstokkink:
validate

xoriole (approved):

PR #1199, opened 2023-09-28
Updated DiscoveryCommunity payloads to use ipv4 instead of 4SH

egbertbouman:
This PR updates the `DiscoveryCommunity` payloads to use `ipv4` instead of `4SH` for serializing IP addresses.
The rest of the codebase has already moved to the newer notation.


egbertbouman:
validate

qstokkink (approved):

PR #1200, opened 2023-09-28
Updated bootstrappers to be less complex

egbertbouman:
This PR is meant to simplify the bootstrapping process and introduces the following changes:
* DNS addresses are resolved on the threadpool.
* `DispersyBootstrapper.initialize`/`UDPBroadcastBootstrapper.initialize` are now defined as async functions, although implementations may still return a `Future`.
* Bootstrappers are no longer removed and re-added from `Community.bootstrappers` as a way of keeping track of which ones are initialized. It uses `Bootstrapper.initialized` instead.

### DIFF
--- a/ipv8/bootstrapping/dispersy/bootstrapper.py
+++ b/ipv8/bootstrapping/dispersy/bootstrapper.py
@@ -1,19 +1,16 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 from random import choice
 from socket import gethostbyname
-from threading import Thread
 from time import time
 from typing import TYPE_CHECKING, Iterable
 
 from ...messaging.interfaces.udp.endpoint import UDPv4Address
-from ...util import succeed
 from ..bootstrapper_interface import Bootstrapper
 
 if TYPE_CHECKING:
-    from asyncio import Future
-
     from ...community import Community
     from ...types import Address
 
@@ -33,6 +30,8 @@ class DispersyBootstrapper(Bootstrapper):
         """
         Create a new bootstrapper without resolving DNS addresses or initializing the blacklist.
         """
+        super().__init__()
+
         self.ip_addresses = [UDPv4Address(*addr) for addr in ip_addresses]
         self.dns_addresses = dns_addresses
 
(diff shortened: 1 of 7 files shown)


ipv8/bootstrapping/dispersy/bootstrapper.py
```diff
@@ -33,41 +30,41 @@ def __init__(self,
...
         """
         Initialize this bootstrapper for the given Community, settings its blacklist.
         """
-        overlay.network.blacklist.extend(self.ip_addresses)
-        self.resolve_dns_bootstrap_addresses()
-        return succeed(True)
+        if not self.initialized:
+            self.initialized = True
+            overlay.network.blacklist.extend(self.ip_addresses)
+            await self.resolve_dns_bootstrap_addresses()
```
qstokkink, line 66: It seems to me that after the resolution of the DNS addresses, these should also be added to the `overlay.network.blacklist`. If that is correct, it is a bug in the original code. If you agree, could you fix that in this PR?

ipv8/community.py
```diff
@@ -643,7 +621,8 @@ def get_new_introduction(self, from_peer: Peer | None = None) -> None:
             available = self.get_peers()
             if available:
                 # With a small chance, try to remedy any disconnected network phenomena.
-                if self._bootstrappers_initialized and random() < 0.05:
+                bootstrappers_ready = all(b.initialized for b in self.bootstrappers)
```
qstokkink, line 624: I don't understand the choice of `all` over `any` here. It seems to me that the latter is preferable.

qstokkink (commented):
I pulled out the pencil and notebook for this PR. The bulk of the logic seems to be in order. However, I have two questions left before I can approve.

Secondarily, nice catch with `ipv8_service.py` 😃

ipv8/community.py
egbertbouman: I used `all` to remain as close as possible to the old code. Before `_bootstrappers_initialized` would be set to `True` if all bootstrappers were initializing or initialized. Hence my choice to do the same here. Having said that, it probably doesn't matter so I'll change it.

ipv8/bootstrapping/dispersy/bootstrapper.py
egbertbouman: Agreed. I'll fix it.

egbertbouman: On second thought, this appears to be happening already by calling `ensure_blacklisted` in `get_addresses`.

egbertbouman:
validate

qstokkink (approved):

PR #1201, opened 2023-10-02
Updated overlays to use a settings parameter

### LINKED ISSUE #1175: Post-ruff minor issues

qstokkink:
Fixes #1175

This PR:

 - Updates all `Overlay` instances to take a `settings` object instead of passing individual settings to the `__init__()`.

### DIFF
--- a/doc/basics/testbase_tutorial_1.py
+++ b/doc/basics/testbase_tutorial_1.py
@@ -4,7 +4,7 @@ import os
 import unittest
 from typing import TYPE_CHECKING, Any
 
-from ipv8.community import DEFAULT_MAX_PEERS, Community
+from ipv8.community import Community, CommunitySettings
 from ipv8.requestcache import NumberCache, RequestCache
 from ipv8.test.base import TestBase
 from ipv8.test.mocking.ipv8 import MockIPv8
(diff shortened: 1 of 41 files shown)


qstokkink:
validate

qstokkink:
validate

qstokkink:
validate

doc/basics/testbase_tutorial_1.py
```diff
@@ -26,16 +26,18 @@ def on_timeout(self) -> None:
         self.overlay.timed_out = True
 
 
+class MyCommunitySettings(CommunitySettings):
```
egbertbouman, line 29: Shouldn't this be added to `MyCommunity.settings_class`?

egbertbouman:
validate

doc/basics/testbase_tutorial_1.py
qstokkink: Hold on, this entire test class is not passing 😨

qstokkink:
validate

doc/basics/testbase_tutorial_1.py
qstokkink: Strange. I wonder why the validation tests did not catch that.. 🤔

egbertbouman (approved):

PR #1204, opened 2023-10-03
Fixed crypto related issues in TunnelCommunity

egbertbouman:
This PR fixes a number of crypto related issues and does the following:
* Ignore timed out created/extended messages (to prevent https://github.com/Tribler/tribler/issues/5513)
* Ignore cells that can't be decrypted as this seems a much safer option then destroying the tunnel entirely.
* The `TunnelCommunity` was catching the wrong exception while verifying the shared secret.

Besides fixing some minor issues, this PR simplyfies circuit destruction by letting `destroy_relay` destroy one side of the connection at a time.


egbertbouman:
validate

egbertbouman:
validate

qstokkink (approved):
<details> <summary> N.v.m. ignore this: there is no lock, I'm dumb </summary>
I'd prefer

```python
with contextlib.suppress(KeyError):
    request_cache.pop(*X)
```
over 

```python 
if request_cache.has(*X):
    request_cache.pop(*X)
```

because the `RequestCache`'s lock is released in between `has()` and `pop()`. However, we don't do threading anymore anyway so</details> I approve.

PR #1172, opened 2023-08-29
Fixed ruff violations for messaging anonymization folder

qstokkink:
This PR:

 - Fixes the `ruff` violations in the `messaging/anonymization` folder.

PR #1210, opened 2023-10-05
Revert of remove_circuit task register introduced in #1172

qstokkink:
This PR:

 - Updates tunnel removals in `hidden_services.py` to no longer register their tasks (reverting #1172).

Note:

The new (w.r.t. the revert) `_ =` assignment is to convince MyPy that we did not forget an `await` or `register_task()`.


qstokkink:
validate

xoriole (approved):

PR #1217, opened 2023-10-12
Fixed the GitHub validate action

### LINKED ISSUE #1056: Test all supported configurations after merge/validate

qstokkink:
Related #1056

```
Error:  .github#L1
a step cannot have both the `uses` and `run` keys
```

See also https://github.com/Tribler/py-ipv8/actions/runs/6492024002

This PR:

 - Fixes the PR validate action by separating the `run` and `uses` keys.

Note that this is PR `2/?`, without any guarantees on the upper limit needed to get this running.


qstokkink:
validate

qstokkink:
validate

egbertbouman (approved):

PR #1219, opened 2023-10-12
Updated validate action to run the unit tests

### LINKED ISSUE #1056: Test all supported configurations after merge/validate

qstokkink:
Fixes #1056

This PR:

 - Updates the validation GitHub action to run the unit tests.


qstokkink:
validate

qstokkink:
Progress! The entire matrix is now "sort of" being tested, see https://github.com/Tribler/py-ipv8/actions/runs/6494407565

jobs

However, for some reason `3.1` is being tested instead of `3.10`.

qstokkink:
validate

PR #1221, opened 2023-10-12
Fixed cache leaks in Network

### LINKED ISSUE #1220: Memory leak in Network
When letting IPv8 run for about 1,5 days, I'm seeing a lot of IPv4 addresses in the `Network` class. For instance, in `Network.reverse_intro_lookup` I have 2009 keys (despite the 500 cache limit), while the total sum of all objects in the lists is 113163.

Memory leaks in other dicts within this class may also exist, but so far I've only seen issues with `reverse_intro_lookup`.

Note that despite these numbers, it takes a while before this actually has a significant impact on memory usage. This will only affect users who run IPv8 24/7.

qstokkink:
Fixes #1220

This PR:

 - Fixes some peer updates in `Network` not checking the new size of a cache.

### DIFF
--- a/ipv8/peerdiscovery/network.py
+++ b/ipv8/peerdiscovery/network.py
@@ -139,6 +139,8 @@ class Network:
                     intro_cache.append(address)
                 else:
                     self.reverse_intro_lookup[peer] = [address]
+                    if len(self.reverse_intro_lookup) > self.reverse_intro_cache_size:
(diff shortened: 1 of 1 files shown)


qstokkink:
validate

qstokkink:
validate

xoriole (approved):

Issue #1232, opened 2023-10-18
Peer addresses aren't being updated


egbertbouman:
Peers sometimes fail to respond to introduction-requests. I noticed this behaviour before, and figured it would probably be due to `max_peers` being reached. However, after some debugging, I found that `peer.address` isn't updating.

The problem is as follows: `on_introduction_request` calls `add_verified_peer` in order to perform an address update (if needed). However, the address that it uses for updating is still the old address. This is because `lazy_wrapper` gets the peer from `self.network.verified_by_public_key_bin`, which contains old address.

PR #1233, opened 2023-10-18
Fixed peer addresses not being updated

egbertbouman:
Fixes https://github.com/Tribler/py-ipv8/issues/1232.


egbertbouman:
validate

qstokkink (approved):

PR #1234, opened 2023-10-18
Added IPv6 support to TunnelCommunity

egbertbouman:
This PR adds IPv6 support to the `TunnelCommunity` and finally closes https://github.com/Tribler/py-ipv8/issues/615. This change will break compatibility with old versions.

Note that when running the unittests in IPv6 mode, we're skipping the tests for `DiscoveryCommunity`/`DispersyBootstrapper` and the old-style introduction requests, since they can't support IPv6.

Also note that I've changed `Address.pack` to work on all types of addresses (namedtuple or not). The reason behind this is that a lot of addresses are still just regular tuples (e.g., `my_estimated_lan`).


egbertbouman:
validate

qstokkink (approved):

PR #1241, opened 2023-10-20
Removed `distutils_forwardport.py`

### LINKED ISSUE #1239: Get rid of `distutils_forwardport.py` A.S.A.P.
For Python 3.12 support, we currently have the `distutils_forwardport.py` file to provide forward compatibility. This is a temporary measure to keep our framework up-and-running while our dependencies update. As soon as we can, we should get rid of this overhead.

We should - at least - reevaluate our dependencies before the next IPv8 release. If this file cannot be removed before our next release, it should be included in our setup:

https://github.com/Tribler/py-ipv8/blob/d95b11cee49d669961883ea77453ac975fbbf5f2/setup.py#L16

### LINKED ISSUE #1240: Missing dependencies in `setup.py`
Currently, our `setup.py` only has the following dependencies specified:

https://github.com/Tribler/py-ipv8/blob/d95b11cee49d669961883ea77453ac975fbbf5f2/setup.py#L17-L29

This is not up-to-date with our requirements:

https://github.com/Tribler/py-ipv8/blob/d95b11cee49d669961883ea77453ac975fbbf5f2/requirements.txt#L1-L11

This should be solved before the next IPv8 release.

qstokkink:
Fixes #1239
Fixes #1240

This PR:

 - Fixes `setup.py` to include all of our high-level dependencies (note that `apispec>=6.0.0` is an indirect dependency and is not included).
 - Removes `distutils_forwardport.py` from our codebase.

Notes:
 - `aiohttp_apispec` is now on a beta version, which is something I tried to avoid initially. Even so, this seems to be preferable over shipping `distutils_forwardport.py`.


qstokkink:
validate

egbertbouman (approved):

PR #1244, opened 2023-11-13
Added direction argument to encrypt_str/decrypto_str

egbertbouman:
The biggest part of this PR is focussed on:
* Adding a direction argument to `encrypt_str`/`decrypt_str`.  This simplifies the tunnel crypto as we don't have to pass the keys/salts separately anymore. 
* Removing `TunnelCommunity.directions`.

The remaining commits are small cleanups/fixes of our anonymization layer.


egbertbouman:
validate

qstokkink (approved):

PR #1260, opened 2024-01-16
WIP: Cache pip dependencies in validate runs

qstokkink:
Related to #1237

This PR:

 - Adds caching for `pip` dependencies in the validate action.

PR #1261, opened 2024-01-16
Fixed validation action errors

qstokkink:
Fixes issues reported in #1260

This PR:

 - Adds target urls for the validation action.
 - Fixes `ls-remote` to use `https://github.com/Tribler/py-ipv8` instead of `origin`, which doesn't exist.
 - Fixes a missing newline after the `ls-remote` call.
 - Updates the validation action job order to first set pending status and then start testing.


qstokkink:
validate

PR #1266, opened 2024-01-23
Updated IPv8 exit node script to use RustEndpoint

egbertbouman:
This PR does the following:
* Enables `RustEndpoint` in the IPv8 exit node script. I haven't enabled it for all IPv8 nodes, as I don't think it makes sense since normal nodes will be making calls from within Python anyway (removing any chance of a performance improvement).
* Add the ability for the `TunnelCommunity` to deal with `DispatcherEndpoints`.
* Fixes a potential `KeyError` in `relays_cell` for rendezvous points.
* Fixes an issue with leftover asyncio tasks after `TunnelExitSocket` removal.

I also fixed some issues with the `CommunitySettings`:
* Fixes an issue with incorrect peer flags during the tunnel tests.
* Fixes an issue with properties in `CommunitySettings`.
* Fixes an issue with the  default tunnel settings getting stored in `overlay.settings.settings` instead of in `overlay.settings`.

### DIFF
--- a/ipv8/attestation/wallet/primitives/cryptography_wrapper.py
+++ b/ipv8/attestation/wallet/primitives/cryptography_wrapper.py
@@ -28,7 +28,8 @@ def generate_safe_prime(bit_length: int, backend: Backend = default_backend()) -
     out = int(backend._ffi.string(generated_hex), 16)
     # Cleanup the memory
     backend._lib.OPENSSL_free(generated_hex)
-    backend._lib.BN_clear_free(generated)
+    backend._lib.BN_set_word(generated, 0)
+    backend._lib.BN_free(generated)
     return out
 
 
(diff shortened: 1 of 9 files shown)


egbertbouman:
validate

egbertbouman:
It looks like in `cryptography` 42.0 (released less than a day ago) the functions `BN_clear_free`/`BN_clear`/`BN_zero` aren't available anymore. So, I've removed `BN_clear_free` in favor of manually setting the BIGNUM to 0, followed by a call to `BN_free`. Not sure this is correct, but given that attestion logic isn't used anyway this seems OK.

egbertbouman:
validate

qstokkink (approved):
I'm guessing the attestation/identity scripts are somehow running an old version. I'll look at that later.

PR #1269, opened 2024-01-29
Updated hidden services to stop removing data circuits

egbertbouman:
This PR stops hidden services removing data circuits, as it may conflict with circuit management in Tribler.


egbertbouman:
validate

qstokkink (approved):

PR #1273, opened 2024-02-13
Fixed possibly unclosed sockets for LAN discovery

qstokkink:
This PR:

 - Fixes resource warnings when sockets are left to garbage collection if an `OSError` occurs before closing.

This fix is slightly ugly because the `close()` of an unclosed socket can again error out, leading to a nested `try .. except`.


qstokkink:
validate

egbertbouman (approved):

PR #1292, opened 2024-04-03
Fixed relaying to exits

egbertbouman:
Currently, peers in the `TunnelCommunity` will suggest extending to peers that just have `PEER_FLAG_RELAY` set, and not `PEER_FLAG_EXIT_BT`. This is meant to remove some of the strain on the exit nodes.

However, the circuit creator will currently only try to extend to the suggested peers. This means that circuit creation can needlessly fail. For instance, it's currently not possible to create a 2-hop circuit in a network that consists only of exit nodes, even if the exit nodes all have `PEER_FLAG_RELAY` set.

This PR addresses this issue by letting the circuit creator extend to non-suggested exit nodes as well.

### DIFF
--- a/ipv8/messaging/anonymization/community.py
+++ b/ipv8/messaging/anonymization/community.py
@@ -576,6 +576,7 @@ class TunnelCommunity(Community):
         circuit.unverified_hop = None
         circuit.add_hop(hop)
         self.circuits.get(circuit_id)  # Needed for notifying the RustEndpoint
+        self.logger.info("Added hop %d (%s) to circuit %d", len(circuit.hops), hop.peer, circuit.circuit_id)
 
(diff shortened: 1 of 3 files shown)


egbertbouman:
The linting errors are because a [bug in ruff 0.3.5](https://www.github.com/astral-sh/ruff/issues/10724).

egbertbouman:
validate

egbertbouman:
validate

qstokkink:
validate

egbertbouman:
validate

qstokkink:
validate

egbertbouman:
validate

egbertbouman:
validate

qstokkink (approved):

PR #1295, opened 2024-06-13
Cache libsodium.dll from the main branch

### LINKED ISSUE #1294: GitHub Actions caching is useless
Currently, our testing infrastructure restores and saves caches for the `libsodium.dll` file so that it doesn't need to be downloaded every PR run.

https://github.com/Tribler/py-ipv8/blob/80b3544d9bdc2b332e489808a67dfc351a6936e6/.github/workflows/unittests.yml#L23-L35

This does not work as intended. PR runs download the file over and over again.

The core issue is that PR branches are only allowed to access their own cache or their parent branch's cache, [see the docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache). Even though the cache key is equal, PRs do not use the cache that other PRs have created.

From the documentation, it does seem like moving cache creation to the parent branch (`master`) should resolve this issue. This would change our workflow(s) as follows:

 - Create an action on the `master` branch to download and `save@v4` the `cache_libsodium_dll` cache. The cache lifetime is 7 days, so this should probably be a CRON action that downloads every three days (allowing one retrieval to fail).
 - Remove saving from the PR unittest workflow. This apparently does nothing useful.

qstokkink:
Fixes #1294

This PR:

 - Fixes the Mac unit tests not finding `libsodium.dylib`.
 - Updates the coverage job to be periodic (with the intent of refreshing a cache).
 - Removes cache storage from the unittest job (this was doing very little).

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - master
+  schedule:
+    - cron: "10 6 * * 1,3,5" # Mon, Wed, Fri @ 6:10 UTC
   workflow_dispatch:
 jobs:
   linux:
@@ -35,7 +37,6 @@ jobs:
         with:
           html_report_name: coverage-windows
       - uses: actions/cache/save@v4
-        if: steps.restore_cache.outputs.cache-hit != 'true'
         with:
           path: libsodium.dll
           key: cache_libsodium_dll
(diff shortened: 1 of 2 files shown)


qstokkink:
validate

qstokkink:
validate

egbertbouman (approved):

PR #1312, opened 2024-10-16
Fix missing file key in ConfigBuilder

mg98:
This PR:

 - Fixes missing key `'file'` in `key_config` which would have led to a crash in another part


qstokkink:
validate

qstokkink (approved):

PR #1322, opened 2024-11-18
Update minimum Python version to 3.9

### LINKED ISSUE #1302: IPv8 3.0 - drop Python 3.7 and Python 3.8 (3.9?)
Coming October, official support for Python 3.8 will be dropped.

Currently, in the 2.x series, IPv8 still supports Python 3.7. This is blocking enhancements like #1196.

When Python 3.8 support is dropped, it seems like a good time to modernize IPv8 again and switch to a newer Python version. Of course, this is backwards incompatible and we should increment the IPv8 version accordingly to a new major version "3.x".

### LINKED ISSUE #1306: Reevaluate need for TaskManager on PyPI
The IPv8 repository contains [create_setup_taskmanager.py](https://github.com/Tribler/py-ipv8/blob/master/create_setup_taskmanager.py) (and [an associated GitHub Action](https://github.com/Tribler/py-ipv8/blob/master/.github/workflows/python-publish-taskmanager.yml)) to individually package and publish the `TaskManager` class to PyPI: https://pypi.org/project/ipv8-taskmanager/

I don't think the `ipv8-taskmanager` package has seen much use, at the very least not for us. That makes it a good candidate for deletion, to save on overhead in this repository.

Once a package is removed, I believe reclaiming the same name on PyPI takes quite a while. So, we should be very sure that we want it removed.

qstokkink:
Fixes #1306
Fixes #1302 

This PR:

 - Adds Python `3.13` and Python `3.14` as supported versions.
 - Fixes pending ruff violations, except for the `doc` folder.
 - Updates the minimum Python version to 3.9 (using `pyupgrade`). Note that this also includes using modern data type annotations (e.g., `typing.Set` becomes `set`, `typing.Tuple` becomes `tuple`, etc.).
 - Removes the `TaskManager` PyPI release.
 - Removes the deprecated attestation REST API.


qstokkink:
validate

qstokkink:
The validation uses the existing `pr-comment-validate.yml` on the repo and not the new modified version. So, I ran the following commands locally to simulate it:

```
doc/identity_tutorial_integration/attestation_tutorial_attest.py
doc/identity_tutorial_integration/attestation_tutorial_verify.py
doc/validate_examples.py
```

egbertbouman (approved):

PR #1326, opened 2024-11-20
Shortened PR doc validation

qstokkink:
This PR:

 - Updates `advanced_identity_1.py` and `advanced_identity_2.py` to stop after 1 second. This is not visible in the docs.
 - Updates `doc/validate_examples.py` to kill scripts after 5 seconds, instead of 30 seconds.
 
 Based on the previous PR: from `4m 31s` down to `1m 2s`.


qstokkink:
validate

egbertbouman (approved):

PR #1330, opened 2024-11-26
Added documentation for the DHT(Discovery)Community

### LINKED ISSUE #1087: Document DHTCommunity
The [`DHTCommunity`](https://github.com/Tribler/py-ipv8/blob/master/ipv8/dht/community.py) (and its subclass the [`DHTDiscoveryCommunity`](https://github.com/Tribler/py-ipv8/blob/master/ipv8/dht/discovery.py)) have very useful functionality to store data and to connect to peers based on their public key. This functionality allows IPv8 to serve as both a distributed file system and allows programmers to both create neighborhood persistence and superpeer networks.

We may want to leave IPv8-as-a-file-system functionality out of the documentation to avoid competing with Tribler's solution. However, it would be useful to give an example of storing known peers on custom `Community.unload()` and attempting to reconnect when initializing (i.e., neighborhood persistence). This one example would also cover "find-your-friend"/superpeer functionality.

qstokkink:
Fixes #1087

This PR:

 - Adds documentation for the `DHTCommunity` and the `DHTDiscoveryCommunity`.
 - Fixes a bug in the return value of `DHTCommunity.find()`.

### DIFF
--- a/doc/further-reading/dht.rst
+++ b/doc/further-reading/dht.rst
@@ -1,0 +1,44 @@
+DHT(Discovery)Community
+=======================
+
+This document contains a description of how to use the ``DHTCommunity`` class for distributed hash table (DHT) data storage and the ``DHTDiscoveryCommunity`` extension of this functionality, which provides functionality to connect given public keys.
+
+In particular this document will **not** discuss how distributed hash table work, for this we refer the reader to other resources on the Internet.
+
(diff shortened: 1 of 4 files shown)


qstokkink:
validate

egbertbouman (approved):

PR #1335, opened 2025-03-06
Updated TunnelEndpoint to test circuits in Rust

egbertbouman:
Recently, we've released a new version of the Rust endpoint that has its own speed test logic built-in. This PR modifies the `TunnelEndpoint` to start using this functionality.

In order for the test in Rust to work properly I had to change the identifiers in the test-request/response messages from 16-bit to 32-bit. Since the tests now work differently in Rust, I've disabled the Python version.


egbertbouman:
validate

PR #1341, opened 2025-07-29
Fixed uncaught getaddrinfo UnicodeError

qstokkink:
Related to upstream https://github.com/Tribler/tribler/issues/8686

This PR:

 - Fixes an uncaught `UnicodeError` being thrown when `socket.getaddrinfo()` is invoked. This is an unsolved bug in CPython.

It seems like we've developed some unfortunate `mypy` errors, since the last PR. Given the number of errors, I think that's best left to a separate PR.


qstokkink:
validate

egbertbouman (approved):

PR #1343, opened 2025-09-04
Fixed circuit creation

egbertbouman:
Fixes an issue where circuits can't be created anymore because all peers are already used as first hop.


qstokkink (approved):
🤔 ah yes, I almost forgot about mypy.

PR #1344, opened 2025-09-09
Fixed memory leak in GetIfAddrs

egbertbouman:
This PR fixes a memory leak in `GetIfAddrs`.

Instead of calling `getifaddrs` with `*ifap`, we now call it with `**ifap` (as per the docs). Additionally, calling `freeifaddrs` seems to be needed, despite the comment in the code stating otherwise.


egbertbouman:
validate

PR #1352, opened 2025-11-28
Add possible exits to extend candidates

egbertbouman:
This PR adds potential exits to the list of extend-candidates included in created/extended messages.

To avoid making breaking changes to the protocol, the first exit is added to the list twice, serving as a separator between the two lists (e.g. [relay1, relay2, exit1, exit1, exit2]).


egbertbouman:
validate

PR #1357, opened 2026-02-18
Disable IPv4 on UDPv6Endpoint

egbertbouman:
Fixes an issue where `UDPv6Endpoint` can receive packets from IPv4-mapped peers.


egbertbouman:
validate

PR #1362, opened 2026-05-13
Update rust module

### LINKED ISSUE #1361: Incompatible with `cryptography>=47.0.0`
Cross-post from Tribler. When using `cryptography>=47.0.0`:

```python
File "tribler\pyipv8\ipv8\keyvault\private\m2crypto.py", line 35, in __init__
    super().__init__(ec_pub=ec.generate_private_key(curve))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cryptography.exceptions.UnsupportedAlgorithm: Curve sect233k1 is not supported
```

We use custom `EllipticCurve` subclasses to tell `cryptography` what the group orders and key sizes are for our custom curves. In response, `cryptography` promptly _ignores these inputs completely_ and uses the name of the curve to see if it should error out: 

https://github.com/pyca/cryptography/blob/99de8d186717dffb964cb48c59beb9e261b6f119/src/rust/src/backend/ec.rs#L42-L64

egbertbouman:
Fixes https://github.com/Tribler/py-ipv8/issues/1361 by moving the core `keyvault` logic to `ipv8-rust-tunnels`. The new  implementation relies on OpenSSL.

Consequently, `ipv8-rust-tunnels` is now a required dependency, replacing `cryptography` and `libnacl`. Note that `PrivateKey` now has a `diffie-hellman` function, and both `PrivateKey`/`PublicKey` have a `get_crypt_sk` function. These are needed for X25519 DH (see `TunnelCrypto` on how to use it).


egbertbouman:
validate

qstokkink (approved):

